PR #32, opened 2018-03-16
Add stats functions to Python API

tdenniston:
Also add to docs.


jakebolewski:
We may want to conditionally enable these (based on the tiledb version).

http://cython.readthedocs.io/en/latest/src/reference/language_basics.html#conditional-compilation

PR #48, opened 2018-07-04
Rework setup process to download/install TileDB if necessary.

tdenniston:
The native library is installed with the wheel (as "package data") if setup.py downloads and builds libtiledb.


jakebolewski:
💯 !

PR #52, opened 2018-07-06
Docs, installation and dependency fixes

tdenniston:
- Add long_description with readme text
- Fix dependency specs in setup.py and requirements files
- Remove use of numpy 1.9 function np.full
- Intercept bdist_egg for native lib installation
- Fix RTD build failure.

### DIFF
--- a/doc/source/conda-env.yml
+++ b/doc/source/conda-env.yml
@@ -1,5 +1,8 @@
 name: libtiledb-env
+channels:
+  - conda-forge
+  - defaults
 dependencies:
-  - python=3.4
-  - tiledb
+  - python=3.5
   - gcc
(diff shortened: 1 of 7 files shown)


doc/source/conda-env.yml
```diff
@@ -1,5 +1,8 @@
 name: libtiledb-env
+channels:
+  - conda-forge
+  - defaults
 dependencies:
   - python=3.4
```
jakebolewski: can this be updated to 3.6 (tiledb-py won't work on python3.4 at the moment)?

tdenniston: Yes I'll update it (though this is just for RTD).

PR #67, opened 2018-07-10
Add 3.4 to supported versions; add to Travis.

tdenniston:


jakebolewski:
can you add this to the tox.ini file as well?

tdenniston:
Added it to tox.

PR #76, opened 2018-08-01
Fix --tiledb arg; don't add cmake unnecessarily.

### LINKED ISSUE #74: Skip cmake requirement if the cmake command is already available
Should be a simple check to see if `cmake` executable is available before adding it.

### LINKED ISSUE #75: setup.py --tiledb arg not handled correctly
E.g. `setup.py build_ext --tiledb=/Users/tyler/tiledb/TileDB/dist/lib` gives this error:

```
RuntimeError: Could not find given --tiledb library path(s): /Users/tyler/tiledb/TileDB/dist/lib/lib/libtiledb.dylib
```

But then `setup.py build_ext --tiledb=/Users/tyler/tiledb/TileDB/dist` gives:

```
RuntimeError: Could not find given --tiledb library path(s): /Users/tyler/tiledb/TileDB/dist/libtiledb.dylib
```

tdenniston:
Closes #74; closes #75.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -268,15 +268,30 @@ class LazyCommandClass(dict):
         return bdist_egg_cmd
 
 
+def cmake_available():
+    """
+    Checks whether CMake command is available and >= version 3.3.
+    :return:
+    """
+    try:
+        output = subprocess.check_output(['cmake', '--version']).split()
+        version = output[2].decode('utf-8').split('.')
+        return int(version[0]) >= 3 and int(version[1]) >= 3
+    except:
+        return False
+
+
 def setup_requires():
     req = ['cython>=0.27',
            'numpy>=1.7',
            'setuptools>=18.0',
            'setuptools_scm>=1.5.4',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -274,9 +274,13 @@ def setup_requires():
...
            'setuptools_scm>=1.5.4',
            'wheel>=0.30']
-    tiledb_path = [TILEDB_PATH] if TILEDB_PATH != '' else []
-    if not libtiledb_exists(tiledb_path):
-        req.append('cmake>=3.11.0')
+    # Add cmake requirement if libtiledb is not found and cmake is not available.
+    tiledb_lib_dir = [os.path.join(TILEDB_PATH, 'lib')] if TILEDB_PATH != '' else []
+    if not libtiledb_exists(tiledb_lib_dir):
+        try:
+            subprocess.check_call(['cmake', '--version'])
```
jakebolewski: should we check the version output here and make sure it is >= 3.11?

PR #101, opened 2018-11-11
fix windows library name as it was missing 

petacube:
i had fixed the followin error on windows
C:\Users\TileDB-Py>python -c "import tiledb"
  File "<string>", line 1, in <module>
  File "C:\Users\TileDB-Py\tiledb\__init__.py", line 16, in <module>
    ctypes.CDLL(os.path.join(lib_dir, lib_name))
NameError: name 'lib_name' is not defined

by modifying __init__.py in tiledb folder

if it easier to submit pull request into dev branch for tiledb i am happy to do it as well.


jakebolewski:
@petacube I have a couple of fixes for Windows in two outstanding PR's. We most importantly need to enable CI testing for Windows, which is currently blocked by the Appveyor toolchain not picking up the correct architecture in the build.  Still debugging that one.

petacube:
ok ok np
i was just trying to fix local unit test errors. 

i think the is only one  one outstanding and it has to do with path being hardcoded in unit tests for the linux os. 

Stan

ihnorton:
Fixed by #107

PR #104, opened 2019-01-17
Update python api for latest dev changes

Shelnutt2:
Update python api for new consolidation, removal of compressor api and other fixes/updates.


jakebolewski:
@ihnorton let's make, `config`  for `tiledb.consolidate()` into a kw argument as well, default to `None`.

PR #106, opened 2019-02-05
Fix macOS 10.14 build

ihnorton:
- as discussed, change target to TileDB stable while working on CI improvements.
- fix build on macOS 10.14, where libstdc++ is no longer available. The deployment target needs to be set in order to properly use libc++ (see `https://github.com/pandas-dev/pandas/pull/24274`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ from pkg_resources import resource_filename
 import sys
 from sys import version_info as ver
 
-TILEDB_VERSION = "dev"
+TILEDB_VERSION = "master"
 
 # Directory containing this file
 CONTAINING_DIR = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


ihnorton:
There is still one oddity: I have to run `setup.py build_ext ...` twice in order to get TBB to build. Is that a known issue?

Tests pass locally, noticed there is no mac CI right now.

jakebolewski:
It would be good to set that up, should be easy

ihnorton:
[mac CI on azure pipelines](https://dev.azure.com/isaiahnorton/isaiahnorton/_build/results?buildId=20) (windows is enabled there too. still WIP but close, it is picking up gcc from git bash I think).

ihnorton:
That is based on this PR, so I'd like to merge this and then make another PR to add the windows and mac CI, if that's ok. [Here is the diff](https://github.com/ihnorton/TileDB-Py/compare/fix_macos10.14_build...ihnorton:azure_ci) for azure pipelines -- straightforward, and so far mac support seems much better (and simpler) than travis.

PR #107, opened 2019-02-06
CI for mac and win (WIP) on Azure pipelines

ihnorton:
includes and closes #97 
fixes #98

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,10 +1,5 @@
-# Python package
-# Create and test a Python package on multiple Python versions.
-# Add steps that analyze code, save the dist with the build record, publish to a PyPI-compatible index, and more:
-# https://docs.microsoft.com/azure/devops/pipelines/languages/python
-
 trigger:
-- azure_ci
+- dev
 
 strategy:
   matrix:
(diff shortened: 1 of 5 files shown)


ihnorton:
WIP on Windows because TileDB and TileDB-Py both build, but tests currently fail because it can't load the DLL. Probably a path/install issue, but debugging on CI isn't really efficient so I'll debug on my Windows laptop at home.

ihnorton:
Success on mac: https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=2

ihnorton:
Ok, Windows build and install is succeeding here, and tests are running. There are several test failures I will have to dig into on laptop, but as far as build and CI this is just waiting on https://github.com/TileDB-Inc/TileDB/pull/1125 and 1126. Once that is done I will squash the commits.

jakebolewski:
Hmm, the returning a different coordinate dtype error is strange on Windows, and the group comparison probably needs some Win. specific changes.    Otherwise LGTM.

ihnorton:
> returning a different coordinate dtype error is strange on Windows

https://github.com/numpy/numpy/issues/9464 🤷‍♂️ ... Fixed in latest commit by using `np.int64`.

> and the group comparison probably needs some Win. specific changes

Yes: fixed by normalizing path so that the comparison succeeds (the internal path was using fwd slash.

CI passed on both platforms: https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=12

PR #129, opened 2019-03-29
HL API additions for dask

ihnorton:
This PR contains

- rework DenseArray to support single-attribute view on a multi-attribute array

- set  of API additions, motivated by dask support (but hopefully generally useful):
  - convenience functions:
    - `open`, `save`
    - aliased `from_numpy` to allow use as simply `tiledb.from_numpy`
  - util functions to make schema and array from numpy-style specifiers (rather than spelling out `Dim`,`Domain`)
    - `schema_like`: create ArraySchema based on a numpy ndarray template, or numpy-style `shape` and `dtype` kwargs
    - `empty_like`: returns a writable, emtpy array similar to an existing ndarray-like object (`.dtype`, `.shape`).

- basic CI for dask integration to make sure we don't break that here, going forward

### DIFF
--- a/tiledb/highlevel.py
+++ b/tiledb/highlevel.py
@@ -1,0 +1,69 @@
+import tiledb
+from tiledb.libtiledb import *
+
+import numpy as np
+
+import pdb
+def open(uri, key=None, attr=None, mode='r', config=None):
+    if config:
+        cfg = tiledb.Config(config)
+        ctx = tiledb.Ctx(cfg)
+    else:
+        ctx = default_ctx()
+
+    schema = ArraySchema.load(uri, ctx=ctx)
+    if not schema:
+        raise Exception("Unable to load tiledb ArraySchema from URI: '{}'".format(uri))
+
+    if schema.sparse:
+        return tiledb.SparseArray(uri, mode=mode, key=key, attr=attr, ctx=ctx)
+    elif not schema.sparse:
+        return tiledb.DenseArray(uri, mode=mode, key=key, attr=attr, ctx=ctx)
+    else:
+        raise Exception("Unknown TileDB array type")
+
+
+def save(uri, array, config=None, **kw):
+    if not isinstance(array, np.ndarray):
+        raise ValueError("expected NumPy ndarray, not '{}'".format(type(array)))
+    if config:
+        cfg = Config(config)
+        ctx = tiledb.Ctx(cfg)
+    else:
+        ctx = default_ctx()
+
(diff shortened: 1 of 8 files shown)


tiledb/libtiledb.pyx
```diff
@@ -5044,6 +5043,15 @@ cdef class FileHandle(object):
 
     def __dealloc__(self):
         cdef Ctx ctx = self.vfs.ctx
+        cdef int rc = TILEDB_OK
+        cdef int isclosed = 0
+        check_error(ctx,
+                    tiledb_vfs_fh_is_closed(ctx.ptr, self.ptr, &isclosed))
+        if not isclosed:
+            with nogil:
```
ihnorton: @jakebolewski

jakebolewski: This seems like the behavior we want, need to be careful about resource use here:
https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#controlling-deallocation-and-garbage-collection-in-cpython

tiledb/highlevel.py
```diff
@@ -0,0 +1,69 @@
+import tiledb
+from tiledb.libtiledb import *
+
+import numpy as np
+
+import pdb
```
jakebolewski: debugging import

tiledb/libtiledb.pyx
```diff
@@ -3312,7 +3353,7 @@ cdef class Array(object):
             _raise_ctx_err(ctx_ptr, rc)
         return
 
-    def __init__(self, uri, mode='r', key=None, timestamp=None, Ctx ctx=default_ctx()):
+    def __init__(self, uri, mode='r', key=None, timestamp=None, attr=None, Ctx ctx=default_ctx()):
```
jakebolewski: do we want to generalize this to multiple attributes (one or more?)

jakebolewski:
Some comments, otherwise it lgtm

tiledb/libtiledb.pyx
ihnorton: I could  add that and stub out the implementation for now... But we can provide the same thing semantically with queries. My reasoning for single-attribute views is to let a multi-attr object act like a regular numpy array.

jakebolewski: If we want to push the notion of arrays as views over some subset or all of the array it seems like it would generalize better.

ihnorton: (discussed out-of-band: decision: go with current API for now and iterate in the future based on feedback)

jakebolewski:
🎉

PR #134, opened 2019-04-05
Mac10.14 out-of-source fix and minor docs clean-up

ihnorton:
- fixes building against out-of-source libtiledb on newer macOS: previously the deployment target override only applied when building libtiledb from source w/in setup.py, but the cython build also needs this setting.
- minor doc fixes/clean-ups

### DIFF
--- a/doc/source/conda-env.yml
+++ b/doc/source/conda-env.yml
@@ -3,6 +3,6 @@ channels:
   - conda-forge
   - defaults
 dependencies:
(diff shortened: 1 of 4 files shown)


ihnorton:
@jakebolewski do you have any more visibility into what readthedocs is doing as an admin? I see errored build attempts, but the build logs don't load for me (spinning wheel). Some things here that might help...

PR #136, opened 2019-04-06
fix readthedocs

ihnorton:
Trying to get RTD to pick up the branch so I can initiate a test build.


ihnorton:
Successful build here: https://docs.tiledb.io/projects/tiledb-py/en/ihn-fix_docs/

PR #142, opened 2019-04-17
Added support for numpy complex types

normanb:
Added support and tests for numpy complex data types.

There is an assumption that array sizes that are equal to two and are of type float 32/64 are to be interpreted as complex data types. This seems reasonable and follows numpy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,19 +29,23 @@ script:
   - echo "Installing wheel file from $whl_file..."
   - pip install $whl_file
   - python -c "import tiledb ; tiledb.libtiledb.version()"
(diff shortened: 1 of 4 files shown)


ihnorton:
Just noting here that this change causes interpretation of 2-cell `float32` or `float64` attributes as `np.complex64` and `np.complex128` respectively. This is definitely worthwhile, in order to have good support for complex datatypes; revisit if complex datatypes are added in core.

Other 2-cell numeric or string types should round-trip as-is (tests added for 2-cell int16 and '|S2').

ihnorton:
Thanks @normanb 👍

PR #148, opened 2019-04-23
Fix issues w/ schema_like_numpy ctx; VFS config initialization

ihnorton:


tiledb/libtiledb.pyx
```diff
@@ -5202,7 +5202,7 @@ cdef class VFS(object):
             tiledb_vfs_free(&self.ptr)
 
     def __init__(self, config=None, Ctx ctx=default_ctx()):
-        cdef Config _config = Config()
+        cdef Config _config = Config(ctx.config())
```
ihnorton: @jakebolewski it seems like the semantics you wanted here is to be able to initialize a VFS object with different config parameters. However, currently the VFS object does not use the context's config at all. So, this change
- copies the existing config, and updates it, *if a dict was passed as `config=dict`.
- if a separate `Config` object is passed, then that one is used

jakebolewski: 👍 , ideally we would fix this in core TileDB but that is a pretty extensive change given the way the internal config object works.

PR #151, opened 2019-05-10
Fix memory leak

### LINKED ISSUE #150: Reading dense array doesn't free memory
Hi,

I'm wondering if this is expected behavior or if you have any tips to fix. On Ubuntu 16, Python 3.7, and  _tiledb_ 0.4.1:

Create toy array:
```
x = np.ones(10000000)
ctx = tiledb.Ctx()
path = 'test_tile_db'
d1 = tiledb.Dim(
    'test_domain', domain=(0, x.shape[0] - 1), tile=10000, dtype="uint32"
)
domain = tiledb.Domain(d1)
v = tiledb.Attr(
    'test_value',
    dtype="float32",
)
schema = tiledb.ArraySchema(
    domain=domain, attrs=(v,), cell_order="row-major", tile_order="row-major"
)
A = tiledb.DenseArray.create(path, schema)
values = x.astype(np.float32)
with tiledb.DenseArray(path, mode="w", ctx=ctx) as A:
    A[:] = {'test_value': values}
```

Read from array:
```
for i in range(10):
    with tiledb.DenseArray(path, mode='r') as data:
        data[:]
    print('Gigs:', round(psutil.virtual_memory().used / (10**9), 2))
Gigs: 0.84
Gigs: 0.89
Gigs: 0.93
Gigs: 0.97
Gigs: 1.01
Gigs: 1.05
Gigs: 1.1
Gigs: 1.14
Gigs: 1.18
Gigs: 1.22
```

Basically memory never seems to get released even when I don't assign the `data[:]` to any variable. I've tried playing around with garbage collection (`import gc; gc.collect()`) but it seems Python is not aware. Have also tried doing some explicit closing of the DenseArray. Eventually have to reset Jupyter notebook to get memory to free. 

In my real use case I am iterating over several tileDBs and pulling full array data out from each, doing some transforms, and writing new tileDBs with transformed data. Works okay except every read call adds around 2GBs to the used memory and never releases it, causing the machine to eventually run out of memory. Current work around is to spin up new process for every iteration.

Thanks!

ihnorton:
`PyArray_NewFromDescr` ignores the `NPY_ARRAY_OWNDATA` flag, so we
need to set it manually so that NumPy frees the memory.

Fixes #150

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ steps:
     architecture: 'x64'
 
 - script: |
-    python -m pip install --upgrade pip setuptools wheel numpy tox setuptools-scm cython
+    python -m pip install --upgrade pip setuptools wheel numpy tox setuptools-scm cython psutil
   displayName: 'Install dependencies'
 
 - script: |
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ numpy>=1.7.2
 setuptools>=18.0.1
 setuptools-scm>=1.5.4
 wheel>=0.30.0
+psutil
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -6,3 +6,4 @@ setuptools==40.8.0
 setuptools-scm==1.5.4
 wheel==0.30.0
 tox==3.0.0
+psutil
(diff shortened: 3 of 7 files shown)


ihnorton:
All issues should be resolved on linux/travis. I added a basic sanity check based on #150: read 40MB 100x, and check that RSS increase stays under 10%.

ihnorton:
Targeting 1.5.1 to get a CI run against that release.

PR #158, opened 2019-06-10
Schema serialization support

ihnorton:
(Also switch core target back to `dev`)

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -3309,22 +3309,41 @@ cdef class ArraySchema(object):
         else:
             raise Exception("Capnp unimplemented")
 
-        cdef char* out = NULL
-        cdef uint64_t out_len = 0
         cdef int rc = TILEDB_OK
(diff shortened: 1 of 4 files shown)


tiledb/libtiledb.pyx
```diff
@@ -3301,6 +3301,43 @@ cdef class ArraySchema(object):
...
+        """Returns a JSON (as string) representation of the array schema"""
+        cdef tiledb_serialization_type_t serialization_type
+        if format == 'JSON':
+            serialization_type = TILEDB_JSON
+        else:
+            raise Exception("Capnp unimplemented")
+
+
+        cdef tiledb_buffer_t* buff
+        check_error(self.ctx, tiledb_buffer_alloc(self.ctx.ptr, &buff))
```
tdenniston: Not necessary to alloc a buffer here -- https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/c_api/tiledb.cc#L5077 does it automatically

ihnorton:
will reopen if needed

PR #160, opened 2019-06-14
Support tiledb as namespace package

ihnorton:


ihnorton:
Python 3.4 install failure is unrelated.

PR #180, opened 2019-07-10
Add dask test for mapping blocks

normanb:
This is a WIP. I will add tests as I develop code with Dask.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,7 @@ steps:
   displayName: 'Print env'
 
 - script: |
-    python -m pip install --upgrade pip setuptools wheel numpy tox setuptools-scm cython psutil
+    python -m pip install --upgrade pip setuptools wheel numpy tox setuptools-scm cython psutil dask
   displayName: 'Install dependencies'
 
 - script: |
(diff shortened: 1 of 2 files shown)


ihnorton:
@normanb I pushed a commit updating the tests here to pass. Tested with both the default (threaded) scheduler and `distributed` as below (not sure how to parameterize unittests so didn't commit this yet):

```
commit ffd0c78fb791939d3af86c60490e0cdd84cf9114 (HEAD -> ncb/dask_blocks)
Author: Isaiah Norton <isaiah@tiledb.io>
Date:   Wed Jul 31 16:08:10 2019 -0400

    Example test with distributed

diff --git a/tiledb/tests/test_dask.py b/tiledb/tests/test_dask.py
index ebe3a68..aee9314 100644
--- a/tiledb/tests/test_dask.py
+++ b/tiledb/tests/test_dask.py
@@ -20,6 +20,12 @@ class DaskSupport(DiskTestCase):
         else:
             super().setUp()

+        from dask.distributed import Client
+        self._client = Client(n_workers=4)
+
+    def tearDown(self):
+        self._client.close()
+
     @unittest.expectedFailure
     def test_dask_from_numpy_1d(self):
         uri = self.path("np_1attr")
```

ihnorton:
Also pushed a commit installing dask on CI, and removing the x-fail designation from those tests.

PR #181, opened 2019-07-10
Fix PyPI wheel build for TileDB 1.6

ihnorton:
- AWS SDK and TileDB core have a number of default constructors which cause
  'unused parameter' warnings under GCC 4.8. I have manually run unit tests
  against Minio with a GCC 4.8 build, and all tests passed.

- ~~disable ~~HDFS and CPP_API, which don't make sense to build for the wheel.~~

### DIFF
--- a/misc/pypi_linux/Dockerfile
+++ b/misc/pypi_linux/Dockerfile
@@ -21,9 +21,14 @@ ENV CMAKE /opt/python/cp27-cp27mu/bin/cmake
 
 ###############################################
 # settings (B)
-ENV TILEDB_VERSION 1.5.0
-ENV TILEDB_PY_VERSION 0.4.0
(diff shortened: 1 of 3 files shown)


jakebolewski:
does building with HDFS cause an issue?  it includes the header in tree and does a dynamic dll lookup at runtime

ihnorton:
I hit a build error when building w/ tests, in order to make sure everything worked under the GCC 4.8 build. But I guess that was due to re-running cmake, so it doesn't really matter.

I don't have an objection to shipping it, just not clear to me whether it would actually work.

ihnorton:
I put it back it, doesn't cost anything to leave it there and maybe it will be useful or at least tested. Will update this branch when I merge my core PRs.

PR #185, opened 2019-07-15
{Dense/Sparse}Array.create must exclude mismatched schema, fixes #145

### LINKED ISSUE #145: SparseArray constructor should not accept dense schema
This should raise an error:
```
import tiledb, numpy as np, os
ctx = tiledb.Ctx()
dims = (tiledb.Dim(ctx=ctx, domain=(0, 6), tile=2),)
dom = tiledb.Domain(*dims, ctx=ctx)
att = tiledb.Attr(ctx=ctx, dtype=np.byte)
schema = tiledb.ArraySchema(ctx=ctx, domain=dom, attrs=(att,))
tiledb.SparseArray.create('/tmp/s1', schema)
with tiledb.DenseArray('/tmp/s1') as T:
  print(T[:])

[0 0 0 0 0 0 0]

schema.dump()

- Array type: dense
```

ihnorton:


jakebolewski:
I don't think this would work properly for subclasses?  ex. `class MyDenseArray(tiledb.DenseArray)` ...

ihnorton:
Good point, thanks. Switched to `issubclass` and added a test for the usual cases and a sparse subclass.

PR #202, opened 2019-09-26
Add `domain_index` operator

### LINKED ISSUE #102: Support negative domain slices in queries
TileDB: `1.4.0`
TileDB-Py: `0.3.0`

I'm investigating using TileDB(-Py) to store the following sparse geospatial data:
- 2.5D (from GeoTIFFs): latitude, longitude, elevation
- 3D (from LAS point clouds): latitude, longitude, elevation, intensity

These types of domains inherently depend on negative, real domain values. [TileDB documentation](https://docs.tiledb.io/en/stable/tutorials/neg-real.html) indicates negative and real domain values are supported, which is encouraging.

The support/use of negative domain values also implies one might need the ability to use slices with negative values when performing queries. Unfortunately, it seems queries with negative domain values are still unsupported. Attempting to use negative domain values in slices results in an `IndexError: index out of bounds <todo>`, thrown [here](https://github.com/TileDB-Inc/TileDB-Py/blob/0.3.0/tiledb/libtiledb.pyx#L2873).

I've tried to mitigate the issue by setting the domain dimensions to be non-negative (latitude: 0-180; longitude: 0-360) and using conversion methods to ensure domain values are non-negative when writing, then converting to their real values for queries or return query results. The following would be examples of the aforementioned conversion methods:
```py
def to_domain_latitude(real_latitude):
    '''Convert real latitude to domain latitude'''
    return real_latitude + 180.0

def to_real_latitude(domain_latitude):
    '''Convert domain latitude to real latitude'''
    return domain_latitude - 180.0
```
However, it would be simplest for us developers to store the domain values in their native ranges/formats.

### LINKED ISSUE #201: Querying data with negative real indecies
Hello there,

I'm working with a TileDB containing to Domains with negative values. When querying values I get:

```
tiledb.libtiledb.TileDBError: [TileDB::Query] Error: Subarray lower bound is larger than upper bound. subarray: [86.75, 86.5, 26.75, 26.75, 157498, 157498] domain: [-89.75, 89.75, -179.75, 179.75, 157498, 157863]
```

My query looks as follows:
```
data = self.data.query(attrs=["some_data"], coords=True)[-3.25,26.75,157498.4375]
```

Also the error message does not display the subarray I define in my query.

ihnorton:
Add `Array.domain_index` operator, which supports slicing over any range within the domain bounds, including negative slices (previously limited for the Array[] operator by NumPy wrap-around semantics). Slice ranges are passed down to libtiledb core for range and point queries.

Fixes #102
Fixes #201

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,6 @@ install:
 script:
   # Build and test extension module
   - python setup.py build_ext --inplace
-  # Run doctests
-  - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then (LD_LIBRARY_PATH="$TRAVIS_BUILD_DIR/tiledb:$LD_LIBRARY_PATH" python -m doctest -o NORMALIZE_WHITESPACE -f tiledb/libtiledb.pyx); fi
-  # Run unittests
-  - export LD_LIBRARY_PATH="$TRAVIS_BUILD_DIR/tiledb:$LD_LIBRARY_PATH"
-  - python -m unittest tiledb.tests.all.suite_test
   # Check wheel build
   - python setup.py bdist_wheel
   - whl_file=`pwd`/dist/`ls dist`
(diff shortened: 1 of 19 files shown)


ihnorton:
> Is there any reason this can't be supported (in the future) for dense arrays with negative domains?

Shouldn't be a problem, I will add in a follow-up.

PR #225, opened 2019-11-14
Update ReadTheDocs API documentation.

tdenniston:
This will need to be released at some point soon so that the RTD "stable" docs (which build from the most recent tag) are rerendered.


tdenniston:
Is the Travis failure because of a change in this PR? Not sure how my changes here would have caused that.

ihnorton:
Should be fixed by the curl PR to core. I think we can go ahead and merge this.

PR #240, opened 2019-12-05
Add tiledb.Array.load_typed for typed open w/out separate schema check

ihnorton:
Adds a factory-like static method, `Array.load_typed`, which returns
a DenseArray or SparseArray instance, depending on the type of the
underlying array. Eliminates the current schema check before class
allocation, which adds latency in non-local usage.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -2917,6 +2917,67 @@ cdef class ArraySchema(object):
         print("\n")
         return
 
+# Wrapper class to allow returning a Python object so that exceptions work correctly
+# within preload_array
+cdef class ArrayPtr(object):
+    cdef tiledb_array_t* ptr
+
+cdef ArrayPtr preload_array(uri, mode, key, timestamp, Ctx ctx=default_ctx()):
+    # ctx
+    cdef tiledb_ctx_t* ctx_ptr = ctx.ptr
+    # uri
+    cdef bytes buri = unicode_path(uri)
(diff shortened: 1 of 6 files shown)


tiledb/libtiledb.pyx
```diff
@@ -2991,57 +3052,64 @@ cdef class Array(object):
...
-        # check the key, and convert the key to bytes
-        if key is not None:
-            if isinstance(key, str):
-                bkey = key.encode('ascii')
+        cdef tiledb_array_schema_t* schema_ptr
+        cdef tiledb_array_type_t array_type
+        cdef Array new_array
+        cdef object new_array_typed
+
+        # *** this call frees array_ptr and raises a Python exception ***
```
joe-maley: Can you explain this in a little more detail? I'm not sure where the exception is coming from and where it is caught.

ihnorton: Nice catch, thanks. This comment is outdated. Initially, this function was returning a bare `tiledb_array_t*`, but Cython requires [extra care](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values) when writing functions which do not return a Python object (PyObject). So, to make the `preload` function safer, I added the `ArrayPtr` wrapper class. Any exceptions inside the `preload` function should be properly raised, without needing to worry about freeing `array_ptr`. (after this call returns, the `try..except` block is responsible for freeing `array_ptr` if there are any exceptions while retrieving the schema info)

I will update the comment to reflect the above discussion.

PR #256, opened 2020-01-03
Default ctx initialization improvements, and custom buffer size

ihnorton:
- initialize `default_ctx` lazily, allowing user control over global per-process settings when necessary (e.g. `sm.num_tbb_threads`)
  - `default_ctx` may be initialized with custom settings by calling `tiledb.libtiledb.initialize_ctx()`
  - TBB uses the thread count set at the *first* pool creation, so a call to `tiledb.Ctx(...)` should specify number of threads (if needed) whenever called before an implicit call to `default_ctx()` (unlikely)
- make `default_ctx` a delayed call in all method definitions
- add support for user-controlled initial buffer sizes for queries with unknown result sizes, using new `py.init_buffer_size` parameter

### DIFF
--- a/tiledb/tests/test_libtiledb.py
+++ b/tiledb/tests/test_libtiledb.py
@@ -1155,8 +1155,12 @@ class DenseArrayTest(DiskTestCase):
 
         # create context with 1 MB memory budget (2 MB total, 1 MB usable)
         config = tiledb.Config({'sm.memory_budget': 2 * 1024**2,
-                                'py.init_buffer_size': 1024**2 })
+                                'py.init_buffer_bytes': 1024**2 })
         ctx = tiledb.Ctx(config=config)
+        self.assertEqual(
+            config['py.init_buffer_bytes'],
+            str(1024**2)
+        )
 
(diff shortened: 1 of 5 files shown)


tiledb/tests/test_libtiledb.py
```diff
@@ -1160,9 +1153,14 @@ def test_incomplete_dense(self):
         with tiledb.from_numpy(path, data, tile=use_tile) as A:
             pass
 
-        # create context with 1 MB memory budget
-        config = tiledb.Config({'sm.memory_budget': 2 * 1024**2})
+        # create context with 1 MB memory budget (2 MB total, 1 MB usable)
+        config = tiledb.Config({'sm.memory_budget': 2 * 1024**2,
+                                'py.init_buffer_bytes': 1024**2 })
```
ihnorton: @stavrospapadopoulos do you foresee any problem with putting a custom value in the Config object, in order to maintain a consistent API? Seems to work... (alternatively, could have a separate `py_config` object)

ihnorton:
> py.init_buffer_bytes is per buffer (attr/coords) right?

yes

PR #258, opened 2020-01-09
Add wheel builder scripts for macOS

### LINKED ISSUE #144: PyPI wheels for macOS
Currently we distribute wheels for `manylinux`. It'd be nice to also distribute wheels for macOS.

ihnorton:
Adds a wheel-build script running on azure pipelines and utilizing cibuildwheel. Currently this is run for macOS only, because the `misc/pypi_linux` Docker setup needs some reorganization before I can run it in a similar way.

Fixes #144 

macOS wheels for 0.5.4 were built with this method and uploaded to PyPI: https://pypi.org/project/tiledb/#files (uploaded manually for now, until the Linux side is added).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,4 +1,4 @@
-jobs:
+stages:
   - template: ./misc/azure-ci.yml
-  #- template: ./misc/azure-release.yml
+  - template: ./misc/azure-release.yml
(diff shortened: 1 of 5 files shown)


ihnorton:
build error will be fixed by https://github.com/TileDB-Inc/TileDB/pull/1474

PR #291, opened 2020-03-24
Fix #290

### LINKED ISSUE #290: zero length unicode string causes assertion failure
tiledb version 0.5.6

attempting to write zero-length unicode array triggers assertion

test case:
```
attrs = [tiledb.Attr(name="foo", dtype=np.unicode)]
domain = tiledb.Domain(tiledb.Dim(domain=(0, 1000), tile=1000))
schema = tiledb.ArraySchema(domain=domain, sparse=False, attrs=attrs)
tiledb.DenseArray.create("foo", schema)

with tiledb.DenseArray("foo", mode="w") as A:
    # change empty string to non-empty and things work
    val = np.array(['' for i in range(0, 1001)], dtype=np.unicode)
    A[:] = val
```

result:
```
# python foo.py
  File "foo.py", line 61, in <module>
    A[:] = np.zeros((1000,), dtype=np.unicode)
  File "tiledb/libtiledb.pyx", line 4175, in tiledb.libtiledb.DenseArrayImpl.__setitem__
  File "tiledb/libtiledb.pyx", line 274, in tiledb.libtiledb._write_array
  File "tiledb/np2buf.pyx", line 108, in tiledb.libtiledb.array_to_buffer
AssertionError
```

ihnorton:
Fixes for #290 and related issue with metadata.

### DIFF
--- a/tiledb/libmetadata.pyx
+++ b/tiledb/libmetadata.pyx
@@ -264,7 +264,12 @@ cdef object get_metadata(array: Array,
     if (value == NULL):
         if value_num == 1:
             # in this case, the key exists with empty value
-            return ()
+            if value_type == TILEDB_CHAR:
+                return b''
+            elif value_type == TILEDB_STRING_UTF8:
+                return u''
+            else:
+                return ()
         raise KeyError(key)
 
     return unpack_metadata_val(value_type, value_num, value)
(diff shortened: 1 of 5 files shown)


ihnorton:
I did a test build against TileDB 1.7.6 which succeeded, will merge this in order to backport.

CI failed due to some unrelated changes in TileDB `dev` which are not updated in TileDB-Py yet.

ihnorton:
https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=3062&view=results

PR #297, opened 2020-04-01
Add test for attribute/dim filename character set

ihnorton:


ihnorton:
Superseded by https://github.com/TileDB-Inc/TileDB/pull/2107

PR #308, opened 2020-05-12
Set start/end_buf_ptr for var size nonEmptyDomain

Shelnutt2:
This fixes a case where for dimensions which are fixed length and end_buf_ptr is not set resulting in:

`nboundLocalError: local variable 'end_buf' referenced before assignment`


Shelnutt2:
@ihnorton I'm not sure about the OSX failure here. Is that a transient failure?

ihnorton:
I saw that once before, but couldn't reproduce it locally or on CI. Will dig in a bit more.

ihnorton:
@Shelnutt2 I couldn't reproduce, going to make the failure into a warning up to 3x per attempt, and see if it is repeatable within a given run. See https://github.com/TileDB-Inc/TileDB-Py/pull/310/commits/981719e1ba174a19497996395651524e4387cce4

PR #319, opened 2020-05-28
Add doc/doctest for tiledb.vacuum, and expand test section

ihnorton:


tiledb/tests/test_libtiledb.py
```diff
@@ -2730,21 +2730,42 @@ def test_array_vacuum(self):
...
+        path2 = self.path("test_array_vacuum_fragment_meta")
+        create_array(path2)
+        write_fragments(path2)
+        tiledb.consolidate(path2,
+                           config=tiledb.Config({'sm.consolidation.mode': 'fragment_meta'}))
+        tiledb.vacuum(path2,
+                      config=tiledb.Config({'sm.vacuum.mode': 'fragment_meta'}))
+        paths = vfs.ls(path2)
+
+        self.assertEqual(len(paths), 3 + 2 * num_writes + 1)
```
ihnorton, line 2768: @stavrospapadopoulos double-checking if this is correct that there should only be one additional `.meta` file after `fragment_meta` consolidation?

PR #360, opened 2020-07-01
Add support for from_csv 'fillna' argument

ihnorton:
The argument will be applied via pandas.DataFrame.fillna, before
any values are passed in to the TileDB writer. ([example](https://github.com/TileDB-Inc/TileDB-Py/compare/ihn/fillna?expand=1#diff-970d04aaca2ebc4aaa52a1eadbf53db9R488))

### DIFF
--- a/tiledb/tests/test_pandas_dataframe.py
+++ b/tiledb/tests/test_pandas_dataframe.py
@@ -470,3 +470,26 @@ class PandasDataFrameRoundtrip(DiskTestCase):
             df_bk = pd.DataFrame(res)
 
             tm.assert_frame_equal(df_bk, df)
+
+    def test_csv_fillna(self):
+        col_size = 10
+        data = np.random.rand(10)
+        data[4] = np.nan
+        df = pd.DataFrame({'v': data})
+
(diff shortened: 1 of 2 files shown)


tiledb/tests/test_pandas_dataframe.py
```diff
@@ -469,4 +469,27 @@ def test_csv_chunked(self):
...
+        tmp_dir = self.path("csv_fillna")
+        os.mkdir(tmp_dir)
+        tmp_csv = os.path.join(tmp_dir, "generated.csv")
+
+        df.to_csv(tmp_csv, index=False)
+
+        tmp_array = os.path.join(tmp_dir, "array")
+        tiledb.from_csv(tmp_array, tmp_csv, fillna={'v': 0})
+
+        df['v'][4] = 0
```
normanb, line 489: Should this be an assert?

ihnorton: No, this updates the value to what is expected on read-back because of the fill. Otherwise the `assert_frame_equal` will fail.

PR #365, opened 2020-07-14
Add support for multi-file CSV ingestion

ihnorton:
This PR implements support for `tiledb.from_csv` to accept a list of CSV files as the first input argument, instead of a single filename. This ingestion mode requires the `chunksize` argument, which will determine the minimum number of rows to read before flushing to disk.


normanb:
:+1:

PR #388, opened 2020-10-07
Add MD5 and SHA256 filter support

Shelnutt2:
Add MD5 and SHA256 filter support


Shelnutt2:
Closing in favor of #389

PR #396, opened 2020-10-19
WIP Support running unit tests against different VFS

ihnorton:
@Shelnutt2 this is roughly what we need to do in order to run the TileDB-Py unit tests against a VFS rather than a local filesystem. Currently it is hard-coded against S3/minio, but can refactor to environment variables or something else depending on how we decide to target the REST system. 

(currently a couple failures against S3 that I haven't tracked down; some missing `ctx=` arguments, and a couple of multipart upload failures when flushing that I believe are spurious and related to minio, but haven't tracked down yet)


ihnorton:
supplanted by #580

PR #402, opened 2020-10-26
add --release-symbols option to obtain release build with symbol info…

d-hoke:
have the fetched tiledb build include symbol information by setting the cmake build type to RelWithDebInfo


ihnorton:
Thank you.

PR #403, opened 2020-10-28
Modified stats_dump() To Return Internal Stats As String

nguyenv:
Previously, tiledb_stats_dump() output to stdout. However, Jupyter notebooks
do not capture or redirect external stdout and thus were not able to display
internal stats. tiledb_stats_dump() now outputs to a temporary file which is
then converted to a Python string.

---
[ch3357]

### DIFF
--- a/tiledb/core.cc
+++ b/tiledb/core.cc
@@ -838,16 +838,19 @@ void init_stats() {
   stats_counters["py.unpack_results_time"] = TimerType();
 }
 
-void print_stats() {
+std::string python_internal_stats() {
   if (!g_stats) {
-    TPY_ERROR_LOC("Stats counters are not unitialized!")
+    TPY_ERROR_LOC("Stats counters are not uninitialized!")
   }
 
   auto counters = g_stats.get()->counters;
 
-  std::cout << "==== Python ====" << std::endl <<
-               "- Read query time: " << counters["py.read_query_time"].count() << std::endl <<
-               "- Buffer conversion time: " << counters["py.unpack_results_time"].count() << std::endl;
+  std::ostringstream os;
+  os << "==== Python ====" << std::endl <<
+      "- Read query time: " << counters["py.read_query_time"].count() << std::endl <<
+      "- Buffer conversion time: " << counters["py.unpack_results_time"].count();
+
+  return os.str();
 }
 
 PYBIND11_MODULE(core, m) {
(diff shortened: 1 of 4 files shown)


tiledb/libtiledb.pyx
```diff
@@ -504,13 +506,23 @@ def stats_reset():
...
 
 def stats_dump():
-    """Print all TileDB internal statistics values to standard output."""
-    tiledb_stats_dump(stdout)
+    """Return TileDB internal statistics as a string."""
+    fd, path = tempfile.mkstemp()
+    stats = ""
+    try:
+        dump_file = fdopen(fd, "w")
+        if tiledb_stats_dump(dump_file) == TILEDB_ERR:
```
ihnorton, line 514: Would you mind to try using `tiledb_stats_dump_str` here, see
https://github.com/TileDB-Inc/TileDB/blob/8f0e60a2713abadd68ba7f7e86e0850b93d1b12c/tiledb/sm/c_api/tiledb.h#L5580-L5596

This will look something like:
```
    cdef char* stats_str_ptr = NULL
    if tiledb_stats_dump_str(&stats_str_ptr) == TILEDB_ERR:
      raise TileDBError("Failed to dump internal stats")

    stats_str = stats_str_ptr.decode('UTF-8', 'strict')
    free(stats_str_ptr)

    return stats_str
```

tiledb/libtiledb.pyx
```diff
@@ -504,13 +506,23 @@ def stats_reset():
...
+            print("Unable to dump stats to file.")
+        fclose(dump_file)
 
-    import tiledb.core
-    tiledb.core.print_stats()
```
ihnorton, line 511: This line prints the python-side statistics, so we need to change this function: https://github.com/TileDB-Inc/TileDB-Py/blob/0d18acdad410e90465d2d5a5fa701d77e58f29d6/tiledb/core.cc#L841-L851

to build and return a `std::string`, which pybind11 which then return as a python string.

Then we can append the python stats for the final string.

ihnorton (changes requested):
Thanks! Some small change suggestions to make sure we still get the stats from the python side.

tiledb/libtiledb.pyx
nguyenv: Should I print or return `stats_str`? In the comment above, you we should print the output, but here you are returning the output.

ihnorton: Still print at the end, just meant to highlight how to get the data back from the `_str` version.

tiledb/core.cc
```diff
@@ -838,16 +838,19 @@ void init_stats() {
   stats_counters["py.unpack_results_time"] = TimerType();
 }
 
-void print_stats() {
+std::string python_internal_stats() {
   if (!g_stats) {
     TPY_ERROR_LOC("Stats counters are not unitialized!")
```
ihnorton, line 843: noticed pre-existing typo here, do you mind to fix?

ihnorton (approved):
Two small tweaks and then LGTM.

ihnorton (approved):
Thank you! 💯

PR #405, opened 2020-10-29
Make 'rows' dimension transparent in dataframe read-back

ihnorton:
This change hides the 'rows' name we use for generated schema dimension name when no index is specified. A read-back dataframe using `.df[]` or `open_dataframe` should look the same as a direct `pd.read_csv` of the source data. See before/after here: https://github.com/TileDB-Inc/TileDB-Py/compare/ihn/df_readback_rows_x?expand=1#diff-1e19c4fe9825fc13a5d649c9d7745a2d49f2c9be3a247f7fa9b0526fc7ac5142L4131-R4138

By request, and also fixes a bug @eddelbuettel noticed running a prior notebook (plus improved reporting for that error by throwing instead of giving a context-free `assert`ion).

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -231,8 +231,11 @@ def dim_for_column(ctx, name, dim_info, col, tile=None, full_domain=False, ndim=
 def get_index_metadata(dataframe):
     md = dict()
     for index in dataframe.index.names:
+        index_md_name = index
+        if index == None:
+            index_md_name = 'rows'
         # Note: this may be expensive.
-        md[index] = dtype_from_column(dataframe.index.get_level_values(index)).dtype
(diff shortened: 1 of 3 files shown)


eddelbuettel:
I can't say I follow the Python code fully.  Why does renaming to `index_md_name` hide from printing? Is that some Pandas fu?

ihnorton:
> Why does renaming to index_md_name hide from printing? Is that some Pandas fu?

Just bookkeeping. We have to translate the TileDB dimension identifier into an index with name `None` (the python type) in order to match the pandas default when it reads rows w/out a specified index.

eddelbuettel (approved):
LGTM (with the caveat that I am not too deep into the Python nuts and bolts)

PR #413, opened 2020-11-12
Add __repr__ For Ctx and Array

nguyenv:
* Array shows URI, mode, and ndim
* Ctx returns the same repr as Config

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -992,6 +992,9 @@ cdef class Ctx(object):
         self.ptr = ctx_ptr
         self._set_default_tags()
 
+    def __repr__(self):
+        return repr(self.config())
+
     def config(self):
         """Returns the Config instance associated with the Ctx."""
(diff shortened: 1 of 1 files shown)


tiledb/libtiledb.pyx
```diff
@@ -3649,6 +3652,10 @@ cdef class Array(object):
         cap = PyCapsule_New(<void *>(self.ptr), name, NULL)
         return cap
 
+    def __repr__(self):
+        return "Array(uri={0}, mode={1}, ndim={2})"\
```
ihnorton, line 3656: Please add single-quote around the uri.

ihnorton (approved):
One minor comment. LGTM, thanks.

tiledb/libtiledb.pyx
nguyenv: OK. Using `uri={0!r}` which will include single-quotes and is consistent with other parts of the code.

PR #422, opened 2020-11-20
Fix CSV/dataframe ingest error due to setting tile=0 with small index range

ihnorton:
Fixes bug where we set tile=0 when ingesting a CSV with only 1 row,
or an indexed dataframe with index range of 0. Leading to tile extent check failure in core.


ihnorton:
Cherry-picked to #423 to reduce CI.

PR #439, opened 2020-12-08
Addition of Github Actions for CI and Release

nguyenv:


nguyenv:
https://github.com/TileDB-Inc/TileDB-Py/actions/runs/407223549

ihnorton (commented):
I'm a little on the fence about running the tagging through here until we have the full release path setup. Do you know if it will update the tag every time the branch is updated? I don't typically tag until I've tested the wheels.

Otherwise LGTM. Thanks.

nguyenv:
I can remove the `tag_name` line or even just shelve this whole Release action file altogether. The reason the tag and release got erroneously updated a few days ago is because I had set up the Release file to run every time someone pushes to any branch. Currently it updates every time someone pushes to a `release-*` branch.

ihnorton:
I was hoping there was a way we could trigger this action from the end of the PublishArtifacts step in the existing release pipeline, which would allow us to test the wheels in an isolated environment before releasing (ideally with some check of S3 functionality too). But triggering that direction looks [somewhat involved](https://medium.com/swlh/trigger-github-actions-with-azure-devops-pipelines-4f0791e6cbbf).

So I think we should comment out the release part for now and revisit, I don't really want to have a separate process creating tags that may not correspond to what is actually tested+released.

ihnorton (approved):

PR #446, opened 2020-12-11
Sync changes from release-0.7.4

ihnorton:
Syncing from [0.7.4](https://github.com/TileDB-Inc/TileDB-Py/tree/release-0.7.4)
- bug fix for Dim constructor found during testing the release
- HISTORY.md sync
- a compatibility fix for 2.1.3 to simplify cherry-picking when we do a 0.7.5 (the FragmentInfo changes are self-contained other than setup.py, so they are easy to revert on the release branch)

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -2584,7 +2584,6 @@ cdef class Dim(object):
                 check_error(ctx,
                     tiledb_dimension_set_filter_list(ctx.ptr, dim_ptr, filter_list_ptr))
         except:
-            tiledb_dimension_free(&dim_ptr)
             raise
 
         self.ctx = ctx
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,7 @@
 # In Progress
 
+# TileDB-Py 0.7.4 Release Notes
+
 ## Improvements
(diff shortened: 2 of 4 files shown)


tiledb/libtiledb.pyx
```diff
@@ -2584,7 +2584,6 @@ cdef class Dim(object):
                 check_error(ctx,
                     tiledb_dimension_set_filter_list(ctx.ptr, dim_ptr, filter_list_ptr))
         except:
-            tiledb_dimension_free(&dim_ptr)
```
nguyenv, line 2587: Does this get freed somewhere else now?

ihnorton: It is freed in `tiledb_dimension_alloc` on error. Initially when I implemented this, I was checking whether `dim_ptr` was null before freeing, but that doesn't work b/c of https://github.com/TileDB-Inc/TileDB/pull/1959. So we don't need to free at all here.

ihnorton: see https://github.com/TileDB-Inc/TileDB/blob/97d400117232fbaf396398ce02637541f99ceb05/tiledb/sm/c_api/tiledb.cc#L1793-L1817 (before my commint in 1959)

nguyenv (approved):

PR #447, opened 2020-12-11
Support Integer Multi-Indexing For Dimensions of Type Numpy Datetime64

nguyenv:


Shelnutt2 (approved):
Thanks for making the change!

ihnorton (approved):
LGTM other than windows test. Please add a comment about why this is supported too.

ihnorton (approved):

PR #452, opened 2020-12-16
Add Verbosity Flag To stats_dump()

nguyenv:
* By default, stats_dump() will display top-level information that is
  most relevent to end users.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -503,7 +503,7 @@ def stats_reset():
     import tiledb.core
     tiledb.core.init_stats()
 
-def stats_dump(version=True, print_out=True, include_python=True):
+def stats_dump(version=True, print_out=True, include_python=True, verbose=False):
     """Return TileDB internal statistics as a string.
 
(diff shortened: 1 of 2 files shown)


tiledb/libtiledb.pyx
```diff
@@ -513,26 +513,46 @@ def stats_dump(version=True, print_out=True, include_python=True):
...
+    if not verbose:
+        import json
+        stats_json_core = json.loads(stats_str_core)
+
     if tiledb_stats_free_str(&stats_str_ptr) == TILEDB_ERR:
         raise TileDBError("Unable to free stats_str_ptr.")
 
-    stats_str = ""
+    if not verbose:
+        stats_str = "==== READ ====\n"
```
ihnorton, line 533: Suggest to print this after the version information, added below.

ihnorton: Also: please add one more, separating newline before the stats.

ihnorton (approved):
Looks great! Minor comment about order, otherwise LGTM.

PR #454, opened 2020-12-17
Addition of unique_dim_values() For SparseArrays

nguyenv:
* Returns the unique dimension values for a given sparse array.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -5429,17 +5429,22 @@ cdef class SparseArrayImpl(Array):
         return out
     
     def unique_dim_values(self, dim=None):
-        if dim and not self.domain.has_dim(dim):
+        if dim is not None and not isinstance(dim, str):
+            raise ValueError("Given Dimension {} is not a string.".format(dim))
+
+        if dim is not None and not self.domain.has_dim(dim):
             raise ValueError("Array does not contain Dimension '{}'.".format(dim))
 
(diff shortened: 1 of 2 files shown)


tiledb/libtiledb.pyx
```diff
@@ -5427,6 +5427,19 @@ cdef class SparseArrayImpl(Array):
                     out[final_name] = arr
 
         return out
+    
+    def unique_dim_values(self, dim=None):
+        if dim and not self.domain.has_dim(dim):
```
ihnorton, line 5432: Need to make sure `dim` is a string here.

tiledb/libtiledb.pyx
```diff
@@ -5427,6 +5427,19 @@ cdef class SparseArrayImpl(Array):
...
+            raise ValueError("Array does not contain Dimension '{}'.".format(dim))
+
+        query = self.query(attrs=[])[:]
+
+        if dim:
+            dim_values = np.unique(query[dim])
+        else:
+            dim_values = [tuple(np.unique(values)) for values in query.values()]
```
ihnorton, line 5440: Seems like this should be an `OrderedDict` rather than a tuple of arrays. What do you think?

nguyenv: OK that makes sense. I'll change it to be an OrderedDict of tuples.

ihnorton (approved):

PR #458, opened 2020-12-22
Query Support For Returning Subsets of Specified Dimensions

nguyenv:


tiledb/libtiledb.pyx
```diff
@@ -4409,7 +4409,9 @@ cdef class Query(object):
             for dname in dims:
                 if not domain.has_dim(dname):
                     raise TileDBError("Selected dimension does not exist: '{name}")
-            self.dims = dims
+
+            from six import u as unicode 
+            self.dims = [unicode(dname) for dname in dims]
```
ihnorton, line 4414: You can use `dname.encode("UTF-8")` instead to avoid the import. (I'm a little surprised this was necessary, I guess because it is user input so might be bytes?)

ihnorton (approved):
Minor change, otherwise LGTM.

ihnorton:
x-ref the PR implementing this for `multi_index`/`df`: https://github.com/TileDB-Inc/TileDB-Py/pull/433

tiledb/libtiledb.pyx
nguyenv: OK will do. I get [this](https://github.com/TileDB-Inc/TileDB-Py/runs/1591496316?check_suite_focus=true#step:9:425) error with Python 2.

tiledb/libtiledb.pyx
```diff
@@ -4444,7 +4444,7 @@ cdef class Query(object):
     def __getitem__(self, object selection):
         return self.array.subarray(selection,
                                    attrs=self.attrs,
-                                   coords=self.coords,
+                                   coords=self.coords if self.coords else self.dims,
```
ihnorton, line 4447: I believe this logic needs to be inverted so that `dims` takes precedence.

nguyenv: What would be a case where this would give an unexpected result?

ihnorton: `coords=True, dims=["a"]`

nguyenv: It should error out in `__init__` before then with `Cannot pass both dims and coords=True to Query`

ihnorton: Good point, ok.

PR #467, opened 2021-01-27
Preload array metadata in the dataframe query path

ihnorton:
This PR adds an asynchronous call to array.metadata_num before
submitting the query in the `.df` path.

We unconditionally check `array.meta` in the `.df` path in order
to get any type mappings stored there. This currently happens
synchronously after the query submission, which requires more
network i/o calls to get metadata, because libtiledb metadata
loading is lazy. Instead, initiate an async metadata operation
so that libtiledb fetches (and caches) the metadata ahead of
time.

### DIFF
--- a/tiledb/core.cc
+++ b/tiledb/core.cc
@@ -4,6 +4,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <future>
 #include <cstring>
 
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):
This is a nice addition, thanks for whipping it up! I've confirmed in the test case we are doing this cuts down the time by 200-400ms! That is 10-20% improvement in the read time, very nice!

PR #473, opened 2021-02-12
Fix and test multi_index bug with ranges at 0

ihnorton:
Fix bug in Array.multi_index with slice range including 0 (incorrectly used the nonempty domain as endpoint)

### DIFF
--- a/tiledb/multirange_indexing.py
+++ b/tiledb/multirange_indexing.py
@@ -47,8 +47,8 @@ def sel_to_subranges(dim_sel, nonempty_domain=None):
                 # we are missing one or both endpoints, maybe use nonempty_domain
                 if nonempty_domain is None:
                     raise TileDBError("Open-ended slicing requires a valid nonempty_domain")
-                rstart = range.start if range.start else nonempty_domain[0]
-                rend = range.stop if range.stop else nonempty_domain[1]
+                rstart =  nonempty_domain[0] if range.start is None else range.start
+                rend =  nonempty_domain[1] if range.stop is None else range.stop
 
             subranges.append( (rstart,rend) )
         elif isinstance(range, tuple):
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+# In Progress
+
+## Bug fixes
+* Fix bug in Array.multi_index with slice range including 0 (incorrectly used the nonempty domain as endpoint) [#473](https://github.com/TileDB-Inc/TileDB-Py/pull/473)
+
 # TileDB-Py 0.8.2 Release Notes
 
 ## Packaging Notes
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ from sys import version_info as ver
 print("setup.py sys.argv is: ", sys.argv)
 
 # Target branch
-TILEDB_VERSION = "2.2.3"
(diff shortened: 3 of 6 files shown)


tiledb/multirange_indexing.py
```diff
@@ -47,8 +47,8 @@ def sel_to_subranges(dim_sel, nonempty_domain=None):
                 # we are missing one or both endpoints, maybe use nonempty_domain
                 if nonempty_domain is None:
                     raise TileDBError("Open-ended slicing requires a valid nonempty_domain")
-                rstart = range.start if range.start else nonempty_domain[0]
-                rend = range.stop if range.stop else nonempty_domain[1]
+                rstart = range.start if (range.start is not None) else nonempty_domain[0]
+                rend = range.stop if (range.stop is not None) else nonempty_domain[1]
```
jp-dark, line 51:
```suggestion
                rstart = nonempty_domain[0] if range.start is None else range.start 
                rend = nonempty_domain[1] if  range.stop is None else range.stop 
```
Stylistic suggestion to make the above look slightly cleaner (feel free to ignore)

jp-dark (approved):
Looks good to me.

gsakkis (approved):
:+1:

tiledb/multirange_indexing.py
ihnorton: Done, thanks!

outdoorpet:
Hi All,

Will this fix make it into the next release? I have just noticed the errors when trying to use multi_index[slice(0, 11, None)] as an example.

It looks like this has been a bug since version 0.7.3.

Thanks!

ihnorton:
Yes, definitely - it will be included in a release this week.

outdoorpet:
Hi @ihnorton, thanks for getting the release out. I still get the error when using multi_index to query an **empty** Sparse array. The expected behaviour would be to return an empty result dictionary rather than this error.

ihnorton:
Hi @outdoorpet, I reproduced a similar issue to this one, when the dimension origin is `0` for a sparse array. I will have a fix for that soon.

Also, if you have a minimal example, that would be helpful -- to make sure your use-case is also fixed. Thanks.

outdoorpet:
Hi @ihnorton, please see below example:

```
import tiledb
import numpy as np

dom = tiledb.Domain(
    tiledb.Dim(name="x", domain=(0, 2**64-2), tile=1, dtype=np.uint64)
)

schema = tiledb.ArraySchema(
    domain=dom, sparse=True, attrs=[tiledb.Attr(name="a", dtype=np.uint64)]
)

array_name = "multi_index_sparse"

tiledb.SparseArray.create(array_name, schema)

slice_index = slice(0, 4, None)

with tiledb.SparseArray(array_name, mode='r') as A:
    data = A.multi_index[slice_index]
```

Thanks!

ihnorton:
Hi @outdoorpet, thanks for the example. I've fixed the issue in the [0.8.4](https://github.com/TileDB-Inc/TileDB-Py/releases/tag/0.8.4) release, and added an expanded version of the example as a [test](https://github.com/TileDB-Inc/TileDB-Py/blob/101f1819822900053b36ac8869966c8f180923b3/tiledb/tests/test_multi_index.py#L653).

PR #479, opened 2021-02-23
Improve Usability of FragmentInfo API

nguyenv:
* fragment_info() is now exposed to the user and does not have to be
  explicitly exported
* The first parameter of fragment_info() is now the URI. The second
  parameter is the context which is now optional and defaults to
  tiledb.default_ctx()
* The fragment module has been renamed to _fragment so that there is
  only one entry point called fragment* at the top-level
* Add usage example

### DIFF
--- a/examples/fragment_info.py
+++ b/examples/fragment_info.py
@@ -1,0 +1,96 @@
+# fragment_info.py
+#
+# LICENSE
+#
+# The MIT License
+#
+# Copyright (c) 2020 TileDB, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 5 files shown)


examples/fragment_info.py
```diff
@@ -0,0 +1,96 @@
...
+fi = tiledb.fragment_info(array_name)
+
+# Note that load() needs to be called each time the array is written to in
+# order to get updated fragment information.
+fi.load()
+
+schema = tiledb.ArraySchema.load(array_name)
+
+for fragment_num in range(fi.fragment_num()):
+    print("Fragment number: {}".format(fragment_num))
```
ihnorton, line 82: Maybe we should add something like this as a `__repr__` for fragment objects?

ihnorton: hmm, I guess that might need an iterator and wrapper class for each individual fragment.

ihnorton (approved):
Minor suggestion for future addition of `__repr__`, otherwise LGTM 👍

examples/fragment_info.py
nguyenv: Yes we should add that soon. Seth talked about wanting a high level wrapper like `fragments()` below where:

`
fi = tiledb.fragment_info(array_uri)
list_of_all_fragment = fi.fragments()
fragment_object = list_of_all_fragment[0]
fragment_num = fragment_object.num
fragment_uri = fragment_object.uri
...
`

PR #481, opened 2021-02-24
Fix docstring for DenseArray and SparseArray

ihnorton:
These subclasses were set to copy the docstring of Array, which is confusing and incorrect.


ihnorton:
cc @Bhadhy

bhadhy:
Thanks @ihnorton

PR #488, opened 2021-02-26
Addition of High Level array_fragments() Wrapper

nguyenv:
* FragmentsInfo is an indexable and sliceable container class where each
  object contains information about a single fragment
* array_fragments() is a high level function that returns a FragmentsInfo object
* The Pybind11 implementation of _fragment is no longer exposed the user
  and array_fragments(uri) should be used instead of fragment_info(uri)

### DIFF
--- a/tiledb/fragment.py
+++ b/tiledb/fragment.py
@@ -81,6 +81,7 @@ class FragmentInfo:
     """
 
     def __init__(self, fragments: FragmentsInfo, num):
+        self.num = num
         self.uri = fragments.uri[num]
         self.version = fragments.version[num]
         self.non_empty_domain = fragments.non_empty_domain[num]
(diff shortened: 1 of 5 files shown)


tiledb/fragment.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, array_uri, ctx=None):
+        schema = tiledb.ArraySchema.load(array_uri)
+
+        self.array_uri = array_uri
+
+        fi = _fragment.info(self.array_uri, ctx)
+        fi.load()
+
+        self.uri = fi.fragment_uri()
+        self.nums = fi.fragment_num()
```
ihnorton, line 25: IMO this should be hidden as `_num_fragments`. using `len(fi)` feels "pythonic"

ihnorton (commented):
If the name is being changed, I would suggest `array_fragments(uri) -> FragmentsInfo` instead of `fragments_info`.

(maybe with a deprecation)

Shelnutt2 (approved):

PR #505, opened 2021-03-16
Remove non_empty_domain_var() Fragment Info PyBind11 Function

nguyenv:
Remove non_empty_domain_var() Fragment Info PyBind11 Function

    * This has now been consolidated into get_non_empty_domain().

### DIFF
--- a/tiledb/fragment.cc
+++ b/tiledb/fragment.cc
@@ -111,7 +111,7 @@ public:
   py::tuple get_non_empty_domain(py::object schema, uint32_t fid, T did) const {
     py::bool_ isvar = get_dim_isvar(schema.attr("domain"), did);
 
-    if(isvar) {
+    if (isvar) {
       pair<string, string> lims = fi_->non_empty_domain_var(fid, did);
       return py::make_tuple(lims.first, lims.second);
     }
--- a/tiledb/fragment.cc
+++ b/tiledb/fragment.cc
@@ -109,6 +109,13 @@ public:
 
   template <typename T>
   py::tuple get_non_empty_domain(py::object schema, uint32_t fid, T did) const {
+    py::bool_ isvar = get_dim_isvar(schema.attr("domain"), did);
+
+    if(isvar) {
+      pair<string, string> lims = fi_->non_empty_domain_var(fid, did);
+      return py::make_tuple(lims.first, lims.second);
+    }
+
     py::dtype type = get_dim_type(schema.attr("domain"), did);
     py::dtype array_type =
         type.kind() == 'M' ? pybind11::dtype::of<uint64_t>() : type;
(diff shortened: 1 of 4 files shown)


tiledb/tests/test_libtiledb.py
```diff
@@ -1927,15 +1907,7 @@ def test_varlen_wrong_domain(self):
...
-        A = np.array(
-            [
-                b"aa",
-                b"bbb",
-                b"cccc",
-                np.uint64([1, 3, 4]),
-            ],
-            dtype=object,
-        )
+        A = np.array([b"aa", b"bbb", b"cccc", np.uint64([1, 3, 4])], dtype=object)
```
ihnorton, line 1910: (This is what I was talking about w/ black :/ -- this file has been formatted before, no reason for it to change now)

tiledb/fragment.cc
```diff
@@ -109,6 +109,13 @@ class PyFragmentInfo {
 
   template <typename T>
   py::tuple get_non_empty_domain(py::object schema, uint32_t fid, T did) const {
+    py::bool_ isvar = get_dim_isvar(schema.attr("domain"), did);
```
ihnorton, line 112: Does this need to be templated? Seems like it is only called for `uint32_t`

ihnorton (approved):
Minor question, but LGTM

tiledb/tests/test_libtiledb.py
nguyenv: Sigh yeah and there seems to be some incongruence between the version of Black being used for GitHub Actions and what I have installed locally. I set it up to always use the latest stable version, so I don't know which one isn't up to date. I also didn't realize how frequently they make changes to this format.

tiledb/fragment.cc
nguyenv: Yes it should be able to take in both the dimension's index the dimension's name.

tiledb/tests/test_libtiledb.py
Shelnutt2: Does black have a way to introduce a format file so its consistent in the repo across versions that change the default?

ihnorton: Last time I looked it seems they always do it The Correct Way so no configuration ... except that way changes :)

tiledb/fragment.cc
ihnorton: I see, thanks 👍

tiledb/tests/test_libtiledb.py
eddelbuettel: Maybe you can tie the call to Black down in a Docker container to freeze it.  Seems like a heavy hammerl though.

PR #509, opened 2021-03-18
tiledb.dataframe_ improvements

gsakkis:
- Update `dtype_from_column` to handle all the different dtypes more consistently.
- Make `ColumnInfo` a (frozen) dataclass, incorporate `dtype_from_column` as its constructor and add exhaustive tests.
- Update `dataframe_to_np_arrays` to handle extension dtypes better.
- Misc. refactorings:
  - Simplify `create_dims` & `attrs_from_df`
  - Drop `dim_info_for_column`


tiledb/dataframe_.py
```diff
@@ -82,129 +72,103 @@ def parse_tiledb_kwargs(kwargs):
...
-            pd.Int32Dtype,
-            pd.Int16Dtype,
-            pd.Int8Dtype,
-            pd.UInt64Dtype,
-            pd.UInt32Dtype,
-            pd.UInt16Dtype,
-            pd.UInt8Dtype,
-        ),
-    )
+@dataclass(init=False, frozen=True)
```
ihnorton, line 75: Technically we currently support 3.6+, and only 3.6 and 3.7 work on TileDB-Cloud... Is the [dataclasses](https://pypi.org/project/dataclasses/) backport package sufficient here? If so that's fine to add as a dependency IMO.

ihnorton (approved):
Comment/question about dataclass, otherwise this looks great!

tiledb/dataframe_.py
ihnorton: Hmm, looks like the 3.6 test was inadvertently dropped, which is why this didn't error out. We do still build wheels for it, so I'm going to add a test target back to the matrix.

gsakkis: Yes I have used the dataclasses backport with 3.6 before, it's the same as the official (there used to be a bug but I see the fix has been backported too since last time I checked). Having said that I'm ok with dropping the dataclass if we don't want to add an optional dependency, it's merely a convenience.

ihnorton: It can be conditional on 3.6 so I'm fine with adding the dep there.

ihnorton: (There are other places I've wanted to use it)

ihnorton:
Please feel free to merge whenever you are ready.

PR #510, opened 2021-03-18
Test on python 3.6 with ubuntu 16.04

ihnorton:
- We have good test coverage for higher versions w/ github actions, so we should test 3.6 as well until EOL.
- Add conditional requirement of [dataclasses](https://pypi.org/project/dataclasses/) backport for python 3.6, which will be used in some dataframe code (and probably other places)


ihnorton:
(merging with #511 to get a full test run w/ 3.6 and 2.2.6)

PR #513, opened 2021-03-20
.df[] Returns Clear Error Message Regarding Lack of Arrow List Support

nguyenv:


tiledb/multirange_indexing.py
```diff
@@ -287,6 +287,31 @@ def __getitem__(self, idx):
         if self.use_arrow:
             import pyarrow as pa
 
+            # TODO currently there is lack of support for Arrow list types. this
```
ihnorton, line 290: Please move this check to a new conditional block above the `results = super...` section so that we error out before  submitting the query.

ihnorton (approved):
Small change, otherwise LGTM. Thanks!

PR #515, opened 2021-03-22
Variable length attribute support for Pandas extension types

gsakkis:
This PR adds support for storing dataframe columns as TileDB [var-length attributes](https://docs.tiledb.com/main/solutions/tiledb-embedded/api-usage/writing-arrays/var-length-attributes). It introduces a new optional parameter for `tiledb.from_pandas`, `varlen_types`, a set of Pandas extension dtypes to be stored as var-length attributes. An extension dtype is considered to be var-length if the result of calling `to_numpy()` on a pandas Series of this dtype returns a "ragged array", a numpy array of arrays.

All the var-length changes are contained in the [last commit](https://github.com/TileDB-Inc/TileDB-Py/commit/ee9cb97767b0feadbdf4a3487edef54065c35020), the previous commits are refactorings.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@
 
 ## Improvements
 * Addition of `VFS()` functions `copy_file()` and `copy_dir()` [#507](https://github.com/TileDB-Inc/TileDB-Py/pull/507)
+* Add support in `from_pandas` for storing Pandas extension types as variable-length attributes [#515](https://github.com/TileDB-Inc/TileDB-Py/pull/515)
 
 ## Bug fixes
 * Multi-length attributes, regardless of fixed or var-length, do not work query properly with PyArrow enabled due to lack of Arrow List support. When using `.df[]` with PyArrow enabled, we are returning a clear message to the user to use `query(use_pyarrow=False)` [#513](https://github.com/TileDB-Inc/TileDB-Py/pull/513)
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -53,6 +53,7 @@ TILEDB_KWARG_DEFAULTS = {
     "row_start_idx": None,
     "fillna": None,
     "column_types": None,
+    "varlen_types": None,
     "capacity": None,
     "date_spec": None,
     "cell_order": "row-major",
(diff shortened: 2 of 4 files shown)


gsakkis:
@ihnorton the unit test currently fails for PyArrow with `pyarrow.lib.ArrowInvalid: Expected 2 buffers for imported type uint16, ArrowArray struct has 3`, not sure why. It's passing without PyArrow.

nguyenv:
> @ihnorton the unit test currently fails for PyArrow with `pyarrow.lib.ArrowInvalid: Expected 2 buffers for imported type uint16, ArrowArray struct has 3`, not sure why. It's passing without PyArrow.

We currently don't have full support for Arrow lists, so any multi-length attribute, regardless of var or fixed length, is not working.
https://github.com/TileDB-Inc/TileDB-Py/pull/513

ihnorton (commented):
It would be nice to avoid a new kwarg b/c there are so many already; but that would probably be more complicated.

Otherwise this looks good, happy to take another quick look whenever it's ready.

ihnorton:
Also please add an entry in HISTORY.

ihnorton:
@gsakkis oh - how about allowing `ColumnInfo` as an argument to `column_types`, instead of adding `varlen_types`?

gsakkis:
> @gsakkis oh - how about allowing `ColumnInfo` as an argument to `column_types`, instead of adding `varlen_types`?

This could work, in fact that was my first idea. The downside is it exposes `ColumnInfo` which is currently "private" and is harder to set up correctly manually. For example `ColumnInfo.dtype` should be (convertible to) the numpy dtype used as the low level TileDB Attr storage, not the high level Pandas dtype. Also `nullable` and `var` can't be both true at the moment.

My second idea to infer var-ness was to check for the presence of a, say, `var_length` bool attribute/property in the `dtype` instance. This avoids exposing `ColumnInfo` but requires customizing or extending existing extension types with a new attribute. Adding a new kwarg to `from_pandas` seemed the least invasive and error-prone solution.

ihnorton:
Ok, that makes sense. Thanks for the discussion/explanation.

gsakkis:
> Also please add an entry in HISTORY.

This will be for v0.8.6, right?

ihnorton:
> This will be for v0.8.6, right?

Yes

ihnorton (approved):

PR #516, opened 2021-03-22
tiledb.from_pandas fixes for fillna and date_spec

gsakkis:
Fixes the following bugs:
- `fillna`: https://github.com/TileDB-Inc/TileDB-Py/commit/4256a976c8d52f3911f81d42ac74b1bd8b722064
   - Don't mutate the input dataframe
- `date_spec`: https://github.com/TileDB-Inc/TileDB-Py/pull/516/commits/6394edd5db8a6ab749cd063709cee362e7fa6ee6
   - Don't mutate the input dataframe
   - Always store the `date_spec` attributes as `datetime`

The previous commits are unrelated refactorings.

PR #517, opened 2021-03-22
Fix Pandas < 1.2 support (and Python 3.6 CI)

ihnorton:
`pd.core.dtypes.common.get_dtype` was added/exposed in Pandas 1.2.

(1.2 supports 3.7+ only)

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -82,7 +82,14 @@ class ColumnInfo:
     def __init__(self, arr_or_dtype):
         import pandas as pd
 
-        dtype = pd.core.dtypes.common.get_dtype(arr_or_dtype)
+        # compatibility for pandas < 1.2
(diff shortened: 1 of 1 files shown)


tiledb/dataframe_.py
```diff
@@ -82,7 +82,14 @@ class ColumnInfo:
     def __init__(self, arr_or_dtype):
         import pandas as pd
 
-        dtype = pd.core.dtypes.common.get_dtype(arr_or_dtype)
+        # compatibility for pandas < 1.2
```
gsakkis, line 85: @ihnorton I removed `get_dtype` in https://github.com/TileDB-Inc/TileDB-Py/pull/516/commits/eb8af19c266e0b761d7965dd9a3cb489501a05f0

ihnorton: Ok, thanks

PR #520, opened 2021-03-26
Extend tiledb.open_dataframe

gsakkis:
- Add `attrs` and `use_arrow` optional parameters
- Reuse the `DataframeIndexer` code path

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -698,7 +698,7 @@ def _tiledb_result_as_dataframe(readable_array, result_dict):
     return df
 
 
-def open_dataframe(uri, ctx=None):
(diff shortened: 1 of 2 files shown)


tiledb/dataframe_.py
```diff
@@ -698,7 +698,7 @@ def _tiledb_result_as_dataframe(readable_array, result_dict):
     return df
 
 
-def open_dataframe(uri, ctx=None):
+def open_dataframe(uri, *, attrs=None, use_arrow=None, ctx=None):
```
ihnorton, line 701: Wondering what is the goal of `*` without capture here?

ihnorton (approved):
Minor question, otherwise LGTM. Thanks!

PR #523, opened 2021-03-26
Misc CI fixes: Python 3.8 warning and add dask[distributed] to requirements_dev

ihnorton:


ihnorton:
cherry-picked to #525

PR #527, opened 2021-04-01
Add test querying on nextafter real dimension values

ihnorton:


ihnorton:
(re-opening to try to fix problem with linter action - unable to load)

PR #529, opened 2021-04-03
Fix linter GHA action

ihnorton:


.github/workflows/format.yml
```diff
@@ -9,8 +9,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
 
-      - name: Check Python Black Format
-        uses: psf/black@stable
+      - name: Check Python Black Formata
```
eddelbuettel, line 12: Micro-typo with the trailing a 'Formata'

PR #532, opened 2021-04-06
Support returning empty result from MultiRangeIndexer and DataFrameIndexer

gsakkis:
- Introduce `tiledb.EmptyRange` sentinel value to denote an empty slice range.
- `A.multi_index[tiledb.EmptyRange]` returns a dict with values 0-length `np.array`s with the expected `dtype`.
- `A.df[tiledb.EmptyRange]` returns an empty `pd.DataFrame` with the expected column/`dtypes`.

Also extend `tiledb.open_dataframe` with an optional `idx` parameter (default=full range `slice(None, None)`).


ihnorton (approved):
LGTM. cc @nguyenv

nguyenv (approved):

ihnorton (approved):

PR #537, opened 2021-04-11
Run tests with pytest

ihnorton:


tiledb/tests/test_libtiledb.py
```diff
@@ -3769,17 +3775,12 @@ def test_tiledb_py_0_6_anon_attr(self):
             self.assertEqual(qres["d"], 0)
 
 
+pytest.mark.skipif(sys.platform == "win32", reason="Only run MemoryTest on linux")
```
gsakkis, line 3778: This should be a `MemoryTest` @decorator, right?

nguyenv (approved):

tiledb/tests/test_dask.py
```diff
@@ -1,28 +1,17 @@
...
 
+import sys
 import tiledb
 from tiledb.tests.common import DiskTestCase
 
 import numpy as np
 from numpy.testing import assert_array_equal, assert_approx_equal
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
```
gsakkis, line 13: If we're standardizing on `sys.platform`, there are a few leftovers of `platform.system()` and `os.name` you may want to port.

gsakkis (approved):

tiledb/tests/test_libtiledb.py
ihnorton: yep, missed the extra \n. thanks

PR #543, opened 2021-04-16
Refactor and extend metadata

gsakkis:
- Cleaned up `libmetadata.pyx` in preparation of adding numpy support.
- Dropped unused `tiledb.Metadata` pure Python class and associated unused functions
- Fixed broken `Metadata.values()`
- Rewrote `Metadata.__contains__` using `tiledb_array_has_metadata_key` 
- Added missing mapping methods to `Metadata`:
  -  `__iter__`, `get`, `update` 
  - `setdefault`, `popitem`, `clear`: raise `NotImplementedError` just like `pop`
- Raise `TypeError` instead of `ValueError` for unsupported metadata value types
- Augmented tests to do property testing using `hypothesis`


tiledb/tests/test_metadata.py
```diff
@@ -1,25 +1,96 @@
...
+    }
+)
+
 
-        with tiledb.from_numpy(path, np.ones((5,), np.float64)) as A:
+class MetadataTest(DiskTestCase):
+    def assert_equal_md_values(self, written_value, read_value):
+        if not isinstance(written_value, (list, tuple)):
+            self.assertEqual(read_value, written_value)
+        # we don't round-trip perfectly sequences
```
ihnorton, line 33: If we are adding other metadata for numpy anyway, we could probably fix this.

ihnorton (approved):

PR #544, opened 2021-04-16
Try to fix hanging CI: don't let pytest recurse into .git [ch6686]

ihnorton:
GHA CI is consistently hanging, and after some [debug builds](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/756722444) it appears the problem may be due to recursing into .git.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,3 +2,6 @@
 python_classes = *Test*
 python_files = "test_*.py"
 testpaths = ./tiledb/tests
+filterwarnings =
+    ignore
+norecursedirs = .git build


nguyenv:
Wow good to know! I've been having major issues with the TileDB-CLI GHA hanging all the time as well.

gsakkis:
Interesting, the default [norecursedirs](https://docs.pytest.org/en/stable/reference.html#confval-norecursedirs) should ignore `.git`: 
> Default patterns are '*.egg', '.*', '_darcs', 'build', 'CVS', 'dist', 'node_modules', 'venv', '{arch}'.

ihnorton:
Hmm. Well maybe this was just luck? I'm pretty sure the hang is in test discovery based on the logs, but will look again if it keeps happening.

PR #546, opened 2021-04-19
Numpy metadata

gsakkis:
## Summary 
Adds support for writing and reading numpy arrays as metadata.

### Implementation

A numpy array is stored as one or two metadata values:
- The 1-D flat data values as returned by [numpy.ravel](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html). The `dtype` is preserved for all numpy dtypes that have a corresponding TileDB type (e.g. `np.uint16 <-> TILEDB_UINT16`)
- The shape tuple, only if the original array is not 1-D. 

The keys for these values are prefixed by a fixed prefix to differentiate them from regular non-numpy keys:
- `_NP_DATA_PREFIX = "__np_flat_"`
- `_NP_SHAPE_PREFIX = "__np_shape_"`

### Performance implications per method

- Get (`__getitem__`): First it looks up `key`; if it succeeds it returns it (non-numpy value), otherwise (numpy value) it does two extra lookups for `_NP_DATA_PREFIX + key` and `_NP_SHAPE_PREFIX + key`.
  - Non-numpy: 1 `tiledb_array_get_metadata` call - no extra overhead.
  - Numpy: 3 `tiledb_array_get_metadata` calls.

- Contains (`__contains__`):
  - Non-numpy: 1 `tiledb_array_has_metadata_key` call - no extra overhead.
  - Numpy: 2 `tiledb_array_has_metadata_key` calls (`key`, `_NP_DATA_PREFIX + key`)
 
- Delete (`__delitem__`): 3 `tiledb_array_delete_metadata` calls (`key`, `_NP_DATA_PREFIX + key`, `_NP_SHAPE_PREFIX + key`). The reason is that `tiledb_array_delete_metadata` doesn't return whether it found and deleted the key or not, so we have to try deleting all 3 potential keys.

- Set (`__setitem__`): First delete `key` and then
  - Non-numpy: 1 `tiledb_array_put_metadata` call.
  - Numpy: 1 (for 1-D arrays) or 2 (for non 1-D arrays) `tiledb_array_put_metadata` calls.
  
  The reason we have to delete the key before (re)setting it is to handle correctly these cases:
   - replacing a non-numpy value with a numpy value, or vice versa
   - replacing a non 1-D numpy value with a 1-D numpy value

- Iteration (`__iter__`, `keys`, `values`, `items`): No extra overhead in terms of `tiledb_array_*` calls. There might be a (presumably minor) overhead for:
  - filtering out or translating internal keys to external
  - combining the stored flat arrays with the respective shape tuples (if any) to restore the original arrays.

- Length (`__len__`): This is the weak spot of the current approach. In addition to the `tiledb_array_get_metadata_num` call, it has to subtract the number of shape entries (=keys prefixed with `_NP_SHAPE_PREFIX`), which it finds by iterating over all keys :disappointed: 

  An alternative approach could be to introduce a new internal key, say `__np_shape_count`, which would be a counter of shape keys. The downsides would be:
  - extra overhead in `__setitem__` and `__delitem__` for incrementing and decrementing the counter, respectively.
  - ~extra code complexity in implementing "client side" [increment](https://redis.io/commands/INCR)/[decrement](https://redis.io/commands/DECR) logic, which most  likely wouldn't guarantee atomicity. The right place for such functionality (if it exists at all) is in TileDB core I think~.
    EDIT: just occurred to me that increment/decrement logic can't be implemented client-side as it would require both read and write access to the array metadata.

### DIFF
--- a/tiledb/libmetadata.pyx
+++ b/tiledb/libmetadata.pyx
@@ -7,7 +7,15 @@ import weakref
 from collections.abc import MutableMapping
 
 from cython.operator cimport dereference as deref
-from libcpp.limits cimport numeric_limits
+
+
+_NP_DATA_PREFIX = "__np_flat_"
+_NP_SHAPE_PREFIX = "__np_shape_"
+
+
+cdef extern from "Python.h":
+    int PyBUF_READ
+    object PyMemoryView_FromMemory(char*, Py_ssize_t, int)
 
 
 cdef class PackedBuffer:
@@ -52,7 +60,6 @@ cdef PackedBuffer pack_metadata_val(value):
     if tiledb_type == TILEDB_INT64:
         int64_ptr = <int64_t*>&data_view[0]
         while pack_idx < value_num:
-            # TODO ideally we would support numpy scalars here
             value_item = value[pack_idx]
             if not isinstance(value_item, int):
                 raise TypeError(f"Mixed-type sequences are not supported: {value}")
@@ -61,7 +68,6 @@ cdef PackedBuffer pack_metadata_val(value):
     else:
         double_ptr = <double*>&data_view[0]
         while pack_idx < value_num:
-            # TODO ideally we would support numpy scalars here
             value_item = value[pack_idx]
             if not isinstance(value_item, float):
                 raise TypeError(f"Mixed-type sequences are not supported: {value}")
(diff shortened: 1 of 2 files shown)


ihnorton (commented):
I'm not too concerned about the performance of multiple lookups because metadata is loaded all-at-once (upon request), so it's either in-memory or it's not, and in-memory requests should be fast. However, in light of that, metadata should be "relatively small", so hopefully this is not intended to be used for 100s of MB arrays.

@gsakkis could you please clarify how is this storing `np.object` and fixed-width string or bytes arrays? eg `a = np.array(["abc", "defhgh"], dtype="O")` vs `np.array(["abc", "defhgh"], dtype='U6')`. I think for consistency with other APIs we want to store all strings as var-length with no internal null-termination (because we should always know the length) -- it looks like that is the case here, but want to confirm.

@jp-dark have you tested this for your use-case?

gsakkis:
> @gsakkis could you please clarify how is this storing np.object and fixed-width string or bytes arrays? 

@ihnorton it isn't; this PR supports all the dtypes that can already be stored in TileDB as tuples, so not strings or arbitrary objects. Once the core supports tuples of strings, it will automatically (or perhaps with a few extra changes) work for ndarrays of strings.

jp-dark (approved):
This works for my use case.

ihnorton (approved):

PR #555, opened 2021-05-06
 Remove duplicated +1 in datetime range calculations & misc fixes

ihnorton:


nguyenv (approved):

tiledb/libtiledb.pyx
```diff
@@ -2620,10 +2620,10 @@ cdef class Dim(object):
                     # this means to use the full extent of the type
                     if dtype.kind == 'M':
                         date_unit = np.datetime_data(dtype)[0]
-                        dtype_min = np.datetime64(dtype_min + 1, date_unit)
+                        dtype_min = np.datetime64(dtype_min, date_unit)
```
gsakkis, line 2623: Isn't this expression what `dtype_range` does to compute `dtype_min`? I think it's not needed anymore.

tiledb/libtiledb.pyx
```diff
@@ -2620,10 +2620,10 @@ cdef class Dim(object):
...
                     elif dtype == np.int64 and not (dtype.kind == 'M'):
                         # except that the domain range currently must fit in UInt64
-                        dtype_min = dtype_min + 1
+                        dtype_min = dtype_min
```
gsakkis, line 2626: This is redundant now.

So with the comment above, I think the whole block can be simplified to:
```
if domain == (None, None):
    domain = (dtype_min, dtype_max)
````

tiledb/libtiledb.pyx
ihnorton: ✔️

tiledb/libtiledb.pyx
ihnorton: Thanks

gsakkis (approved):

PR #563, opened 2021-05-11
Ctx context manager and ContextVar

gsakkis:
### Problem description

Most `tiledb` functions and classes accept an optional `ctx` parameter. This is less than ideal for the following reasons:

- **Verbose**
Taking [examples/string_float_int_dimensions.py](https://github.com/TileDB-Inc/TileDB-Py/blob/dev/examples/string_float_int_dimensions.py#L43) as an example, `ctx` is initialized and passed to 6 different calls:

```
ctx = tiledb.Ctx()

dom = tiledb.Domain(
    *[
        tiledb.Dim(name="str_dim", domain=(None, None), dtype=np.bytes_, ctx=ctx),
        tiledb.Dim(name="int64_dim", domain=(0, 100), tile=10, dtype=np.int64, ctx=ctx),
        tiledb.Dim(
            name="float64_dim",
            domain=(-100.0, 100.0),
            tile=10,
            dtype=np.float64,
            ctx=ctx,
        ),
    ],
    ctx=ctx
)

att = tiledb.Attr(name="a", ctx=ctx, dtype=np.int64)
schema = tiledb.ArraySchema(
    ctx=ctx, domain=dom, attrs=(att,), sparse=True, capacity=10000
)
tiledb.SparseArray.create(path, schema)
```
- **Error-prone**
In the example above `ctx` is not passed to the `tiledb.SparseArray.create` call, most likely an accidental omission. In this particular example it doesn't matter because the default global `Ctx` instance is equal to `tiledb.Ctx()` but it could matter if `ctx` was initialized with a different `Config`. Such accidental omissions happen in several places in [tiledb.dataframe_](https://github.com/TileDB-Inc/TileDB-Py/blob/dev/tiledb/dataframe_.py).

- **Global variable**
A global `Ctx` instance can be initialized and used by default via `tiledb.default_ctx()`. This solves the above verbosity and error-proneness issues at the cost of introducing a global variable. In addition to the [usual drawbacks](https://wiki.c2.com/?GlobalVariablesAreBad) of global variables, the global `Ctx` can be initialized only once in a program's lifetime.

---

### Proposal
This PR introduces the following enhancements:
- Adds a new `tiledb.ctx(config=None)` context manager for specifying a `tiledb.Ctx` within a block of code. This `Ctx` is both returned by the context manager (for explicit usage) and set as "global" within the scope of the block (for implicit usage).
- Replaces the global variable default with a [ContextVar](https://www.python.org/dev/peps/pep-0567/) (introduced in Python 3.7 and available as 3rd party backport for 3.6). In short, a `ContextVar` can be thought of as a generalization of global or thread-local variables that also plays well with asynchronous (async/await) code.
- Removes the (long deprecated) `tiledb.initialize_ctx` function.

Once/if this is accepted, a following PR will clean up the redundant and/or verbose explicit `ctx` usage with appropriate `tiledb.ctx()` context blocks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 numpy>=1.16.*
 setuptools
+contextvars ;python_version<"3.7"
 dataclasses ;python_version<"3.7"
--- a/setup.py
+++ b/setup.py
@@ -685,6 +685,7 @@ setup(
     install_requires=[
         "numpy>=1.16",
         "wheel>=0.30",
+        "contextvars ;python_version<'3.7'",
         "dataclasses ;python_version<'3.7'",
     ],
     packages=find_packages(),
--- a/tiledb/__init__.py
+++ b/tiledb/__init__.py
@@ -33,6 +33,7 @@ except:
         # Otherwise try loading by name only.
         ctypes.CDLL(lib_name)
 
+from .ctx import default_ctx, scope_ctx
 from .libtiledb import (
     Array,
     Ctx,
@@ -59,7 +60,6 @@ from .libtiledb import (
     ChecksumMD5Filter,
     ChecksumSHA256Filter,
     consolidate,
-    default_ctx,
     group_create,
     object_type,
     ls,
--- a/tiledb/ctx.py
+++ b/tiledb/ctx.py
@@ -1,0 +1,47 @@
+from contextlib import contextmanager
+from contextvars import ContextVar
+
+import tiledb
+
+_ctx_var = ContextVar("ctx")
+
+
+@contextmanager
+def scope_ctx(config=None):
+    """
+    Context manager for setting the default `tiledb.Ctx` context variable when entering
+    a block of code and restoring it to its previous value when exiting the block.
+
+    :param config: :py:class:`tiledb.Config` object or dictionary with config parameters.
+    :return: Ctx
+    """
+    token = _ctx_var.set(tiledb.Ctx(config))
+    try:
+        yield _ctx_var.get()
+    finally:
+        _ctx_var.reset(token)
+
+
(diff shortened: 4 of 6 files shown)


ihnorton:
I like this very much for usage in the test blocks. I have an old prototype of this on my "post Python 3" list, so thanks!

However, I'm a bit hesitant about encouraging it as a general pattern, because contexts should be created rarely. Contexts are relatively expensive: each one has a separate thread pool and cache, and recreating a context in an interactive session could be detrimental to performance (because the caches will be invalidated). The pattern I'm worried about is:

```
with tiledb.ctx(...):
  with tiledb.open(...):
     ..

with tiledb.ctx(...):
  with tiledb.open(...):
    ...
```

I also think the `Ctx`/`ctx` overlap could be confusing; what do you think about `ctx -> scope_ctx`? (soft suggestion - please discuss if you don't like).

gsakkis:
> The pattern I'm worried about is:
> 
> ```
> with tiledb.ctx(...):
>   with tiledb.open(...):
>     ..
> 
> with tiledb.ctx(...):
>   with tiledb.open(...):
>     ...
> ```

Presumably one would(should) use this only if the second context uses a different config from the first, in which case the extra cost of a new `ctx` is unavoidable. Besides one can already create explicitly contexts with `ctx = tiledb.Ctx()`, not sure why the context manager would be more prone to suboptimal usage.

> I also think the `Ctx`/`ctx` overlap could be confusing; what do you think about `ctx -> scope_ctx`? (soft suggestion - please discuss if you don't like).

No strong opinion on naming, `scope_ctx` is fine; thanks!

ihnorton (approved):
💯 thanks!

PR #564, opened 2021-05-11
Fix and test from_pandas(attr_filters=None, dim_filters=None)

ihnorton:


tiledb/dataframe_.py
```diff
@@ -363,14 +377,15 @@ def create_dims(
...
-            else:
-                dim_filters = None
-        elif filters is not None:
+            dim_filters = filters.get(name, None)
+        elif isinstance(filters, list):
+            dim_filters = tiledb.FilterList(filters)
+        elif isinstance(filters, tiledb.FilterList):
             dim_filters = filters
-        else:
+        elif filters or filters is None:
```
gsakkis, line 385: Shouldn't this condition be just `elif filters is None:`?

gsakkis, line 385: Shouldn't this condition be just `elif filters is None`?

ihnorton: #568

PR #568, opened 2021-05-13
Follow up fix for ch7372

ihnorton:


tiledb/dataframe_.py
```diff
@@ -569,6 +573,20 @@ def from_pandas(uri, dataframe, **kwargs):
...
+    elif coords_filters is None:
+        # empty coords filters support
+        coords_filters = tiledb.FilterList()
+
+    if offsets_filters is True:
+        offsets_filters = None
+    elif offsets_filters is None:
+        # empty offsets filters support
+        offsets_filters = tiledb.FilterList()
+
```
gsakkis, line 588: 1. Can `coords_filters` and `offsets_filters` be a list? If yes shouldn't they be wrapped in a ` tiledb.FilterList`?
2. Unlike `attr_filters` and `dim_filters` no ValueError is raised here for invalid filters.

tiledb/dataframe_.py
```diff
@@ -383,7 +384,9 @@ def create_dims(
             dim_filters = tiledb.FilterList(filters)
         elif isinstance(filters, tiledb.FilterList):
             dim_filters = filters
-        elif filters or filters is None:
+        elif filters is None:
+            dim_filters = tiledb.FilterList()
+        elif filters:
             dim_filters = None
         else:
```
gsakkis, line 391: Is this deliberately inconsistent with the `attr_filters` handling? 
```
if attr_filters is True:
    # default case
    filters = tiledb.FilterList([tiledb.ZstdFilter(-1, ctx=ctx)], ctx=ctx) 
...
elif attr_filters is None:
    filters = None
```

gsakkis (changes requested):

tiledb/dataframe_.py
ihnorton: Updated, thanks.

tiledb/dataframe_.py
ihnorton: Updated

tiledb/dataframe_.py
```diff
@@ -176,25 +177,34 @@ def _get_column_infos(df, column_types, varlen_types):
...
+    elif isinstance(filters, tiledb.libtiledb.Filter):
+        return tiledb.FilterList([filters])
+    else:
+        raise ValueError("Unknown FilterList type!")
+
+
+def _get_attr_dim_filters(name, filters):
+    if isinstance(filters, dict):
+        # support passing a dict of filters per-attribute
+        return filters.get(name, None)
```
gsakkis, line 198: I'd suggest `_get_schema_filters(filters.get(name, True))` to support all cases `_get_schema_filters` handles

tiledb/dataframe_.py
```diff
@@ -524,7 +525,8 @@ def from_pandas(uri, dataframe, **kwargs):
     mode = tiledb_args.get("mode", "ingest")
     attr_filters = tiledb_args.get("attr_filters", True)
     dim_filters = tiledb_args.get("dim_filters", True)
-    coords_filters = tiledb_args.get("coords_filters", None)
+    coords_filters = tiledb_args.get("coords_filters", True)
+    offsets_filters = tiledb_args.get("offsets_filters", True)
```
gsakkis, line 529: Not part of this PR but as a future TODO refactoring: remove the duplication of defaults between these `.get()` calls and `TILEDB_KWARG_DEFAULTS`

gsakkis (approved):

tiledb/dataframe_.py
ihnorton: ✔️

PR #576, opened 2021-05-26
Addition of `QueryCondition` API to Filter Query On Attributes

nguyenv:


setup.py
```diff
@@ -18,7 +18,7 @@
 print("setup.py sys.argv is: ", sys.argv)
 
 # Target branch
-TILEDB_VERSION = "2.2.9"
+TILEDB_VERSION = "dev"
```
nguyenv, line 21: 2.3.0?

tiledb/tests/test_query_condition.py
```diff
@@ -0,0 +1,69 @@
...
+        coords = np.linspace(1, 10, num=10, dtype=np.uint32)
+        data = {"U": U, "D": D, "S": S}
+
+        with tiledb.open(path, "w") as A:
+            A[coords] = data
+
+        with tiledb.open(path) as A:
+            # bytestrings with PyArrow not yet support in TileDB-Py
+            qc = tiledb.QueryCondition("U < 5")
+            result = A.query(attr_cond=qc, use_arrow=False).df[:]
```
ihnorton, line 58: Shouldn't this still work with arrow for the non-string attrs?

ihnorton (changes requested):
This is great @nguyenv, really nice work!

Some minor tweaks from the first pass, and I think we should expand the testing a bit (parametrize test over all supported types; hypothesis test to flush out edge cases), but that can done done in follow-up.

tiledb/tests/test_query_condition.py
nguyenv: If the array contains strings at all then I still need to pass `use_arrow=False` or it will complain

ihnorton: How about if you restrict the attrs, eg `attrs=["U"]`?

tiledb/query_condition.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Values are any Python-valid number or a string enclosed in quotes.
+
+        val ::= <num> | "<str>" | '<str>'
+
+    Example
+    -------
+    with tiledb.open(uri, mode="r") as A:
+        # select cells where the attribute values for foo are less than 5,
+        # baz greater to or equal to 1.324, and bar equal to string asdf.
```
nguyenv, line 52: Correct this documentation

tiledb/query_condition.py
```diff
@@ -0,0 +1,149 @@
...
+class QueryCondition(ast.NodeVisitor):
+    """
+    Class representing a TileDB query condition object for attribute filtering
+    pushdown. Set the query condition with a string representing an expression
+    as defined by the grammar below. A more straight forward example of usage is
+    given beneath.
+
+    BNF
+    ---
+    An expression is made up of one or more Boolean expressions. Multiple Boolean
```
nguyenv, line 21: Use term "statement" instead of "expression" to clearly differentiate from "Boolean expression"

tiledb/tests/test_query_condition.py
nguyenv: Yes that works

ihnorton (approved):
LGTM 💯 I have some notes on some minor issues, but I think we can merge this as-is and then address those in follow-up.

PR #578, opened 2021-05-26
Pass The Context to ArraySchema in FragmentInfolist

nguyenv:


tiledb/fragment.py
```diff
@@ -15,7 +15,7 @@ class FragmentInfoList:
     """
 
     def __init__(self, array_uri, ctx=None):
-        schema = tiledb.ArraySchema.load(array_uri)
+        schema = tiledb.ArraySchema.load(array_uri, ctx=ctx)
```
ihnorton, line 18: Pls change to also use the new _get_ctx arrangement to support scope_ctx

ihnorton (changes requested):

ihnorton (approved):
Thanks!

PR #581, opened 2021-05-31
Fix and test ch7481, resize validity buffer correctly for incomplete …

ihnorton:
…retries


Shelnutt2 (approved):
Verified this works great!

PR #589, opened 2021-06-07
Various QueryCondition API Updates

nguyenv:


tiledb/query_condition.py
```diff
@@ -100,17 +117,48 @@ def visit_Compare(self, node):
...
+                val = cast(val)
+                dtype_name = dtype.name
+            except ValueError:
+                raise tiledb.TileDBError(
+                    f"Type mismatch between attribute `{att}` and value `{val}`."
+                )
 
-        return qc.qc(att, val, op, self._ctx)
+        c_obj = qc.qc(self._ctx)
+        init_qc = getattr(c_obj, f"init_{dtype_name}")
```
ihnorton, line 158: What happens when:
- the dtype isn't supported -- we get an `AttributeError`, right? IMO that's going to be a little confusing so probably need a try/except here to explain the error
- the supplied value is out of bounds for the dtype: we get a `ValueError` or a `TypeError` for invalid cast? Probably also could use some error clarification.

nguyenv: Good points. I will add error checks for these then.

nguyenv: For your second question, an out of bounds value will throw a `tiledb.libtiledb.TileDBError: OverflowError: Python int too large to convert to C long` from the C++ end.

tiledb/query_condition.py
```diff
@@ -57,11 +58,27 @@ class QueryCondition(ast.NodeVisitor):
...
+        self._query_attrs = query_attrs
+
+        if self.tree.body:
+            self._c_obj = self.visit(self.tree.body[0])
+
+            if not isinstance(self._c_obj, tiledb._query_condition.qc):
+                raise tiledb.TileDBError(
+                    "Malformed query condition statement. A query condition must "
+                    "be made up of one or more Boolean expressions."
+                )
```
ihnorton, line 81: Do we need to raise an error in the (invisible) `else` case here?

tiledb/tests/test_query_condition.py
```diff
@@ -67,3 +100,29 @@ def test_ints_floats_bytestrings(self):
...
+        with tiledb.open(input_array_UIDS) as A:
+            qc = tiledb.QueryCondition("U < 0.1")
+            result = A.query(attr_cond=qc, attrs=["U"]).df[:]
+            assert all(result["U"] < 0.1)
+
+            qc = tiledb.QueryCondition("U < 1.0")
+            result = A.query(attr_cond=qc, attrs=["U"]).df[:]
+            assert all(result["U"] < 1.0)
+
+            qc = tiledb.QueryCondition("U < '1'")
```
ihnorton, line 114: How does this work - is it casting the string to an int somewhere?

(Kind of seems like it should be an error to me, unless I'm misunderstanding what is happening here)

ihnorton (approved):
Two minor (I hope) questions. Otherwise LGTM, thanks!

tiledb/tests/test_query_condition.py
nguyenv: https://github.com/TileDB-Inc/TileDB-Py/pull/589/files/85713a41f182faa67090bcc20490c36a9aaf9b52#diff-bc89368eea7e3bcb8e4e64c38e4b2caf1f1ae9ddbf27846dd9363e56d1114367R149-R150

NumPy is doing the casting here to turn this string into an int. In the case that the string can't be casting to an int, it errors out.

PR #590, opened 2021-06-08
Fix index dims parameter for tiledb from pandas

gsakkis:
Fix `tiledb.from_pandas` to support different index dimensions through the `index_dims` parameter.


ihnorton (approved):
Thank you!

PR #595, opened 2021-06-14
Add Array.open support for timestamp ranges

ihnorton:
tiledb.Array.open `timestamp` kwarg now supports a tuple of `(start: Int, end: Int)`
to open the array within a specific timestamp range.

Example:

```
    with tiledb.open(path, timestamp=(1234, 5678)) as A:
        ... # read A considering only fragments between
            #   (1234, 5678)
            # both endpoints inclusive
```

The existing `timestamp: Int` kwarg signature is unchanged.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -3769,9 +3769,27 @@ cdef ArrayPtr preload_array(uri, mode, key, timestamp, Ctx ctx=None):
         key_ptr = <void *> PyBytes_AS_STRING(bkey)
         #TODO: unsafe cast here ssize_t -> uint64_t
         key_len = <unsigned int> PyBytes_GET_SIZE(bkey)
-    cdef uint64_t _timestamp = 0
+    cdef uint64_t ts_start = 0
+    cdef uint64_t ts_end = 0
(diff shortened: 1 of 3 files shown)


tiledb/libtiledb.pyx
```diff
@@ -3769,24 +3769,51 @@ cdef ArrayPtr preload_array(uri, mode, key, timestamp, Ctx ctx=None):
...
+                raise ValueError("'timestamp' argument expects either int or tuple(start: int, end: int)")
+            if timestamp[0] is not None:
+                ts_start = <uint64_t>timestamp[0]
+                set_start = True
+            if timestamp[1] is not None:
+                ts_end = <uint64_t>timestamp[1]
+                set_end = True
+        elif isinstance(timestamp, int):
+            ts_end = <uint64_t> timestamp
+            set_start = True
```
Shelnutt2, line 3788: Should this be false?

ihnorton: Cruft, thanks for the catch. Will remove.

Shelnutt2 (approved):

nguyenv (approved):

PR #601, opened 2021-06-19
Check test output

ihnorton:
This PR adds a fixture to use pytest capturing in order capture all stdout, for two
use-cases:

  (1) by default assert that no output is written
  (2) if capfd is manually requested for test, we can get
      override the no_output assertion and check
      for expected output (when using functions that are expected
      to write to stdout)

Important details:
  - capfd does not flush, so we use libc.fflush to ensure that capfd
    actually gets all output
  - for test_examples, we need to add a no-op `no_output` signature
    which does not take capfd; otherwise, there will be a conflict
    with the usage of capsys in test_docs

### DIFF
--- a/tiledb/tests/common.py
+++ b/tiledb/tests/common.py
@@ -5,6 +5,7 @@ import os
 import random
 import shutil
 import subprocess
+import sys
 import tempfile
 import traceback
 import tiledb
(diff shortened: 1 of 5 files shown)


ihnorton:
- Disabled on windows because capfd did not seem to work correctly on Windows in CI. Linux/mac should be sufficient coverage, will debug at some point but low priority.
- Added test to test the fixture

nguyenv (approved):

PR #613, opened 2021-06-30
Use CIBW To Build Release Wheels on Linux

nguyenv:


misc/azure-libtiledb-darwin.yml
```diff
@@ -27,10 +27,15 @@ steps:
 
     $TILEDB_SRC/bootstrap --force-build-all-deps --disable-werror --enable=s3,gcs,azure,serialization --enable-static-tiledb --disable-avx2 --disable-tests --prefix=$TILEDB_INSTALL
 
+    export CXXFLAGS="-Wno-unused-parameter -lrt -DKJ_USE_EPOLL=0 -D__BIONIC__=1"
```
ihnorton, line 30: This should only be set for CIBW builds, eg with CIBW_ENV...

misc/azure-libtiledb-darwin.yml
```diff
@@ -59,8 +64,11 @@ steps:
     set -xe
     if [[ "$AGENT_OS" == "Windows_NT" ]]; then
       7z a ${TILEDB_INSTALL}/libtiledb-${LIBTILEDB_VERSION}-${LIBTILEDB_SHA} ${TILEDB_INSTALL}/*
+    elif [[ "$AGENT_OS" == "Darwin" ]]; then
+      tar -czf ${TILEDB_INSTALL}/libtiledb-${LIBTILEDB_VERSION}-${LIBTILEDB_SHA} -C ${TILEDB_INSTALL} . 
     else
-      tar -czf ${TILEDB_INSTALL}/libtiledb-${LIBTILEDB_VERSION}-${LIBTILEDB_SHA} -C ${TILEDB_INSTALL} .
+      # ignores the file change warning: https://stackoverflow.com/a/49112839
```
ihnorton, line 70: Do you know why this was happening? It makes me a little uneasy (eg some kind of race condition)

misc/azure-release.yml
```diff
@@ -27,14 +27,24 @@ stages:
...
+  - job: build1_libtiledb_on_linux
+    container: quay.io/pypa/manylinux2010_x86_64:2021-06-07-00faba2
+
+    steps:
+    - task: UsePythonVersion@0
+    - template: azure-libtiledb-darwin.yml
 
   - job: build2_python
-    dependsOn: build1_libtiledb
+    dependsOn: [build1_libtiledb, build1_libtiledb_on_linux]
```
ihnorton, line 39: Can we use `dependsOn` within each matrix entry so that we don't block all the builds?

ihnorton (approved):
One tweak and some minor questions. Thanks for seeing this through 👍

misc/azure-libtiledb-darwin.yml
nguyenv: Yeah it is suspicious for sure. I will look into it some more.

misc/azure-release.yml
nguyenv: https://github.com/microsoft/azure-pipelines-yaml/issues/420

This does not seem to be possible yet.

ihnorton: Ok. We could split into separate jobs and use a template for the rest of the steps, but probably not worth it unless we really notice a bottleneck.

PR #618, opened 2021-07-08
Implement Pickling for `SparseArray`

nguyenv:


ihnorton (approved):
Could parametrize the tests to avoid repetition, but will leave that up to you. Otherwise LGTM, thanks!

tiledb/libtiledb.pyx
```diff
@@ -4484,6 +4484,33 @@ cdef class Array(object):
...
+        cdef:
+            unicode uri, mode
+            object view_attr = None
+            object timestamp_range = None
+            object key = None
+            dict config_dict = {}
+        uri, mode, key, view_attr, _timestamp_range, config_dict = state
+
+        if mode == 'r':
+            timestamp_range = _timestamp_range
```
gatesn, line 4504: What is the reason for timestamp range only being preserved for reads?

ihnorton: I think that was (my) oversight in the previous dense-only implementation. We should support for writes as well. cc @nguyenv

PR #619, opened 2021-07-12
Temporary(?) fix of test_var_length broken on pandas 1.3.0

gsakkis:
I suspect this is a Pandas [bug](https://github.com/pandas-dev/pandas/issues/42501); for now just override the `copy` method of `RaggedArray`.

### DIFF
--- a/tiledb/tests/datatypes.py
+++ b/tiledb/tests/datatypes.py
@@ -17,12 +17,8 @@ class RaggedDtype(ExtensionDtype):
     type = np.ndarray
     na_value = None
 
-    def __init__(self, dtype=np.float64):
-        self._dtype = np.dtype(dtype)
-
-    @property
-    def subtype(self):
-        return self._dtype
+    def __init__(self, subtype=np.float64):
+        self.subtype = np.dtype(subtype)
 
     @property
     def name(self):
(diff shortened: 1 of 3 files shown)


ihnorton (approved):

tiledb/tests/datatypes.py
```diff
@@ -43,14 +39,10 @@ def construct_from_string(cls, string):
...
-    def __init__(self, arrays, dtype=None):
-        if isinstance(dtype, RaggedDtype):
-            self._dtype = dtype
-            dtype = dtype.subtype
-        else:
-            self._dtype = RaggedDtype(dtype)
-
-        self._flat_arrays = [np.asarray(array, dtype=dtype) for array in arrays]
+    def __init__(self, arrays, dtype):
+        assert isinstance(dtype, RaggedDtype)
```
ihnorton, line 43: Should we add a warning if `copy` is implemented in the superclass, so that we can remove it?

ihnorton: Nevermind, it's already implemented in `RaggedArray` and no way to detect changes [here](https://github.com/holoviz/datashader/blob/615b659f79336e0c65e7f4d898a63aeed1645fcf/datashader/datatypes.py#L593).

PR #625, opened 2021-07-16
Create Issue on GitHub if Nightly Test Build Fails

nguyenv:
See example of auto-created issue at https://github.com/nguyenv/TileDB-Py/issues/8


ihnorton (changes requested):
This is great. Before merging, we need to limit the permissions on the workflow to the minimum necessary. See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions -- I believe we need at least `issues: write`, not sure about other permissions.

ihnorton (approved):
🎆

PR #631, opened 2021-07-20
Modify Azure CI To Always Run on Latest Available Ubuntu and Windows Version

nguyenv:


ihnorton (changes requested):
I don't think we should change the macOS images, because we need to be able to support a minimum LSX_SDK_VERSION which may change with the OS version. There's no notification about deprecations right now, so IMO we should leave it alone for now.

I am less worried about Windows so IMO the change is fine and we can see what happens.

ihnorton (approved):

PR #632, opened 2021-07-21
Get `nonempty_domain` Directly From `_nonempty_domain_var`

nguyenv:
* This also fixes a bug where `nonempty_domain` was returning extents of dtype `np.array` instead of scalars.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -4308,36 +4308,31 @@ cdef class Array(object):
 
         """
         cdef Domain dom = self.schema.domain
-        dom_dims = [dom.dim(idx) for idx in range(dom.ndim)]
-        dom_dtypes = [dim.dtype for dim in dom_dims]
 
-        if any(dim.isvar for dim in dom_dims) or \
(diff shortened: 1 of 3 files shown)


tiledb/libtiledb.pyx
```diff
@@ -4308,36 +4308,31 @@ cdef class Array(object):
 
         """
         cdef Domain dom = self.schema.domain
-        dom_dims = [dom.dim(idx) for idx in range(dom.ndim)]
-        dom_dtypes = [dim.dtype for dim in dom_dims]
 
-        if any(dim.isvar for dim in dom_dims) or \
-                dom_dims.count(dom_dims[0].dtype) != len(dom_dims):
-            return self._nonempty_domain_var()
+        extents = self._nonempty_domain_var()
```
ihnorton, line 4312: Would you mind to inline this so it's just a single function? I don't think it needs to be separate anymore (unless I'm missing something).

ihnorton (changes requested):

tiledb/libtiledb.pyx
nguyenv: Will do.

ihnorton (approved):

PR #636, opened 2021-07-26
Add More Valid Python Syntax for `QueryCondition` Statements

nguyenv:
* Can now use bitwise & in addition to "and"
* Can do both lhs and rhs for operators (e.g. "3 < x and x < 5"
  can now be written as "3 < x < 5")

### DIFF
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -27,12 +27,12 @@ class QueryCondition(ast.NodeVisitor):
     A Boolean expression contains a comparison operator. The operator works on a
     TileDB attribute name and value.
 
-        bool_expr ::= attr compare_op val | val compare_op attr
+        bool_expr ::= attr compare_op val | val compare_op attr | val compare_op attr compare_op val
 
-    "and" is the only Boolean operator supported at the moment. We intend to
-    support "or" and "not" in future releases.
+    "and" and "&" are the only Boolean operators supported at the moment. We
+    intend to support "or" and "not" in future releases.
 
-        bool_op ::= and
+        bool_op ::= and | &
 
     All comparison operators are supported.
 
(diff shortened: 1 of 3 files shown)


tiledb/query_condition.py
```diff
@@ -165,21 +179,38 @@ def visit_Compare(self, node):
...
+        return result
+
+    def visit_BinOp(self, node):
+        AST_TO_TILEDB = {ast.BitAnd: qc.TILEDB_AND}
+
+        try:
+            op = AST_TO_TILEDB[type(node.op)]
+        except KeyError:
+            raise tiledb.TileDBError(
+                "Unsupported binary operator. Only & is currently supported."
```
ihnorton, line 205: Could we print the op here to make the error message a little bit more specific?

tiledb/tests/test_query_condition.py
```diff
@@ -65,16 +65,42 @@ def test_errors(self, input_array_UIDS):
 
         with self.assertRaises(tiledb.TileDBError):
             with tiledb.open(input_array_UIDS) as A:
-                qc = tiledb.QueryCondition("NO_CONDITION")
-                A.query(attr_cond=qc, use_arrow=False).df[:]
+                qc = tiledb.QueryCondition("U < 10000000000000000000000.0")
```
ihnorton, line 68: Would it make sense to check whether the op value can be losslessly cast to the target dtype?

ihnorton: (or maybe we do that already? this test would still throw b/c out of range, but a smaller value could be cast safely to uint32)

ihnorton (approved):
One minor question, and a potential (follow-up) change. This looks great 👍

tiledb/tests/test_query_condition.py
nguyenv: Yes we already do the cast if we can. If it were 10.0 instead of this large float then it would have been casted as 10.

PR #637, opened 2021-07-29
Addition Of `ascii` dtype For Attributes

nguyenv:


tiledb/libtiledb.pyx
```diff
@@ -1098,86 +1098,67 @@ def _tiledb_datetime_extent(begin, end):
...
-    elif tiledb_type == TILEDB_DATETIME_NS:
-        return np.datetime64('', 'ns')
-    elif tiledb_type == TILEDB_DATETIME_PS:
-        return np.datetime64('', 'ps')
-    elif tiledb_type == TILEDB_DATETIME_FS:
-        return np.datetime64('', 'fs')
-    elif tiledb_type == TILEDB_DATETIME_AS:
-        return np.datetime64('', 'as')
-    else:
+    tiledb_type_to_datetime = {
```
ihnorton, line 1112: Minor nit: this should be defined out of line, because otherwise the dictionary is constructed on every function call (I don't think Cython does any static optimization here, but I could be wrong - CPython does not).

tiledb/libtiledb.pyx
```diff
@@ -1098,86 +1098,67 @@ def _tiledb_datetime_extent(begin, end):
...
-        return TILEDB_DATETIME_NS
-    elif date_unit == 'ps':
-        return TILEDB_DATETIME_PS
-    elif date_unit == 'fs':
-        return TILEDB_DATETIME_FS
-    elif date_unit == 'as':
-        return TILEDB_DATETIME_AS
-    else:
+
+    tiledb_dtype_datetime = {
```
ihnorton, line 1142: same as above

tiledb/tests/test_libtiledb.py
```diff
@@ -471,6 +471,29 @@ def test_datetime_attribute(self):
...
+        unicode_data = ["±", "×", "÷", "√"]
+
+        with tiledb.open(path, "w") as A:
+            with self.assertRaises(tiledb.TileDBError):
+                A[np.arange(1, 5)] = unicode_data
+            A[np.arange(1, 5)] = ascii_data
+
+        with tiledb.open(path, "r") as A:
+            # print(A.schema)
+            assert A.schema.attr("A").dtype == np.bytes_
```
ihnorton, line 494: Please add check of the `schema.dump` and make sure the attribute is `TILEDB_STRING_ASCII` (can use capfd, there's a few other tests that use it for checking output).

ihnorton (changes requested):
Nice work (especially the cleanups 💯). Just a couple minor nits.

tiledb/libtiledb.pyx
```diff
@@ -1098,86 +1165,35 @@ def _tiledb_datetime_extent(begin, end):
...
-    elif tiledb_type == TILEDB_DATETIME_NS:
-        return np.datetime64('', 'ns')
-    elif tiledb_type == TILEDB_DATETIME_PS:
-        return np.datetime64('', 'ps')
-    elif tiledb_type == TILEDB_DATETIME_FS:
-        return np.datetime64('', 'fs')
-    elif tiledb_type == TILEDB_DATETIME_AS:
-        return np.datetime64('', 'as')
-    else:
+    if tiledb_type not in _tiledb_dtype_to_datetime_convert:
```
ihnorton, line 1179: This can be written as 

```
ck_type = _tiledb_dtype_to_datetime_convert.get(tiledb_type, None)
if not ck_ctype:
    raise ...
return ck_type
```
which avoids the extra lookup. It "probably" doesn't make a noticeable difference.

ihnorton (approved):

PR #640, opened 2021-08-05
Support Writing and Reading Variable-Sized, Nullable Attributes

nguyenv:


tiledb/libtiledb.pyx
```diff
@@ -5201,6 +5227,9 @@ def _setitem_impl_sparse(self: Array, selection, val, dict nullmaps):
                                      "typed attribute '{}'!".format(name))
 
                 attr_val = np.ascontiguousarray(attr_val, dtype=attr.dtype)
+            
+            if attr.isnullable and attr.name not in nullmaps:
+                nullmaps[attr.name] = np.array([int(v is None) for v in attr_val], dtype=np.uint8)
```
ihnorton, line 5232: Why do we need to do this? (also, calling `int(v is None) for v in attr_val` seems like it will be really really slow...)

tiledb/tests/test_pandas_dataframe.py
```diff
@@ -886,6 +886,7 @@ def check_array(path, df):
                 sparse=True,
             )
             df_to_check = copy.deepcopy(df2)
+            tmp = tiledb.open_dataframe(tmp_array2)
```
ihnorton, line 889: Should we check this result?

ihnorton (approved):

PR #642, opened 2021-08-06
Add Fragment Info API to `python-api.rst`

nguyenv:


doc/source/python-api.rst
```diff
@@ -144,6 +144,14 @@ Object Management
 .. autofunction:: tiledb.ls
 .. autofunction:: tiledb.walk
 
+Fragment Info
```
ihnorton, line 147: Would you mind to add some small usage examples either here or inline (eg as doctests)? Later on we can move those to the gitbook docs...

ihnorton:
cc @jp-dark

tiledb/fragment.py
```diff
@@ -12,6 +12,86 @@
...
+    :param cell_num: Number of cells in each fragment
+    :param timestamp_range: Timestamp range of when each fragment was written
+    :param dense: For each fragment, True if fragment is dense, else False
+    :param sparse: For each fragment, True if fragment is sparse, else False
+    :param has_consolidated_metadata: For each fragment, True if fragment has consolidated fragment metadata, else False
+    :param unconsolidated_metadata_num: Number of unconsolidated metadata fragments in each fragment
+    :param to_vacuum_num: Number of already consolidated fragments to vacuum
+    :param to_vacuum_uri: URIs of already consolidated fragments to vacuum
+
+    **Example:**
```
ihnorton, line 28: This might just be a local issue, but the examples here are not rendering for me.

tiledb/fragment.py
```diff
@@ -12,6 +12,86 @@
...
+    has consolidated metadata: False
+    nonempty domain: ((1, 2), (1, 4))
+    timestamp range: (1628273555397, 1628273555397)
+    version: 9
+    ===== FRAGMENT NUMBER 1 =====
+    is dense: True
+    cell num: 16
+    has consolidated metadata: False
+    nonempty domain: ((2, 3), (2, 3))
+    timestamp range: (1628273555409, 1628273555409)
```
ihnorton, line 86: Shouldn't this docstring test fail because the timestamps will be different every time it is run?

Also we might need to do this at the bottom to run as doctest?: https://stackoverflow.com/a/2708239

ihnorton (approved):
Thanks!

PR #648, opened 2021-08-19
Typo-fix in comment in multirange_indexing.py

johnkerl:
https://github.com/TileDB-Inc/TileDB-Py/blob/dev/examples/multirange_indexing.py#L56 is apparently an inadvertent copypasta from https://github.com/TileDB-Inc/TileDB-Py/blob/dev/examples/multi_attribute.py#L57


snagles (approved):

ihnorton:
Thanks

PR #653, opened 2021-08-25
Check `filter` Argument When Appending to `FilterList`

nguyenv:
* This fixes a bug where appending `None` causes a segfault


Shelnutt2 (approved):
Thanks for the quick fix!

PR #656, opened 2021-09-08
Return Empty Result With Empty Nonempty Domain With Multi-Index

nguyenv:
* Previously, this operation would error out.

### DIFF
--- a/tiledb/tests/test_multi_index.py
+++ b/tiledb/tests/test_multi_index.py
@@ -641,9 +641,7 @@ class TestMultiRange(DiskTestCase):
             data = A.multi_index[slice_index]
 
             assert_array_equal(data["a"], np.array([], dtype=np.uint64))
-
-            with self.assertRaises(tiledb.TileDBError):
-                A.multi_index[:]
+            assert_array_equal(A.multi_index[:], [])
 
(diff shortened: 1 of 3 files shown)


tiledb/tests/test_multi_index.py
```diff
@@ -760,3 +758,20 @@ def test_multi_index_with_implicit_full_string_range(self):
...
+        schema = tiledb.ArraySchema(domain=dom, sparse=True, attrs=[attr])
+        tiledb.Array.create(uri, schema)
+
+        with tiledb.open(uri, mode="w", timestamp=2) as A:
+            d1 = np.array(np.random.randint(1, 11, size=3, dtype=np.int32))
+            A[np.arange(1, 4)] = d1
+
+        with tiledb.open(uri, mode="r", timestamp=1) as A:
+            assert A.nonempty_domain() is None
+            assert_array_equal(A.multi_index[:][""], A[:][""])
```
ihnorton, line 775: What does this return value look like?

nguyenv: It is an empty list `[]`

ihnorton: Thanks

ihnorton (approved):
Thanks!

PR #662, opened 2021-09-10
Do Not Require `domain=(None, None)` For String Dimensions

nguyenv:


tiledb/libtiledb.pyx
```diff
@@ -2551,6 +2548,9 @@ cdef class Dim(object):
             # coordinates are written.
             dim_datatype = TILEDB_STRING_ASCII
         else:
+            if domain is None or len(domain) != 2:
```
ihnorton, line 2551: We need to handle `str` / `"U*"` dtypes as well - this only works for `bytes`.

ihnorton: (also IMO `domain=(None,None)` is still _valid_ just not required, so let's keep the `len` check too)

ihnorton (changes requested):

tiledb/libtiledb.pyx
nguyenv: OK I see.

ihnorton (approved):
LGTM, thanks

PR #663, opened 2021-09-11
Improve Checking For Byte-String Attributes

nguyenv:
* Errors out when `var=True` but the string is a fixed-width type and when `var=False` but the string is a variable-length type.


ihnorton (approved):
Thanks!

PR #665, opened 2021-09-11
Print warning for ipykernel < 6 when using scope_ctx or default_ctx

ihnorton:
(part of ch7578)

### DIFF
--- a/tiledb/ctx.py
+++ b/tiledb/ctx.py
@@ -5,6 +5,33 @@ import tiledb
 
 _ctx_var = ContextVar("ctx")
 
+already_warned = False
+
+
+def check_ipykernel_warn_once():
+    """
+    This function checks if we have imported ipykernel version < 6 in the
+    current process, and provides a warning that default_ctx/scope_ctx will
+    not work correctly due to a bug in IPython contextvar support."""
+    global already_warned
+    if not already_warned:
+        try:
+            import sys, warnings
+
+            if "ipykernel" in sys.modules and tuple(
+                map(int, sys.modules["ipykernel"].__version__.split("."))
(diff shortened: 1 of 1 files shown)


nguyenv (approved):

tiledb/ctx.py
```diff
@@ -5,6 +5,33 @@
...
+already_warned = False
+
+
+def check_ipykernel_warn_once():
+    """
+    This function checks if we have imported ipykernel version < 6 in the
+    current process, and provides a warning that default_ctx/scope_ctx will
+    not work correctly due to a bug in IPython contextvar support."""
+    global already_warned
+    if not already_warned:
```
gsakkis, line 17: Normally this wouldn't be necessary, the warnings mechanism issues a warning only once for each location (module + line number) where the warning is issued. Apparently IPython [deliberately changed this](https://github.com/ipython/ipython/issues/11207) with the reasoning that [people often don't see important warnings when running code in IPython, because (to a first approximation) any given warning
will only issue once per session](https://github.com/ipython/ipython/pull/6680#issue-22598608). I guess it's debatable what's the most sensible behavior here.

ihnorton: True; but i also didn’t really want to run the check code every single time, and a guard is easy enough.

gsakkis: The condition can be determined once globally, that's not a problem. The (theoretical at least) issue is It kind of goes against the spirit of the warning module that (just like logging) separates what to warn about (library producer) and when/how/if to display it (library consumer). Anyway practicality beats purity, your call if you think the repeated warning is more of a nuisance than help.

gsakkis (approved):

PR #668, opened 2021-09-13
Add wrapper class and docs for ArraySchemaEvolution

ihnorton:
(also add history for #665)


ihnorton:
The first actions run failed due to a network issue downloading dep. Second run all jobs succeeded, but it hasn't updated here: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/1230222565

nguyenv (approved):

PR #671, opened 2021-09-13
Nightly Test Wheels on Azure

nguyenv:


ihnorton (commented):
Let's see if we can create an issue on failure by triggering from azure pipelines: https://medium.com/swlh/trigger-github-actions-with-azure-devops-pipelines-4f0791e6cbbf

(but ok to punt for now if that looks like it will be too involved)

nguyenv:
As per the docs [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#repository_dispatch), `repository_dispatch` events only trigger a workflow run if the workflow file is on the default branch. Thus, testing was done on [this](https://github.com/nguyenv/TileDB-Py-Test-REST) separate test repo as to not push to TileDB-Py's `dev` branch here.

The `repository_dispatch` is properly triggered [here](https://github.com/nguyenv/TileDB-Py-Test-REST/actions) and automatically opens the issues [here](https://github.com/nguyenv/TileDB-Py-Test-REST/issues).

.github/workflows/daily-test-build.yml
```diff
@@ -1,13 +1,31 @@
 name: Daily Test Build TileDB-Py Against Core
 
-#on: [push]
+on: [push]
 
-on:
-  schedule:
-    # runs every day at 5:00 UTC (1:00AM EST / Midnight CST)
-    - cron: "0 5 * * *"
+# on:
```
ihnorton, line 5: un-comment?

ihnorton (approved):
Nice work! 💯

.github/workflows/daily-test-build.yml
```diff
@@ -1,13 +1,31 @@
...
+    - name: Get current date
+      id: date
+      run: echo "::set-output name=date::$(date +'%a-%Y-%m-%d')"
+    - name: Create Test Branch for Azure Wheel Nightly Build
+      uses: peterjgrainger/action-create-branch@v2.0.1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        branch: 'azure-wheel-test-${{ steps.date.outputs.date }}'
+    - name: Clean Up Test Branch
```
ihnorton, line 23: Maybe this part should run un-conditionally on the repo trigger event? Just wondering if there is any risk of a race condition where the branch has been deleted before the azp job starts.

misc/azure-release.yml
```diff
@@ -172,3 +172,22 @@ stages:
...
+    - task: InvokeRESTAPI@1
+      inputs:
+        connectionType: 'connectedServiceName'
+        connectedServiceName: 'GitHub-TileDB-Py'
+        method: 'POST'
+        body: |
+          {
+          "event_type": "failed"
+          }
+        urlSuffix: 'repos/TileDB-Inc/TileDB-Py/dispatches'
```
ihnorton, line 190: Can this repo name be conditional so that we don't accidentally trigger notifications when running on a fork in the future?

nguyenv: I tried doing this by using the `Build.Repository.Name` variable to see if it matched `TileDB-Inc/TileDB-Py` but haven't been able to get it to work.

ihnorton: Does substitution work here? If we can make it point at the originating repository then that solves the issue - allows testing on a fork w/out impacting the main repo.

ihnorton: Actually, let's get this merged as-is and we can do that in follow-up.

ihnorton (approved):

PR #694, opened 2021-09-24
Fix Linker Warning macOS 10.14 vs 10.15

nguyenv:
* Fixes linker warnings for incompatible 10.15 vs 10.14


ihnorton (approved):
This is ok, but we may not have to use the older image: I haven't looked closely, but I think the alternative would be to set the `MACOSX_DEPLOYMENT_TARGET` correctly everywhere (assuming it is missing). Your call though.

Issue #339, opened 2020-06-17
Option for rewrite on .array.create


michael-imbeault:
Getting started with TileDB - it would be nice if there was a parameter you could pass to the array .create functions (on DenseArray and SparseArray) - current behavior is to fail if the file exists:

[TileDB::StorageManager] Error: Cannot create array; Array 'file:///mnt/d/quickstart_sparse' already exists

This is sensible as a default, but is there an option to force a reset? Zarr has an optional 'overwrite' boolean parameter.

stavrospapadopoulos:
+1 

Just noting here though that this will be a bit tricky on S3 due to eventual consistency, which is more pronounced particularly with deletes.

nguyenv:
We have added an `overwrite` parameter to `tiledb.Array.create`. See an example of usage [here](https://github.com/TileDB-Inc/TileDB-Py/blob/57dcffecb44f0ee79ef4c7eb6860f67f517607a9/tiledb/tests/test_libtiledb.py#L1000).

PR #713, opened 2021-10-07
Addition of `overwrite` Param to `Array.create`

nguyenv:
* For local arrays only, if it already exists, allow recreating the array by passing the `overwrite` parameter


nguyenv:
#339

PR #725, opened 2021-10-14
Create Issue if GHA Daily Test Is "Cancelled"

nguyenv:
* The following daily run was cancelled by GitHub due to exceeding the time limit:
	https://github.com/TileDB-Inc/TileDB-Py/runs/3890774602?check_suite_focus=true#step:8:73
* Since this cancellation was done by GH and not by the user, it should really count
  as a failure. There will very unlikely be an instance in which we purposely go in to cancel
  a nightly test ourselves.


ihnorton:
Nice catch!

PR #744, opened 2021-10-18
Don't decrement the offset count when no results are returned

ihnorton:
~Fixes part of sc-11077~ Fixes sc-11239


ihnorton:
This PR requires: https://github.com/TileDB-Inc/TileDB/pull/2540 so will have to hold it until next patch release, or x-fail the test for non-nightly PRs.

Issue #710, opened 2021-10-06
Impossibile to install using pip on Apple Silicon


lucalianas:
Hi, I can't install TileDB using pip when running Python3.9 version compiled for Apple Silicon hardware (Python3.9 was installed using Homebrew)

This is the output I get from pip

```
❯ pip install tiledb
Collecting tiledb
  Using cached tiledb-0.10.2.tar.gz (232 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Collecting numpy>=1.16.5
  Using cached numpy-1.21.2-cp38-cp38-macosx_11_0_arm64.whl (12.3 MB)
Building wheels for collected packages: tiledb
  Building wheel for tiledb (PEP 517) ... error
  ERROR: Command errored out with exit status 1:
   command: /Users/luca/work/python_env/promort_tools/bin/python /Users/luca/work/python_env/promort_tools/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py build_wheel /var/folders/23/q0syrgm57qgfffdmdh6mnffw0000gn/T/tmpf_vbra8q
       cwd: /private/var/folders/23/q0syrgm57qgfffdmdh6mnffw0000gn/T/pip-install-gmstgq04/tiledb_8c853fe44b4141e88dff3b4b536c80a0
  Complete output (23 lines):
  TILEDB_PATH from env: ''
  Traceback (most recent call last):
    File "/Users/luca/work/python_env/promort_tools/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 349, in <module>
      main()
    File "/Users/luca/work/python_env/promort_tools/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 331, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/Users/luca/work/python_env/promort_tools/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 248, in build_wheel
      return _build_backend().build_wheel(wheel_directory, config_settings,
    File "/private/var/folders/23/q0syrgm57qgfffdmdh6mnffw0000gn/T/pip-build-env-bti5b7wd/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 221, in build_wheel
      return self._build_with_temp_dir(['bdist_wheel'], '.whl',
[...]

PR #747, opened 2021-10-19
Enable serialization by default in (lib)TileDB source builds

ihnorton:
For https://github.com/TileDB-Inc/TileDB-Py/issues/710


ihnorton:
> Can you also change the files under .github/workflows?

Done, thanks! (I grep'd but that was excluded, thanks for the catch)

PR #774, opened 2021-11-08
Addition of `tiledb.delete_fragments`

nguyenv:


tiledb/highlevel.py
```diff
@@ -120,6 +120,42 @@ def array_fragments(uri, ctx=None):
...
+        if (
+            timestamp_range[0] <= frag.timestamp_range[0]
+            and frag.timestamp_range[1] <= timestamp_range[1]
+        ):
+            if verbose or dry_run:
+                print(frag.uri)
+
+            if not dry_run:
+                tiledb.VFS().remove_file(f"{frag.uri}.ok")
+                tiledb.VFS().remove_dir(frag.uri)
```
Shelnutt2, line 156: This creates a new VFS object for each fragment. Can you create one instance outside of the for loop? Also you will need to pass config/context parameters to vfs.

nguyenv: Oic. Will modify this in a bit.

tiledb/highlevel.py
```diff
@@ -120,6 +120,54 @@ def array_fragments(uri, ctx=None):
...
+        )
+
+    if verbose or dry_run:
+        print("Deleting fragments...")
+
+    if not ctx:
+        ctx = tiledb.default_ctx()
+
+    if config is None:
+        config = tiledb.Config()
```
Shelnutt2, line 154: If the config is empty, should we set it to the ctx's config? Or leave it as none for vfs?

Shelnutt2 (commented):
Thanks for the adjustments! One more question

tiledb/highlevel.py
nguyenv: Thanks for catching this. You're right it should be `ctx.config()`.

Shelnutt2 (approved):

tiledb/highlevel.py
```diff
@@ -120,6 +120,53 @@ def array_fragments(uri, ctx=None):
...
+def delete_fragments(
+    uri, timestamp_range, config=None, ctx=None, verbose=False, dry_run=False
+):
+    """
+    Delete fragments from an array located at uri that fall within a given
+    timestamp_range.
+
+    :param str uri: URI for the TileDB array (any supported TileDB URI)
+    :param config: Override the context configuration. Defaults to ctx.config()
+    :param (int, int) timestamp: (default None) If not None, vacuum the array
```
ihnorton, line 132: Nit: `timestamp` vs `timestamp_range` in the signature.

tiledb/tests/test_libtiledb.py
```diff
@@ -3923,6 +3923,32 @@ def test_ctx_thread_cleanup(self):
...
+        def create_array(target_path, dshape):
+            dom = tiledb.Domain(tiledb.Dim(domain=dshape, tile=len(dshape)))
+            att = tiledb.Attr(dtype="int64")
+            schema = tiledb.ArraySchema(domain=dom, attrs=(att,), sparse=True)
+            tiledb.libtiledb.Array.create(target_path, schema)
+
+        def write_fragments(target_path, dshape, num_writes):
+            for i in range(1, num_writes + 1):
+                with tiledb.open(target_path, "w", timestamp=i) as A:
+                    A[[1, 2, 3]] = np.random.rand(dshape[1])
```
ihnorton, line 3939: Would suggest adding a functional test here as well. Write a single coordinate at a time based on `i`, and then do a read at the end and make sure the deleted coordinates are gone. eg `A[[i]] = ...` in the loop.

Then, before delete we would expect coordinates 1..10, but after delete we should get 1,2 and 7,8,9,10.

ihnorton (approved):
Minor comments, otherwise LGTM. Thanks!

PR #776, opened 2021-11-12
Release 0.10.5

nguyenv:


misc/azure-release.yml
```diff
@@ -6,7 +6,7 @@ stages:
       LIBTILEDB_VERSION: dev
       LIBTILEDB_SHA: dev
     ${{ else }}: 
-      TILEDBPY_VERSION: 0.10.4
+      TILEDBPY_VERSION: 0.10.5
       LIBTILEDB_VERSION: 2.4.3
       LIBTILEDB_SHA: 5cb082458f6e4214d5ae5beeeb667ec2231991ab
```
ihnorton, line 11: Need to update the SHA too

ihnorton: wait nvm

ihnorton (approved):

misc/azure-release.yml
nguyenv: SHA should be the same since it's still on 2.4.3.

PR #777, opened 2021-11-12
Addition of `array_schema_name` Getter to Fragment Info API; Addition of `create_array_from_fragments`

nguyenv:


nguyenv:
I rebased this branch with the MBR branch to make it easier to merge everything into `dev` in the future, so there are "extra" commits.

tiledb/highlevel.py
```diff
@@ -164,19 +163,114 @@ def delete_fragments(
...
+            schema_name = os.path.join("__schema", schema_name)
+        src_schema = os.path.join(src_uri, schema_name)
+        dst_schema = os.path.join(dst_uri, schema_name)
+
+        if verbose or dry_run:
+            print(f"Copying schema `{src_schema}` to `{dst_schema}`\n")
+
+        if not dry_run:
+            if is_new_style_schema:
+                vfs.create_dir(os.path.join(dst_uri, "__schema"))
```
ihnorton, line 259: I think this needs to be conditional on whether the target exists. There could be multiple fragments referring to the same schema file, and I'm not sure if this `create_dir` will error out if the target already exists.

ihnorton: and the copy below too

ihnorton (changes requested):

ihnorton:
Two things for testing in follow up:

- use schema evolution to create multiple schema versions, along with corresponding writes at those versions. Test that `create_array_from_fragments` works correctly.
- we should also test creation in the presence of old style schema (or error out). one way to do this is to create an array with older tiledb-py version, then evolve it with a newer version, then test `create_array_from_fragments` (the old array can be stored as bytes inline in the test file, tar -> base64 encode; there are several existing compatibility test arrays like this)

tiledb/highlevel.py
```diff
@@ -164,19 +163,118 @@ def delete_fragments(
...
+        vfs.copy_file(f"{src_lock}", f"{dst_lock}")
+
+    for frag in fragment_info:
+        if not (
+            timestamp_range[0] <= frag.timestamp_range[0]
+            and frag.timestamp_range[1] <= timestamp_range[1]
+        ):
+            continue
+
+        schema_name = fragment_info.array_schema_name[0]
```
ihnorton, line 247: Need to use `frag.array_schema_name`. Currently this is only copying the first array schema version, so the array is corrupted after copy.

ihnorton (changes requested):
A couple change requests:
-  bug inline w/ schema copy
- error out if the target directory exists (currently does not error until the fragment copy step)

For follow up:
- would be nice to eventually support half-open intervals, eg `(start,None)`, currently that's a little bit fiddly (have to manually check/specify the fragment timestamps on both sides)

tiledb/highlevel.py
```diff
@@ -164,19 +163,128 @@ def delete_fragments(
...
+        print(f"Copying lock file {dst_uri}\n")
+
+    if not dry_run:
+        vfs.copy_file(f"{src_lock}", f"{dst_lock}")
+
+    list_new_style_schema = [ver >= 10 for ver in fragment_info.version]
+    is_mixed_versions = len(set(list_new_style_schema)) > 1
+    if is_mixed_versions:
+        raise tiledb.TileDBError(
+            "This array contains a mix of old and new style schemas"
```
ihnorton, line 247: Please clarify - "cannot copy fragments"

ihnorton (approved):
💯

PR #797, opened 2021-11-23
Update CMake pin to [3.21,3.22]

ihnorton:


ihnorton:
wheel build test: https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=17364&view=results

PR #813, opened 2021-12-02
`delete_fragments` Needs To Also Delete Unused Schemas

nguyenv:
* Move fragment functions from `highlevel.py` to `fragment.py`
* Move fragment tests to `test_fragments.py`


ihnorton (approved):

ihnorton:
cc @bdeng-xt @Shelnutt2 -- we think this makes sense, but FYI in case you see any issue with it:  we are deleting `__schema` files which are unreferenced after removing fragments within a timestamp range. Otherwise, if we evolve an array to add an attribute, then remove all fragments with that new attribute, when we read the array libtiledb will return the new attribute w/ fill values.

bdeng-xt:
It makes sense to me

PR #824, opened 2021-12-10
Add Rich Display for Various TileDB Objects in Jupyter Notebook

nguyenv:


nguyenv:
reprhtml

ihnorton (changes requested):
Could you please left-justify the (key, value) table listings, eg for fragment info in the first screenshot? I think that better matches what I've seen for other HTML rendering, and it should be easier to read when the column is widened by a path.

ihnorton:
> when the column is widened by a path

Could potentially truncate long URIs in the future, and show the full thing on hover only.

ihnorton:
cc @jp-dark

ihnorton (approved):

PR #834, opened 2021-12-14
Bulk range-set optimization

ihnorton:
Python side of https://github.com/TileDB-Inc/TileDB/pull/2738

### DIFF
--- a/tiledb/core.cc
+++ b/tiledb/core.cc
@@ -593,6 +593,36 @@ public:
     }
   }
 
+#if TILEDB_VERSION_MAJOR == 2 && TILEDB_VERSION_MINOR == 6
+  void set_ranges_bulk(py::iterable ranges) {
+    // ranges are specified as one iterable per dimension
+
+    uint32_t dim_idx = 0;
+    for (auto dim_range : ranges) {
+      // py::print(dim_range);
+      if (py::isinstance<py::array>(dim_range)) {
+        py::array r_array = dim_range.cast<py::array>();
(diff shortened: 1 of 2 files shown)


tiledb/core.cc
```diff
@@ -593,6 +593,36 @@ class PyQuery {
     }
   }
 
+#if TILEDB_VERSION_MAJOR == 2 && TILEDB_VERSION_MINOR == 6
```
nguyenv, line 596: `TILEDB_VERSION_MINOR >= 6`?

tiledb/multirange_indexing.py
```diff
@@ -184,6 +187,9 @@ def __getitem__(self, idx: Any) -> Dict[str, np.ndarray]:
             if idx is EmptyRange:
                 return _get_empty_results(self.array.schema, self.query)
 
+            # if any([lambda x: isinstance(x, np.ndarray), idx]):
+            #    self.ranges = tuple(x for x in idx)
+            # else:
```
nguyenv, line 192: need to rm?

nguyenv (approved):

tiledb/core.cc
ihnorton: The API is currently experimental so I wanted to pin it to 2.6 for now.

tiledb/multirange_indexing.py
ihnorton: Yes, thanks

PR #849, opened 2021-12-21
`test_import_buffer` Trim Trailing Zeros in Byte String

nguyenv:
- Use `dtype="U0"` for string attrs


nguyenv:
I have not been able to reproduce this locally either. But I notice on GHA twice it has been mismatching due to trailing zeros.

https://github.com/TileDB-Inc/TileDB-Py/runs/4556657693?check_suite_focus=true#step:8:470
https://github.com/TileDB-Inc/TileDB-Py/runs/4590963875?check_suite_focus=true#step:8:471

PR #854, opened 2021-12-23
Support Writes to Zero Attribute Array

nguyenv:
* Allow writing to dimension-only array (zero attributes) by using assignment to `None`, for example: `A[coords] = None` (given `A: tiledb.Array`)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,4 +1,10 @@
+# TileDB-Py 0.11.6 Release Notes
+
+## API Changes
+* Allow writing to a non-attribute array by setting `tiledb.Array(coords) = None` [#854](https://github.com/TileDB-Inc/TileDB-Py/pull/854)
+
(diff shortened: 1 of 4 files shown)


nguyenv:
Would also fix Shortcut story 13073

HISTORY.md
```diff
@@ -1,4 +1,10 @@
+# TileDB-Py 0.11.6 Release Notes
+
+## API Changes
+* Allow writing to a non-attribute array by setting `tiledb.Array(coords) = None` [#854](https://github.com/TileDB-Inc/TileDB-Py/pull/854)
```
ihnorton, line 4: Suggest slight tweak: 
```
Allow writing to dimension-only array (zero attributes) by using assignment to `None`, for example: `A[coords] = None` (given `A: tiledb.Array`)
```

ihnorton (approved):

PR #866, opened 2022-01-10
Fix Attribute View for `multi_index`

nguyenv:


ihnorton (approved):
I don't remember if this is supposed to work for dimensions? Please double-check, and make sure we error out appropriately for a dimension if not. Thanks!

PR #882, opened 2022-01-18
Fix leak of C++ Array object in PyQuery

ihnorton:
REQUIRES https://github.com/TileDB-Inc/TileDB/pull/2823


nguyenv (approved):

ihnorton:
rebased against #894

PR #886, opened 2022-01-19
Cast Operands in `QueryCondition` Using attr() or val()

nguyenv:
* Previously, there was no way to apply query conditions to attributes
  with spaces. With the addition of the attr() cast, it is now possible to
  do `tiledb.QueryCondition("attr('attributes with spaces') == 'val'))`

### DIFF
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -1,5 +1,7 @@
 import ast
+from dataclasses import dataclass, field
 import numpy as np
+from typing import List, Union
 
 import tiledb
 import tiledb.main as qc
@@ -11,7 +13,8 @@ filtering query results on attribute values.
 """
 
 
-class QueryCondition(ast.NodeVisitor):
+@dataclass
+class QueryCondition:
     """
     Class representing a TileDB query condition object for attribute filtering
     pushdown. Set the query condition with a string representing an expression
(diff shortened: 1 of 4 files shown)


Shelnutt2 (approved):
Awesome! Thanks for supporting this!

tiledb/query_condition.py
```diff
@@ -234,6 +242,21 @@ def visit_BoolOp(self, node):
...
+
+        if node.func.id not in visit_cast:
+            raise tiledb.TileDBError(f"Valid casts are attr() or val().")
+
+        if len(node.args) != 1:
+            raise tiledb.TileDBError(
+                f"Exactly one argument must be provided to {node.func.id}()."
+            )
+
+        node.args[0].qc_type = node.func.id
```
ihnorton, line 256: Can we use `Optional` here and declare the maybe-contained type?

ihnorton (approved):

tiledb/query_condition.py
```diff
@@ -113,11 +158,29 @@ def aux_visit_Compare(self, att, op, val):
...
+    def is_att_node(self, att: QueryConditionNodeElem) -> bool:
+        return isinstance(att, ast.Name) or (
+            (
+                (
+                    isinstance(att, ast.Constant)
+                    or isinstance(att, ast.Num)
+                    or isinstance(att, ast.Str)
+                    or isinstance(att, ast.Bytes)
+                )
+                and hasattr(att, "qc_type")
```
ihnorton, line 176: I think I need to better understand what `qc_type` is indicating here.

ihnorton: (let's discuss in meeting later)

tiledb/query_condition.py
```diff
@@ -10,8 +12,23 @@
 filtering query results on attribute values.
 """
 
+QueryConditionNodeElem = TypeVar(
```
ihnorton, line 15: IIUC this should be a `Union` rather than a `TypeVar`, because the in/out types can be different.

ihnorton (approved):

PR #898, opened 2022-01-26
Add Setter and Getter for Validity Filter

nguyenv:


tiledb/tests/test_libtiledb.py
```diff
@@ -745,10 +757,12 @@ def test_sparse_schema_filter_list(self, capfd):
             tile_order="row-major",
             coords_filters=coords_filters_pylist,
             offsets_filters=off_filters,
+            validity_filters=validity_filters,
             sparse=True,
         )
         self.assertEqual(len(schema2.coords_filters), 1)
         self.assertEqual(len(schema2.offsets_filters), 1)
+        self.assertEqual(len(schema2.validity_filters), 1)
```
ihnorton, line 765: Do you mind to also add a check that the filter type matches what we put in?

ihnorton (approved):
Minor test nit, otherwise LGTM. Thanks!

PR #910, opened 2022-02-03
Add `QueryCondtion` Unit Test `test_var_length_str`

nguyenv:
* Tests strings greater than len(1)
* Tests both "ascii" `and np.bytes_`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ from pkg_resources import resource_filename
 from setuptools import Extension, find_packages, setup
 
 # Target branch
-TILEDB_VERSION = "2.6.2"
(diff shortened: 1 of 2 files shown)


ihnorton (approved):
👍 LGTM.

setup.py
```diff
@@ -15,7 +15,7 @@
 from setuptools import Extension, find_packages, setup
 
 # Target branch
-TILEDB_VERSION = "2.6.2"
+TILEDB_VERSION = "dev"
```
ihnorton, line 18: (nit just in case: remove before merge 😄 )

PR #921, opened 2022-02-18
PyBind11 Wrap `FilterList` and `Filter` Classes

nguyenv:


tiledb/cc/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.20)
```
ihnorton, line 1: This file is cruft and can be removed now that we have top-level setup.py support 😄

tiledb/filter.py
```diff
@@ -0,0 +1,667 @@
...
+from typing import overload, List, Sequence, Union
+
+import tiledb.cc as lt
+from .ctx import default_ctx
+
+
+class Filter(lt.Filter):
+    """Base class for all TileDB filters."""
+
+    from .libtiledb import Ctx
```
nguyenv, line 11: Use `from typing import TYPE_CHECKING`

tiledb/libtiledb.pxd
```diff
@@ -1256,18 +1256,18 @@ cdef class ConfigValues(object):
 cdef class Ctx(object):
     cdef tiledb_ctx_t* ptr
 
-cdef class Filter(object):
-    cdef Ctx ctx
-    cdef tiledb_filter_t* ptr
+# cdef class Filter(object):
```
ihnorton, line 1259: Delete rather than comment.

tiledb/cc/filter.cc
```diff
@@ -0,0 +1,61 @@
...
+             // TODO check_type(a.dtype)
+             // Use the C API here because we are doing typecheck
+             ctx.handle_error(tiledb_filter_get_option(
+                 ctx.ptr().get(), filter.ptr().get(), option, &level));
+             return level;
+           });
+
+  py::class_<FilterList>(m, "FilterList")
+      .def(py::init<const Context &>(), py::keep_alive<1, 2>())
+      .def(py::init<const Context &, py::capsule>(), py::keep_alive<1, 2>())
```
ihnorton, line 43: Does this take ownership of the pointer in the capsule? I'm assuming yes, because `own = True` in this constructor, but want to be sure...

tiledb/libtiledb.pyx
```diff
@@ -2911,7 +2028,8 @@ cdef class Dim(object):
         check_error(self.ctx,
                     tiledb_dimension_get_filter_list(self.ctx.ptr, self.ptr, &filter_list_ptr))
 
-        return FilterList.from_ptr(filter_list_ptr, self.ctx)
+        return FilterList(PyCapsule_New(filter_list_ptr, "fl", NULL),
```
ihnorton, line 2031: Extra spaces at end of line, I think?

ihnorton (approved):
Small nits and then let's get this merged :rocket: Really nice work on passing FilterList between cython and pybind11, it looks really clean!

tiledb/cc/filter.cc
nguyenv: Yes it does. I have it set to `own = True`.

PR #924, opened 2022-02-19
Build With Serialization Disabled

nguyenv:


ihnorton (approved):
Thanks, LGTM.

If there are other situations like this in the future where we need to have conditional compilation of bindings, we might want to put in a pure python stub which raises an explanatory error -- but I don't think anything in `serialization.cc` is really user-facing, right?

nguyenv:
> Thanks, LGTM.
> 
> If there are other situations like this in the future where we need to have conditional compilation of bindings, we might want to put in a pure python stub which raises an explanatory error -- but I don't think anything in `serialization.cc` is really user-facing, right?

No, nothing in `serialization.cc` is user facing.

PR #930, opened 2022-02-25
Deprecate Partial Vacuuming

nguyenv:


HISTORY.md
```diff
@@ -1,5 +1,13 @@
 # TileDB-Py 0.13.0 Release Notes
 
+## API Changes
+* Deprecate partial vacuuming []()
+
+# TileDB-Py 0.12.4 Release Notes
```
nguyenv, line 6: Need to see where I messed up the README here...

HISTORY.md
```diff
@@ -1,12 +1,16 @@
 # TileDB-Py 0.13.0 Release Notes
 
-## Bug Fixes
-* Enable building with serialization disabled [#924](https://github.com/TileDB-Inc/TileDB-Py/pull/924)
-* Do not print out `FragmentInfo_frags` for `repr` [#925](https://github.com/TileDB-Inc/TileDB-Py/pull/925)
+## API Changes
+* Deprecate partial vacuuming []()
```
ihnorton, line 4: ^ reminder

ihnorton (changes requested):
IIUC, this change was implemented in core by https://github.com/TileDB-Inc/TileDB/pull/2909

That PR will not be backported, so this functionality will still be available in 0.13 / libtiledb 2.7. So I think we should keep the functionality in place with the deprecation warning (it looks like this PR replaces the functionality with a warning?)

Also needs rebase.

ihnorton (approved):

PR #934, opened 2022-02-26
PyBind11 Wrap `FileIO`, `FileHandle`, and `VFS`

nguyenv:


tiledb/vfs.py
```diff
@@ -0,0 +1,260 @@
...
+            warnings.warn(
+                f"`tiledb.VFS().open` now returns a a FileIO object. Use "
+                "`FileIO.write`.",
+                DeprecationWarning,
+            )
+        if isinstance(buff, str):
+            buff = buff.encode()
+        file.write(buff)
+
+    def read(self, file: lt.FileHandle, offset: int, nbytes: int):
```
ihnorton, line 97: Type signature? `-> bytes`?

tiledb/vfs.py
```diff
@@ -0,0 +1,260 @@
...
+        :param str scheme: scheme component of a VFS resource URI (ex. 'file' / 'hdfs' / 's3')
+        :rtype: bool
+        :return: True if the linked libtiledb version supports the storage backend, False otherwise
+        :raises ValueError: VFS storage backend is not supported
+
+        """
+        if scheme == "file":
+            return True
+
+        scheme_to_fs_type = {
```
ihnorton, line 133: We need a test to check whether this is exhaustive by looking at the underlying enum. Can be done in follow-up (pls create story).

ihnorton (approved):
Some notes for follow-up, but LGTM. Also we need to add docstrings for FileIO in the future.

tiledb/vfs.py
ihnorton: (also, hard-coding this here is not ideal -- opened SC-16386 to add an API in core we can use instead)

PR #949, opened 2022-03-04
Addition of `ArraySchema.version`

nguyenv:


tiledb/tests/test_libtiledb.py
```diff
@@ -587,25 +587,25 @@ def test_dense_array_schema(self):
...
-        )
+        assert schema.sparse == False
+        assert schema.cell_order == "row-major"
+        assert schema.tile_order == "row-major"
+        assert schema.domain == domain
+        assert schema.ndim == 2
+        assert schema.shape == (8, 8)
+        assert schema.nattr == 1
+        assert schema.domain.homogeneous == True
+        assert hasattr(schema, "version")  # don't pin to a specific version
```
ihnorton, line 598: We could look at a fragment (eg `_10`) and check that it matches `.version`, but not strictly necessary to do that.

ihnorton (approved):
Thanks! One suggestion but your call on that (or defer and create an issue).

ihnorton:
Labeled as `backport-0.13`, which will need to be done manually onto `release-0.13` branch since we don't have a bot set up right now.

PR #966, opened 2022-03-16
Correct Disable Pandas Testing for GHA Nightly Builds

nguyenv:


nguyenv:
Proof of fix:
https://github.com/TileDB-Inc/TileDB-Py/actions/runs/1994204703

.github/workflows/daily-test-build.yml
```diff
@@ -1,7 +1,5 @@
 name: Daily Test Build TileDB-Py Against Core
 
-#on: [push]
```
ihnorton, line 3: Should this be removed completely?

ihnorton (approved):
Seems good based on the successful run. I'm a little confused by what's actually changing here due to the indentation changes -- are those necessary for the fix? Why weren't we getting a syntax error before? (or were we?)

PR #977, opened 2022-03-18
Set `fail-fast` To `false`

nguyenv:
* By default on GitHub actions, all jobs are cancelled when a job fails. 
  This setting makes it so that other jobs will continue to run.


ihnorton:
Do we want this for all CI or only nightlies?

ihnorton (approved):

nguyenv:
> Do we want this for all CI or only nightlies?

I think it would be useful for both. I think it's annoying when it fails only one of the jobs on CI, and it cancels everything.

PR #995, opened 2022-03-25
Fix default validity for non-pandas write to nullable attribute

ihnorton:


nguyenv (approved):

ihnorton:
Hi @flying-sheep, this was intended to be a temporary work-around for `https://github.com/pypa/setuptools/issues/3196`. Removing in #1322.

PR #998, opened 2022-03-26
Fix `repr` For `datetime` Dimensions By Casting `tile` To `int`

nguyenv:


ihnorton (approved):
We might want this to be shown as a `timedelta`? But good to get this merged for release, if that's not easy to do quickly.

PR #999, opened 2022-03-26
Allow Setting `ascii` in `column_type` For `from_pandas`/`from_csv`

nguyenv:


tiledb/dataframe_.py
```diff
@@ -174,7 +177,10 @@ def _get_column_infos(df, column_types, varlen_types):
     column_infos = {}
     for name, column in df.items():
         if column_types and name in column_types:
-            column_infos[name] = ColumnInfo.from_dtype(column_types[name], varlen_types)
+            column_infos[name] = ColumnInfo.from_dtype(
+                "ascii" if column_types[name] == "ascii" else column_types[name],
```
ihnorton, line 181: This should be in the `from_dtype` implementation IMO.

ihnorton (approved):

PR #1056, opened 2022-04-29
Arrow Buffer Ownership Handoff

ihnorton:
* This fixes an issue where `.query(return_arrow=True)` did not
  properly return an Arrow Table due to `~PyQuery` destroying the
  NumPy arrays holding the Arrow buffers upon returning from
  `multirange_indexing.DataFrameIndexer.__getitem__`. The buffers
  are now linked to the lifetime of the PyArrow Table rather
  than the `PyQuery` object
* Move `arrow_io_impl.h` from core to `py_arrow_io_impl.h`
* Addition of `BufferHolder` struct to hold `py::object`s `data`,
  `offset`, and `validity` which respectively hold the necessary
  buffers for the Arrow Table
* In `PyQuery.pa_buffer`, create `BufferHolder` and pass buffers
  to increment refcount. Set `CPPArray.cbb_data` to `BufferHolder`
  pointer and `CPPSchema.release_cb` to `free_buffer_holder`.
  When the Arrow Table destructor is called, `free_buffer_holder`
  calls `delete` on the `BufferHolder` object which in turns
  decrements the refcount for the buffers

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -12,3 +12,5 @@ setuptools_scm >= 1.5.4
 wheel >= 0.30
 contextvars ;python_version<"3.7"
 dataclasses ;python_version<"3.7"
+black
+clang-format == 10.0.1
--- a/tiledb/core.cc
+++ b/tiledb/core.cc
@@ -22,8 +22,11 @@
 #include <tiledb/tiledb>                // C++
 #include <tiledb/tiledb_experimental.h> // C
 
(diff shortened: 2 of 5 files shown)


nguyenv:
[sc-16296]

nguyenv:
Confirmed that this fixes the original reported bug (was segfaulting before).

original-pyarrow-bug

nguyenv (approved):

PR #1059, opened 2022-04-29
Add Nightly Workflow To Test Against Earliest Supported NumPy Version

nguyenv:


nguyenv:
Small note that I did try 

1. Doing this in a single workflow
2. Making a reusable workflow

But it ended up being a lot easier to have these two workflow files even though I know there is quite a bit of code duplication.

ihnorton (approved):
I think we should either run this as nightly or prioritize adding support for using the prebuild libtiledb artifacts, because otherwise this is going to add a lot of time to CI.

nguyenv:
> I think we should either run this as nightly or prioritize adding support for using the prebuild libtiledb artifacts, because otherwise this is going to add a lot of time to CI.

OK I can switch this to running nightly instead.

nguyenv:
Proof of passing: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/2245912954

PR #1071, opened 2022-05-04
Use `sm.compute_concurrency_level` For `libtiledb`>=2.10

nguyenv:
Fixes deprecation error

`[2022-05-04 11:44:18.984] [Process: 5350] [error] [Context: 20] Config parameter "sm.num_reader_threads" has been removed; use config parameter "sm.compute_concurrency_level".`


ihnorton (approved):
Thanks!

ihnorton (commented):
Oh, actually - this needs to be conditional because the change is only on dev and will not be backported to 2.9

PR #1089, opened 2022-05-11
[Merge 2.11] Addition of `test_from_numpy_empty_str`

nguyenv:
* Remove `\x00` for `test_varlen_sparse_all_empty_strings`

### DIFF
--- a/tiledb/tests/test_libtiledb.py
+++ b/tiledb/tests/test_libtiledb.py
@@ -1849,13 +1849,7 @@ class TestVarlen(DiskTestCase):
             assert_array_equal(A, T.multi_index[1 : len(A)][""])
 
     def test_varlen_sparse_all_empty_strings(self):
-        # this test addresses a fix for specific need for reads on a
-        # large existing array, see
-        #   https://github.com/TileDB-Inc/TileDB-Py/pull/475
-        # we currently have to write a placeholder at the end to
-        # avoid zero-length cell error
-        # TODO: follow-up with improved testing for empty var-length/strings
-        A = np.array(["", "", "", "", "", "\x00"], dtype=object)
+        A = np.array(["", "", "", "", ""], dtype=object)
         dim_len = len(A)
         uri = self.path("varlen_all_empty_strings")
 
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
🚀

HISTORY.md
```diff
@@ -3,6 +3,9 @@
 ## Misc Updates
 * Wheels will no longer be supported for macOS 10.14 Mojave; the minimum supported macOS version is now 10.15 Catalina [#1080](https://github.com/TileDB-Inc/TileDB-Py/pull/1080)
 
+# Bug Fixes
+* Correct issue where writing a NumPy array containing emppy strings would error out [#1089](https://github.com/TileDB-Inc/TileDB-Py/pull/1089)
```
ihnorton, line 7: `emppy` -> `empTy`

tiledb/tests/test_libtiledb.py
```diff
@@ -3477,6 +3477,16 @@ def test_from_numpy_timestamp(self):
...
 
+    @pytest.mark.parametrize("empty_str", ["", b""])
+    @pytest.mark.parametrize("num_strs", [1, 1000])
+    def test_from_numpy_empty_str(self, empty_str, num_strs):
+        uri = self.path("test_from_numpy_empty_str")
+        np_array = np.asarray([empty_str] * num_strs, dtype="O")
+        tiledb.from_numpy(uri, np_array)
+
+        with tiledb.open(uri, "r") as A:
+            assert_array_equal(A[:], np_array)
```
ihnorton, line 3488: Should test the `.df` reader here as well if pandas is available.

tiledb/tests/test_libtiledb.py
```diff
@@ -1805,7 +1805,7 @@ def test_varlen_sparse_all_empty_strings(self):
         # we currently have to write a placeholder at the end to
         # avoid zero-length cell error
         # TODO: follow-up with improved testing for empty var-length/strings
-        A = np.array(["", "", "", "", "", "\x00"], dtype=object)
```
ihnorton, line 1808: We can probably just remove this test in favor of the new one below, but please at least remove the comments here since they are now out of data (this was fixed by Luc a few months ago I think).

ihnorton (approved):
Thanks, just minor nits.

tiledb/tests/test_libtiledb.py
```diff
@@ -3477,6 +3477,16 @@ def test_from_numpy_timestamp(self):
         with tiledb.open(path, timestamp=(10, 10)) as A:
             assert A.nonempty_domain() == ((0, 2),)
 
+    @pytest.mark.parametrize("empty_str", ["", b""])
+    @pytest.mark.parametrize("num_strs", [1, 1000])
```
ihnorton, line 3481: Can this start from zero?

tiledb/tests/test_libtiledb.py
nguyenv: Using Arrow and then converting to df is giving the NULL character instead of an empty string. .df works fine when not using Arrow.

tiledb/tests/test_libtiledb.py
nguyenv: It is giving this error `TypeError: invalid domain extent, domain cannot be safely cast to dtype dtype('uint64')`. This seems like a different issue, so I can make a new Shortcut story for it.

nguyenv:
This bug has been fixed in core, so a fix in TileDB-Py is no longer needed. This PR now just contains modified and additional unit tests for empty strings.

johnkerl (approved):
🚢

PR #1096, opened 2022-05-13
Fix race condition in n {DenseArray,SparseArray}.__new__

gsakkis:
The actual fix is on https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/278, this PR just enables it. 

The current approach based on injecting `CloudArray` to `__bases__` is preserved (and slightly improved) for backwards compatibility with previous tiledb-cloud versions. At some point in the future it should better be removed.


thetorpedodog:
Would it be better to just include `TypeError` in your `except` block? That seems like it would guarantee the exactly-once behavior:

```python
try:
  from tiledb.cloud import cloudarray
  cls.__bases__ += (cloudarray.CloudArray,)
except (ImportError, TypeError):
  # If we couldn't import, tiledb.cloud isn't installed.
  # If there was a TypeError, the base class was already added.
  pass
```

gsakkis:
> Would it be better to just include TypeError in your except block? That seems like it would guarantee the exactly-once behavior:

Great idea @thetorpedodog, this removes the need for a monkeypatch method in tiledb-cloud. PR updated.

tiledb/array.py
```diff
@@ -19,20 +18,21 @@ class DenseArray(DenseArrayImpl):
...
-                from tiledb.cloud import cloudarray
-
-                DenseArray.__bases__ = DenseArray.__bases__ + (cloudarray.CloudArray,)
-            except ImportError:
-                pass
-
+        try:
+            from tiledb.cloud import cloudarray
+
+            cls.__bases__ += (cloudarray.CloudArray,)
```
thetorpedodog, line 25: Looking at this again, I was slightly wrong at first: we do want to do `DenseArray.__bases__` here instead of `cls.__bases__`. (For more details see the log in the other comment.)

tiledb/array.py
```diff
@@ -19,20 +18,21 @@ class DenseArray(DenseArrayImpl):
...
+        except (ImportError, TypeError):
+            # If we couldn't import, tiledb.cloud isn't installed
+            # If there was a TypeError, CloudArray was already added (by another thread)
+            pass
+        try:
+            # Mixin initialization completed: delete this method for subsequent calls
+            del DenseArray.__new__
```
thetorpedodog, line 32: I am admittedly a little wary of modifying object pieces like this in the middle of a call (e.g. is `super()` guaranteed to handle the disappearance of `__new__` properly if it’s deleted as it runs?) but I can confirm that, in the absence of thread concerns, this works:

```
In [1]: class A:
   ...:     def __new__(cls, *args, **kwargs):
   ...:         print(f"A.__new__({cls}, ...)")
   ...:         ret = super().__new__(cls, *args, **kwargs)
   ...:         print("A.__new__ done")
   ...: 

In [2]: class B(A):
   ...:     def __new__(cls, *args, **kwargs):
   ...:         print(f"B.__new__({cls}, ...")
   ...:         ret = super().__new__(cls, *args, **kwargs)
   ...:         print("B.__new__ done")
   ...: 

In [3]: a = A()
A.__new__(<class '__main__.A'>, ...)
A.__new__ done

In [4]: b = B()
B.__new__(<class '__main__.B'>, ...
A.__new__(<class '__main__.B'>, ...)
A.__new__ done
B.__new__ done

In [5]: del B.__new__

In [6]: b2 = B()
A.__new__(<class '__main__.B'>, ...)
A.__new__ done
```

thetorpedodog (commented):
I’ll leave it to Vivian or Isaac to give the final OK on this, but assuming that `del SomeCls.__new__` is safe, all that needs to be changed here is the two `cls.__bases__` locations.

tiledb/array.py
thetorpedodog: ha ha ha i forgot to `return ret` good job me

tiledb/array.py
gsakkis: You are technically correct, [the best kind of correct](https://www.youtube.com/watch?v=0ZEuWJ4muYc); in practice though I believe `DenseArray/SparseArray` are effectively "final", i.e. they're not meant to be subclassed so `cls is DenseArray`.

tiledb/array.py
gsakkis: As commented above, in general this would need more testing in the presence of (single, let alone multiple) inheritance. In this case we can even replace `super()` with `DenseArrayImpl/SparseArrayImpl` and not worry about method resolution.

nguyenv (approved):
Just tested with the example in the Shortcut story and looks OK

```
import sys
import time
import tiledb
import concurrent.futures


def worker(uri, sleep=1):
    with tiledb.DenseArray(uri) as A:
        pass
    time.sleep(sleep)
    return hasattr(A, "apply")


def main(uri):
    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
        futures = [executor.submit(worker, uri) for _ in range(0, 5)]
        res = [f.result() for f in concurrent.futures.as_completed(futures)]
    assert all(res), res


if __name__ == "__main__":
    main(sys.argv[1])
```

```
(tiledb-3.10) vivian@mangonada:~/tiledb-bugs/17758$ python test.py tiledb://TileDB-Inc/quickstart_dense
(tiledb-3.10) vivian@mangonada:~/tiledb-bugs/17758$
```

PR #1123, opened 2022-05-24
Addition of `keys`, `values`, and `items` to `Group.meta`

nguyenv:


gsakkis:
Btw you may extend [MutableMapping](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) and get all these methods (and more) for free.

tiledb/cc/group.cc
```diff
@@ -14,45 +14,76 @@ using namespace tiledb;
...
+  if (tdb_type == TILEDB_STRING_UTF8) {
+    value_type = py::dtype("|S1");
+  }
 
-  py::array py_buf(value_type, value_num);
-  memcpy(py_buf.mutable_data(), c_buf, py_buf.nbytes());
+  py_buf = py::array(value_type, value_num);
 
-  return py_buf;
+  memcpy(py_buf.mutable_data(), value, py_buf.nbytes());
```
ihnorton, line 84: There is a `py::array` signature which copies from a pointer (by default, unless you pass a base object): https://github.com/pybind/pybind11/blob/68f80105007a574f114cfe899243bb3d96272a2e/include/pybind11/numpy.h#L745-L747

(usage/test: https://github.com/pybind/pybind11/blob/master/tests/test_numpy_array.cpp#L224-L230)

johnkerl:
🎉

PR #1141, opened 2022-06-01
`Group.member` Also Returns `name` If Given

nguyenv:
* If the name is not given, returns None


johnkerl (approved):
Looks great, thanks!! :D

PR #1146, opened 2022-06-02
Support OR Query Condition

nguyenv:
- [x] Add `TILEDB_OR` enum
- [x] Add unit tests for OR
- [x] Update grammar in documentation


ihnorton (approved):
Tested locally, LGTM. Let's get the first part of this merged so we can run this in nightly, and do the refactor as follow-up.

PR #1154, opened 2022-06-06
Support `bool` in `dataframe_.py`

nguyenv:
* For libtiledb 3.10+, use `bool` instead of `uint8` for Boolean dtype

### DIFF
--- a/tiledb/tests/test_pandas_dataframe.py
+++ b/tiledb/tests/test_pandas_dataframe.py
@@ -1395,7 +1395,8 @@ class TestPandasDataFrameRoundtrip(DiskTestCase):
         tiledb.from_pandas(uri, df)
 
         with tiledb.open(uri) as A:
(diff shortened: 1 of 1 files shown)


johnkerl (commented):
LGTM! :)

tiledb/dataframe_.py
```diff
@@ -143,7 +144,9 @@ def from_dtype(cls, dtype, varlen_types=()):
 
         # bool type
         if pd_types.is_bool_dtype(dtype):
-            return cls(np.dtype("uint8"), repr="bool")
+            # as of libtiledb 2.10, we support TILEDB_BOOL
+            dtype = "uint8" if tiledb.libtiledb.version() < (2, 10) else "bool"
+            return cls(np.dtype(dtype), repr="bool")
```
gsakkis, line 149: Nitpick: the `repr` is not needed if dtype=="bool"

PR #1070, opened 2022-05-04
Addition of Filestore API

nguyenv:

PR #1159, opened 2022-06-08
Add Support for Reading `TILEDB_BLOB`

nguyenv:


nguyenv:
I have not added tests to this PR as I have not added support for writing `TILEDB_BLOB` here. But #1070 has unit tests.

With the original bug report:
```
import tiledb

def config():
    cfg = tiledb.Config()
    # use your s3 sandbox keys
    cfg["vfs.s3.aws_access_key_id"] = ...
    cfg["vfs.s3.aws_secret_access_key"] = ...
    cfg["vfs.s3.region"] = "us-east-2"
    return cfg

uri = "s3://tiledb-anastaspy/staging/Test2.9.1/plan9"
arr = tiledb.DenseArray(uri, mode="r", ctx=tiledb.Ctx(config()))
print(arr.meta.items())
arr.schema.dump()
print(arr.schema)
contents = arr[slice(0, arr.meta["file_size"])]
print(contents)
```

No longer errors out:
```
(tiledb-3.10) vivian@mangonada:~/TileDB-Py$ python ~/tiledb-bugs/tiledb_blob_support.py
(('file_extension', 'pdf'), ('file_size', 130218), ('mime_type', 'application/pdf'), ('original_file_name', '1654516429846978711.pdf'))
- Array type: dense
- Cell order: row-major
- Tile order: row-major
- Capacity: 10000
- Allows duplicates: false
- Coordinates filters: 1
  > ZSTD: COMPRESSION_LEVEL=-1
- Offsets filters: 1
  > ZSTD: COMPRESSION_LEVEL=-1
- Validity filters: 1
  > RLE: COMPRESSION_LEVEL=-1

### Dimension ###
- Name: position
- Type: UINT64
- Cell val num: 1
- Domain: [0,18446744073709550591]
- Tile extent: 1024
- Filters: 0

### Attribute ###
- Name: contents
- Type: BLOB
- Nullable: false
- Cell val num: 1
- Filters: 0
- Fill value:


ArraySchema(
  domain=Domain(*[
    Dim(name='position', domain=(0, 18446744073709550591), tile=1024, dtype='uint64'),
  ]),
  attrs=[
    Attr(name='contents', dtype='int8', var=False, nullable=False),
  ],
  cell_order='row-major',
  tile_order='row-major',
  capacity=10000,
  sparse=False,
)

OrderedDict([('contents', array([37, 80, 68, ..., 79, 70, 10], dtype=int8))])
```

anastasop:
I tested locally a build of this branch with TileDB 2.9.3 and a rest server build with BLOB support. It works OK

ihnorton (approved):
LGTM.

(at some point we should provide a py::buffer interface to BLOB data, but I'm not quite sure how that should interact with queries..)

PR #1165, opened 2022-06-10
Fix For Segfault When Passing `Context` to `Group`

nguyenv:
* This is a temporary fix that makes a copy of the `Context` so that
  the new `Context` is tied to the lifetime of the `Group`. Because
  `tiledb.Ctx` is still in Cython, the `tiledb_ctx_t*` pointer may
  be freed prior to `~Group` despite the usage of `py::keep_alive<1, 2>`
* This should be reverted once `tiledb.Ctx` has been completely moved
  out of Cython


johnkerl (approved):
🚢

ihnorton (changes requested):
Contexts have a lot of state attached (eg thread pools) which should ideally not be duplicated like this. I believe we can fix this by changing the Context pybind11  signature from:
```
.def(py::init<py::capsule, bool>())
```
to
```
.def(py::init<py::object, bool>(), py::keepalive<1>())
```
and then getting/unwrapping the capsule from the object inside the pybind11 def before constructing a non-owning C++ Context object. This will keep the (Cython) Ctx object alive due to a refcount increment, but will not try to free it from the pybind11 side when the pybind11 wrapper is finalized.

ihnorton (approved):
Perfect, LGTM. Thanks!

PR #1167, opened 2022-06-10
Fix Arrow Table Incorrect Boolean Results

nguyenv:
* `TILEDB_BOOL` is represented as a `uint8_t` whereas Arrow's Boolean
  type is 1 bit
* Use Arrow's uint8 type in `tiledb_buffer_arrow_fmt` for `TILEDB_BOOL`
* Cast Boolean types in the resulting Arrow table in `_run_query` from
  uint8 to bool


nguyenv:
Testing wheels against 2.10.0-rc0 with this fix.

https://github.com/TileDB-Inc/TileDB-Py/tree/release-against-2.10.0-rc0

ihnorton (approved):

PR #1182, opened 2022-06-16
Index out of bounds when nonempty domain is empty string

robertbindar:
The nonempty domain for the `clinvar-annotations` array turns out to be the empty string for the third dimension.
This was discovered whilst debugging ch18654 when queries to this array resulted in `Index out of bounds` exceptions.
This is unrelated though unfortunately with the original problem described in ch1865.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -3828,7 +3828,7 @@ cdef class Array(object):
                     continue
 
                 buf_dtype = 'S'
-                start_buf = np.empty(end_size, 'S' + str(start_size))
+                start_buf = np.empty(start_size, 'S' + str(start_size))
                 end_buf = np.empty(end_size, 'S' + str(end_size))
                 start_buf_ptr = np.PyArray_DATA(start_buf)
                 end_buf_ptr = np.PyArray_DATA(end_buf)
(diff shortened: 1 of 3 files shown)


tiledb/libtiledb.pyx
```diff
@@ -3825,7 +3825,11 @@ cdef class Array(object):
                     _raise_ctx_err(ctx_ptr, rc)
                 if is_empty:
                     return None
-                results.append((start_buf.item(0), end_buf.item(0)))
+
+                if start_size > 0 and end_size > 0:
+                    results.append((start_buf.item(0), end_buf.item(0)))
+                else:
+                    results.append(('', ''))
```
ihnorton, line 3832: I think this should be `(None,None)` to match other coe paths. cc @nguyenv for decision though.

nguyenv: Yes I agree to make it `(None, None)`

nguyenv:
I might be misunderstanding what the original issue is. I wrote this unit test and ran it on `dev` but am not getting an out of bounds error.

```
def test_nonempty_domain_empty_string(self):
        uri = self.path("test_nonempty_domain_empty_string")
        dims = [tiledb.Dim(dtype="ascii")]
        attrs = [tiledb.Attr(dtype=int)]
        schema = tiledb.ArraySchema(tiledb.Domain(dims), attrs=attrs, sparse=True)

        with tiledb.open(uri, "w") as A:
            A[""] = 0

        with tiledb.open(uri, "r") as A:
            assert_array_equal(A[:][""], 0)
            assert_array_equal(A.nonempty_domain(), ((None, None),))
```

PR #1190, opened 2022-06-23
Increase Azure Devops Timeout To 120 Minutes

nguyenv:
* The macOS build frequently takes more than 60 minutes

### DIFF
--- a/misc/azure-ci.yml
+++ b/misc/azure-ci.yml
@@ -18,6 +18,7 @@ stages:
           imageName: 'ubuntu-latest'
           python.version: '3.10'
       maxParallel: 4
+    timeoutInMinutes: 120
 
     steps:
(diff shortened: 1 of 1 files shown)


ihnorton (approved):
I'm a bit puzzled why this would be taking so long now...

nguyenv:
I was wondering about that too. It has been like this for nearly a month now. When I restart the job, it is able to build and pass.

nguyenv:
Core does nightly builds now too, right? Are we just exhausting all the available macOS runners and needing to wait until one is available?

ihnorton:
> Are we just exhausting all the available macOS runners and needing to wait until one is available?

I thought this was the timeout of the job-run itself, once it actually has a runner and starts?

Also core nightlies are on github actions.

PR #1191, opened 2022-06-23
Correct `tiledb_version.h`  Search in `setup.py`

nguyenv:


nguyenv:
BTW this modification only fixes the part where 'prefix_dir' had to be replaced with '"/usr"'

eddelbuettel (approved):
Equivalent to the version we tested last week and which worked so :+1:

ihnorton (approved):
Approved pending validation of the permissions as-installed from a wheel after this change.

nguyenv:
[sc-18730]

PR #1199, opened 2022-06-30
Dense Array `[:]` Returns Data From Nonempty Domain Only

nguyenv:
* Querying dense array with `[:]` returns shape that matches nonempty
  domain, consistent with `.df[:]` and .multi_index[:]`
* Previously, the entire domain was returned where values outside of the
  nonempty domain contained default fill values

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -2525,7 +2525,7 @@ def replace_scalars_slice(dom: Domain, idx: tuple):
     return tuple(new_idx), tuple(drop_axes)
 
 
-def index_domain_subarray(array: Array, dom: Domain, idx: tuple):
+def index_domain_subarray(array: Array, dom: Domain, idx: tuple, read_array: bool = False):
     """
     Return a numpy array representation of the tiledb subarray buffer
     for a given domain and tuple of index slices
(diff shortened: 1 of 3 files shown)


tiledb/libtiledb.pyx
```diff
@@ -2543,9 +2543,18 @@ def index_domain_subarray(array: Array, dom: Domain, idx: tuple):
 
         if np.issubdtype(dim_dtype, np.unicode_) or np.issubdtype(dim_dtype, np.bytes_):
             ned = array.nonempty_domain()
-            (dim_lb, dim_ub) = ned[r] if ned else (None, None)
+            (dim_lb, dim_ub) = ned[r] if ned is not None else (None, None)
         else:
-            (dim_lb, dim_ub) = dim.domain
+            if read_array:
+                with Array.load_typed(array.uri) as A:
```
ihnorton, line 2549: I'm not quite following why this is done here, but we need to avoid re-opening the array inline if at all possible... There are many events that could violate expectations here (eg if the array has been updated since the parent-array caller's timestamp, which might change the fragments, nonempty domain, etc.).

nguyenv: Is there a way to get the non-empty domain without opening the array? When I tried doing `array.nonempty_domain()` it complained "Array is not open for reading."

ihnorton (approved):

PR #1226, opened 2022-07-15
Remove Unused/Extraneous `Filestore.uri_import`

nguyenv:
* This function was accidentally left in; it does not work
* The function to write a file located at URI to a filestore array is
  already correctly implemented by `Filestore.copy_from`

### DIFF
--- a/tiledb/filestore.py
+++ b/tiledb/filestore.py
@@ -1,4 +1,5 @@
-from typing import ByteString, Optional, overload, TYPE_CHECKING
+from typing import ByteString, TYPE_CHECKING
+import warnings
 
 import tiledb.cc as lt
 from .ctx import default_ctx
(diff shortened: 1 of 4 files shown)


nguyenv:
Failing windows tests due to #1218

tiledb/filestore.py
```diff
@@ -254,33 +254,6 @@ def copy_to(filestore_array_uri: str, file_uri: str, ctx: "Ctx" = None) -> None:
 
         lt.Filestore._uri_export(lt.Context(ctx, False), filestore_array_uri, file_uri)
 
-    def uri_import(self, uri: str, mime_type: str = "AUTODETECT") -> None:
```
ihnorton, line 257: Please keep the function, mark it deprecated, and throw an error with a message suggesting what to use instead.

ihnorton (approved):
Small change request, but approving so you can merge after update.

PR #1236, opened 2022-07-20
Update dataframe_.py

royassis:
if sparse == False and (not index_dims or "index_col" not in kwargs):
NameError: name 'kwargs' is not defined

kwargs was replaced with tiledb_args

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -479,7 +479,7 @@ def _from_pandas(uri, dataframe, tiledb_args):
 
     # TODO: disentangle the full_domain logic
     full_domain = tiledb_args.get("full_domain", False)
-    if sparse == False and (not index_dims or "index_col" not in kwargs):
+    if sparse == False and (not index_dims or "index_col" not in tiledb_args):
         full_domain = True
     if full_domain is None and tiledb_args.get("nrows"):
         full_domain = False


royassis:
now I get another error

tiledb.cc.TileDBError: [TileDB::Dimension] Error: Domain check failed; Domain range (upper + lower + 1) is larger than the maximum unsigned number

nguyenv:
Hi @royassis.

Thanks for submitting this PR. Can you provide us a reproducible example that hits this error?

Thanks.

royassis:
I might tried to use sparse data to create this, but i'm not sure. 

Pseudocode
```python
df.shape  == (~50000**2, 3)
df.columns == ["gene_a_id", "gene_b_id", "val"]
df.dtypes == (np.int32,np.int32,np.float32)

tiledb.from_pandas(uri, df, index_dims=["gene_a_id", "gene_b_id"], sparse=False)
```

The **_from_pandas** function in **dataframe_.py**  line **482** 
```python
 if sparse == False and (not index_dims or "index_col" not in **kwargs**):
```

**kwargs** is referenced but no initiated in the function

nguyenv:
Thanks - I am able to reproduce your original error now with:

```
import numpy as np
import pandas as pd
import tiledb

uri = "./test_pandas"

# create the df
df = pd.DataFrame(
    {
        "gene_a_id": np.random.randint(0, 2, size=(10,)),
        "gene_b_id": np.random.randint(0, 2, size=(10,)),
        "val": np.random.randint(0, 2, size=(10,)),
    }
)

tiledb.from_pandas(uri, df, index_dims=["gene_a_id", "gene_b_id"], sparse=False)

with tiledb.open(uri, "r") as a:
    print(a.df[:])
```

I'll keep investigating.

stavrospapadopoulos:
@royassis `tiledb.from_pandas(uri, df, index_dims=["gene_a_id", "gene_b_id"], sparse=False)` is not meaningful, as it is trying to create a multi-dimensional array for a dataframe. That is not permissible. A **dense** dataframe is always **1D**, with the indexed dimension being (implicitly) the row id. A **sparse** dataframe instead can be any-D. with the indexed dimensions being any subset of the columns of the dataframe. 

Please check out [this tutorial](https://cloud.tiledb.com/notebooks/details/TileDB-Inc/8a88286f-d66c-470d-ad65-bac29b10509f/preview) for the different use of a dense vs. a sparse dataframe. You can replace `from_csv` in the tutorial with `from_pandas` and everything should work. Please let me know if you have any questions.

stavrospapadopoulos:
@nguyenv, we need a better error message here though.

nguyenv:
@stavrospapadopoulos  OK I will figure out a better error message.

nguyenv:
However, this seems be part of a larger issue with using the full domain. I get the error message `tiledb.cc.TileDBError: [TileDB::Dimension] Error: Domain check failed; Domain range (upper + lower + 1) is larger than the maximum unsigned number` if I use
```
tiledb.from_pandas(
    uri,
    df,
    index_dims=["gene_a_id", "gene_b_id"],
    sparse=True,
    full_domain=True,
)
```
It works fine if `full_domain=False`.

stavrospapadopoulos:
Please see end of [this section](https://docs.tiledb.com/main/how-to/dataframes/write-dataframes/write-dataframe-from-csv#create-and-write-to-a-sparse-dataframe). If you are appending to a dataframe, you need to set `full_domain=True`. Also see rest of the docs in case you are appending to a dense array, you need to set `row_start_idx`.

royassis:
> @royassis `tiledb.from_pandas(uri, df, index_dims=["gene_a_id", "gene_b_id"], sparse=False)` is not meaningful, as it is trying to create a multi-dimensional array for a dataframe. That is not permissible. A **dense** dataframe is always **1D**, with the indexed dimension being (implicitly) the row id. A **sparse** dataframe instead can be any-D. with the indexed dimensions being any subset of the columns of the dataframe.
> 
> Please check out [this tutorial](https://cloud.tiledb.com/notebooks/details/TileDB-Inc/8a88286f-d66c-470d-ad65-bac29b10509f/preview) for the different use of a dense vs. a sparse dataframe. You can replace `from_csv` in the tutorial with `from_pandas` and everything should work. Please let me know if you have any questions.

Hey @stavrospapadopoulos, I already created a dense array from a dataframe. It seemed to work ok. 
This dataframe has 3 columns, 2 out of 3 are dims. Please correct me if there is a problem.

 I would also suggest adding a link to the csv files in the tutorial.

stavrospapadopoulos:
For dense arrays, there is no column dimension - there is only a single dimension, which is implicit and allows you to slice rows. So probably you created a sparse dataframe. Please post the schema. 

```python
A = tiledb.open(<your_array_name>, "r")
print(A.schema)
```

For either the dense or sparse case, when you create the dataframe in TileDB, you need to specify `full_domain=True`. @nguyenv will investigate if a bug is lurking there. After that, you'll be able to write more pandas dataframes or CSV files using `mode=append`, noting that you will need to set `row_start_idx` for the dense case (the tutorial and the How To docs hopefully cover these topics - please let me know if there is anything missing, other than the `from_pandas` documentation that will add shortly). 

> I would also suggest adding a link to the csv files in the tutorial.

You just need to run the following as shown in the tutorial:
```python
dataset = os.path.expanduser("~/yellow_tripdata_2020-01.csv")
urllib.request.urlretrieve("https://tiledb-inc-demo-data-us-east-1.s3.amazonaws.com/tutorials/dataframes/yellow_tripdata_2020-01.csv", dataset)
```

PR #1239, opened 2022-07-21
Correct `full_domain=True` For `from_pandas`

nguyenv:
* Previously, the `tile_max` was only correctly calculated for
  `np.int64`. This correctly set the `dim_max` to `dtype_max - tile`
  but only for `np.int64`
* All other integer dtypes had `dim_max` set to `dtype_max` which
  resulted in a domain range that was too large
* We also need to account for when the tile extent is larger than
  the range of the full domain. For an instance, when we have a dim
  dtype of `np.int8`, the default tile extent of 10000 will be much
  larger than the range of the full domain


nguyenv:
I am running into one issue though. When I try to create the full domain for `np.int8`:
```
>>> tiledb.Dim(domain=(-128, 126), tile=254, dtype=np.int8)
```

I get this error:
```
  File "tiledb/libtiledb.pyx", line 1982, in tiledb.libtiledb.Dim.__init__
  File "tiledb/libtiledb.pyx", line 1963, in tiledb.libtiledb.Dim.__init__
  File "tiledb/libtiledb.pyx", line 579, in tiledb.libtiledb.check_error
  File "tiledb/libtiledb.pyx", line 575, in tiledb.libtiledb._raise_ctx_err
  File "tiledb/libtiledb.pyx", line 560, in tiledb.libtiledb._raise_tiledb_error
tiledb.cc.TileDBError: [TileDB::Dimension] Error: Tile extent check failed; Tile extent exceeds dimension domain range
```

But I don't think I should be?

ihnorton:
I'll have to look closer later, it fails even with `tile=120`. I think the error message for what's happening here has regressed.

nguyenv:
[sc-8097]
[sc-20752]

Taking a look at this branch again due to a recent related bug.

PR #1240, opened 2022-07-21
Go Back To Parsing `libtiledb` Version From `tiledb_version.h`

nguyenv:
The conda packages for aarch64 and arm64 do not build correctly when using ctypes to call `tiledb_get_version` from the TileDB shared object library due to a permissions issue. We are reverting back to parsing from `tiledb_version.h`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,24 @@ def _libtiledb_exists(library_dirs):
     try:
         # note: this is a relative path on linux
         #       https://bugs.python.org/issue21042
-        CDLL(lib_name)
+
+        lib = CDLL(lib_name)
+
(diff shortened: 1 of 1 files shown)


nguyenv:
I tested this on the EC2 aarch64 instance by (1) running `python setup.py` and (2) running `conda build . -c conda-forge` in `tiledb-py-feedstock/recipe`.

nguyenv:
I checked that this works for building libtiledb from scratch, using `--tiledb=...`, and global installs with `/usr/local`. @eddelbuettel has confirmed that these changes work on their end too.

ihnorton (approved):

PR #1275, opened 2022-08-10
[2.12] Migrate CI To Run On macOS 11.0

nguyenv:
- Microsoft is currently issuing brownouts for macOS 10.15. Support in GitHub Actions will cease at the end of August and in Azure at the end of September. Our minimum supported version will need to be bumped to macOS 11 Big Sur.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,11 @@ jobs:
         shell: bash
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-10.15, windows-latest]
+        os: [ubuntu-latest, macos-11, windows-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
       fail-fast: false
     env:
-      MACOSX_DEPLOYMENT_TARGET: 10.15
+      MACOSX_DEPLOYMENT_TARGET: "10.15"
     steps:
       - name: Checkout TileDB-Py `dev`
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


nguyenv:
[sc-19634]

nguyenv:
Will need to look at issues with building wheels with macOS 11.0 upgrade:
> `ERROR: tiledb-0.1.dev1-cpxx-cpxx-macosx_11_0_x86_64.whl is not a supported wheel on this platform`

nguyenv:
Microsoft has just announced that deprecation of 10.15 is being pushed to December 1st: https://github.com/actions/runner-images/issues/5583#issuecomment-1219501124. We should still make sure that this gets updated sooner rather than later.

nguyenv:
We now build wheels on macOS 11 as of https://github.com/TileDB-Inc/TileDB-Py/pull/1300. However, we still need to run CI on macOS 11. 

The culprit lines are:
https://github.com/TileDB-Inc/TileDB-Py/blob/dev/.github/workflows/ci.yml#L58-L65

nguyenv:
https://github.com/TileDB-Inc/TileDB-Py/actions/runs/3054432721/jobs/4926319975#step:6:48

https://github.com/actions/runner-images/issues/2187

https://github.com/actions/setup-python/issues/279
https://github.com/pypa/pip/issues/10670

Wheels built on Microsoft's macOS-11 runners are not actually compatible with itself seemingly due to Intel vs M1 usage and the fact Microsoft does not yet support M1 (not sure I completely understand)? I think this means we still need to upgrade the runners to macOS-11 but keep `MACOSX_DEPLOYMENT_TARGET` at 10.15.

ihnorton (approved):

PR #1286, opened 2022-08-17
Use Arrow Type Fixed-Width Binary ("w:") For Fixed-Width `TILEDB_CHAR`

nguyenv:
* Previously, this errored out with `ArrowInvalid: Expected 3 buffers for imported type large_binary, ArrowArray struct has 2`
* Fixed-width `TILEDB_CHAR` was using type large binary ("Z") which requires 3 buffers, including an offset buffer, that is not present or necessary for fixed-width binary strings
* This does not handle fixed-width `TILEDB_UTF8`

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 * Fix issue where querying an array with a Boolean type when `arrow=True`, but is unselected in `.query(attr=...)`, results in an error `pyarrow.lib.ArrowInvalid: Invalid column index to set field.` []()
+* Use Arrow type fixed-width binary ("w:") for non-variable TILEDB_CHAR []()
 
 # TileDB-Py 0.17.1 Release Notes
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,7 @@
 
 ## Bug Fixes
 * Fix issue where querying an array with a Boolean type when `arrow=True`, but is unselected in `.query(attr=...)`, results in an error `pyarrow.lib.ArrowInvalid: Invalid column index to set field.` []()
-* Use Arrow type fixed-width binary ("w:") for non-variable TILEDB_CHAR []()
+* Use Arrow type fixed-width binary ("w:") for non-variable TILEDB_CHAR [#1286](https://github.com/TileDB-Inc/TileDB-Py/pull/1286)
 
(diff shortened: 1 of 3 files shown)


Shelnutt2:
> Fixed-width TILEDB_CHAR was using type large binary ("Z") which requires 3 buffers, including an offset buffer, that is not present or necessary for fixed-width binary strings

Do we need to check if the attribute is nullable then use `Z` else use `w:`?

nguyenv:
> > Fixed-width TILEDB_CHAR was using type large binary ("Z") which requires 3 buffers, including an offset buffer, that is not present or necessary for fixed-width binary strings
> 
> Do we need to check if the attribute is nullable then use `Z` else use `w:`?

Let me double check. I am not sure how the validity buffer might affect this.

nguyenv:
I slightly modified the unit test I wrote to use `nullable=True`, and it passes as is using "w:".
```
    @pytest.mark.skipif(not has_pandas(), reason="pandas not installed")
    def test_fixed_width_char(self):
        uri = self.path("test_fixed_width_char")
        schema = tiledb.ArraySchema(
            domain=tiledb.Domain(tiledb.Dim(name="dim", domain=(0, 2), dtype=np.uint8)),
            attrs=[tiledb.Attr(dtype="|S3", nullable=True, var=False)],
            sparse=True,
        )

        data = np.array(["cat", "dog", "hog"], dtype="|S3")

        with tiledb.open(uri, mode="w") as A:
            A[[0, 1, 2]] = data

        with tiledb.open(uri, mode="r") as A:
            assert all(A.query(use_arrow=True).df[:][""] == data)
```
```
(tiledb-3.10) vivian@mangonada:~/TileDB-Py/tiledb/tests$ pytest -k test_fixed_width_char
================================================ test session starts ================================================
platform linux -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/vivian/TileDB-Py, configfile: pyproject.toml
plugins: hypothesis-6.46.9
collected 520 items / 519 deselected / 1 skipped / 1 selected

test_multi_index.py .                                                                                         [100%]

=================================== 1 passed, 1 skipped, 519 deselected in 0.49s ====================================
```

Shelnutt2:
Thanks for testing!

ihnorton (approved):

PR #1291, opened 2022-08-19
Fix Bug For Querying With Unselected Boolean Type

nguyenv:
* Fix issue where querying an array with a Boolean type when `arrow=True`, but is unselected in `.query(attr=...)`, results in an error `pyarrow.lib.ArrowInvalid: Invalid column index to set field.`

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+# In Progress
+
+## Bug Fixes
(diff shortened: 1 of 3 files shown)


Shelnutt2 (approved):

johnkerl (approved):
🚀

johnkerl:
thanks @nguyenv !! 😁

PR #1294, opened 2022-08-24
Add support for `XORFilter`

ihnorton:
Adds support for new XOR filter (https://github.com/TileDB-Inc/TileDB/pull/3383).


ihnorton:
(wrote this up for testing the core PR; submitting here so I don't forget about the branch - will not be available until 2.12)

PR #1300, opened 2022-08-26
Deprecate Python 3.6 Wheels

nguyenv:
* Bump to macOS 11 
* Increase timeout for `libtiledb` build to 120 minutes
* Disable `musllinux` wheels builds
* All wheels now build use `USE_CIBW_VERSION: 2.3.0`. Previously, only wheels for Python 3.10 were built with 2.3.0. The earlier Python versions still used `USE_CIBW_VERSION: 1.6.4` because of the need to support Python 3.6 (and only support up to Python 3.9)


misc/azure-release.yml
```diff
@@ -42,162 +44,21 @@ stages:
           - task: UsePythonVersion@0
           - template: azure-libtiledb-darwin.yml
 
-      - job: build2_python310
```
ihnorton, line 45: 😍 this whole deleted section! nice clean-up

misc/azure-release.yml
```diff
@@ -321,6 +182,144 @@ stages:
             displayName: "List all the files"
             condition: always()
 
+      # - job: build2_python
```
ihnorton, line 185: Remove?

ihnorton (approved):
💯

PR #1309, opened 2022-08-31
Pybind11 Support for Query DELETE + Tests

ihnorton:
Example usage:
```
with tiledb.open(path, "d") as A:
    A.query(cond="attr < 4")
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,7 @@
 * Array and Group metadata now store bytes as `TILEDB_BLOB` [#1384](https://github.com/TileDB-Inc/TileDB-Py/pull/1384)
 * Addition of `{Array,Group}.metadata.dump()` [#1384](https://github.com/TileDB-Inc/TileDB-Py/pull/1384)
 * Addition of `Group.is_relative` to check if the URI component of a group member is relative [#1386](https://github.com/TileDB-Inc/TileDB-Py/pull/1386)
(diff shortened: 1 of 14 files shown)


nguyenv:
One thing I am unsure about with the delete API design as I have it is you need to use `__getitem__` in order to actually submit the query:

```
with tiledb.open(path, "d", timestamp=3) as A:
    A.query(attr_cond=tiledb.QueryCondition("ints == 1"))[:]
```

Whereas this won't do anything and just creates a PyQuery object but would make more sense for deletes because we aren't actually getting anything back:
```
with tiledb.open(path, "d", timestamp=3) as A:
    A.query(attr_cond=tiledb.QueryCondition("ints == 1")
```

nguyenv:
> One thing I am unsure about with the delete API design as I have it is you need to use `__getitem__` in order to actually submit the query:
> 
> ```
> with tiledb.open(path, "d", timestamp=3) as A:
>     A.query(attr_cond=tiledb.QueryCondition("ints == 1"))[:]
> ```
> 
> Whereas this won't do anything and just creates a PyQuery object but would make more sense for deletes because we aren't actually getting anything back:
> 
> ```
> with tiledb.open(path, "d", timestamp=3) as A:
>     A.query(attr_cond=tiledb.QueryCondition("ints == 1")
> ```

I've modified this to make it so that the latter does do a delete without needing to apply the regular indexer.

Issue #1315, opened 2022-09-06
empty_like does not accept shape tuple


ocsmit:
Hi, using `tiledb.empty_like` and noticed it does not allow a shape tuple to be passed contrary to documentation. This is due to arr being passed as an arg to schema_like within empty_like without checking if it is a nd.array object or not. 
```
  File "tile.py", line 79, in create
    tiledb.empty_like(str(self.uri), self.shape)
  File "/lib/python3.10/site-packages/tiledb/highlevel.py", line 60, in empty_like
    schema = tiledb.schema_like(arr, tile=tile, ctx=ctx)
  File "/lib/python3.10/site-packages/tiledb/schema.py", line 103, in schema_like
    raise ValueError("expected ndarray-like object")
ValueError: expected ndarray-like object

[Process exited 1]
```

PR #1316, opened 2022-09-06
Add ability to pass shape tuple to empty_like 

ocsmit:
Hopefully addresses #1315

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 # In Progress
 
 ## API Changes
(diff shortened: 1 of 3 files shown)


ihnorton:
Thanks for the PR @ocsmit! I've enabled CI to run through here and @nguyenv will take a look ASAP.

ocsmit:
Thanks!

ocsmit:
No problem, just added! Let me know if the type checking for the shape tuple needs to be more robust as well.

nguyenv:
> No problem, just added! Let me know if the type checking for the shape tuple needs to be more robust as well.

Thanks! And no, I think the checking in `schema_like` should be sufficient.

PR #1319, opened 2022-09-07
Update test_bytes_* again with more information and consolidate

ihnorton:


shaunrd0 (approved):
Cool to see VFS used for creating test arrays. Initially was curious why `from_numpy` didn't need a context like `from_pandas`, but noticed it uses default context [if one is not provided](https://github.com/TileDB-Inc/TileDB-Py/blob/353f2d1c5724874321547e60dc135afe4697486f/tiledb/highlevel.py#L153).

PR #1326, opened 2022-09-17
Move `Attr` To Pure Python (1/3)

nguyenv:
This is a subset of #1188. I'm separating into multiple PRs because it's becoming too big and unwieldy to change and review everything in one go.


tiledb/indexing.pyx
```diff
@@ -21,7 +21,7 @@ def _index_as_tuple(idx):
 cdef class DomainIndexer(object):
 
     @staticmethod
-    def with_schema(ArraySchema schema):
+    def with_schema(schema):
```
ihnorton, line 24: Can you use python type annotation here? (for future transfer to pure python)

tiledb/indexing.pyx
```diff
@@ -31,6 +31,10 @@ cdef class DomainIndexer(object):
         self.array_ref = weakref.ref(array)
         self.schema = array.schema
         self.query = query
+    
+    @property
+    def schema(self):
+        return self.array.array_ref().schema
```
ihnorton, line 37: Does this return a copy? Please document

ihnorton (approved):
💯 nice work!

PR #1327, opened 2022-09-20
Move `Domain` and `Dim` Into Pure Python (2/3)

nguyenv:
This is a subset of https://github.com/TileDB-Inc/TileDB-Py/pull/1188. Includes changes from #1326.


tiledb/cc/domain.cc
```diff
@@ -14,57 +14,193 @@ using namespace tiledbpy::common;
...
+
+      .def("_domain_to_str", &Dimension::domain_to_str);
+
+  py::class_<Domain>(m, "Domain")
+      .def(py::init<Context &>(), py::keep_alive<1, 2>())
 
-      .def_property_readonly("ncell",
+      .def(py::init<const Context &, py::capsule>(), py::keep_alive<1, 2>())
+
+      .def("__capsule__",
```
ihnorton, line 181: Does this need a `keep_alive`?

tiledb/dimension.py
```diff
@@ -0,0 +1,365 @@
...
+        :dtype: the Dim numpy dtype object, type object, or string \
+            that can be corerced into a numpy dtype object
+        :raises ValueError: invalid domain or tile extent
+        :raises TypeError: invalid domain, tile extent, or dtype type
+        :raises: :py:exc:`TileDBError`
+        :param tiledb.Ctx ctx: A TileDB Context
+
+        """
+        self._ctx = ctx or default_ctx()
+        _cctx = lt.Context(self._ctx, False)
```
ihnorton, line 114: Isn't this going to be making a copy?

ihnorton: I guess maybe not due to `False`?... But then why do we need a second pybind11 object created at all here? Ideally we should be able to just pass in `self._ctx` on [line 177](https://github.com/TileDB-Inc/TileDB-Py/pull/1327/files#diff-d1da7d0bde5761bf12df07572dd2ee80138056d625c2d9a83f64f066facbb22bR177) rather than unwrapping and re-wrapping it here.

ihnorton (commented):
Let's talk through the design a bit more here.
- `_lt_obj` arguments don't quite seem right. `__init__` methods should take arguments that are used to construct the class, and those `_lt_obj` arguments could be moved into a staticmethod factor constructor which forwards to `__init__`. But I want to make sure I understand why those are necessary (similar to the next point)
-

tiledb/cc/domain.cc
```diff
@@ -14,57 +14,197 @@ using namespace tiledbpy::common;
...
+            case TILEDB_FLOAT64: {
+              auto dom = dim.domain<double>();
+              return py::make_tuple(dom.first, dom.second);
+            }
+            case TILEDB_FLOAT32: {
+              auto dom = dim.domain<float>();
+              return py::make_tuple(dom.first, dom.second);
+            }
+            case TILEDB_STRING_ASCII: {
+              // auto dom = dim.domain<std::string>();
```
ihnorton, line 105: nit: remove

tiledb/cc/domain.cc
```diff
@@ -14,57 +14,197 @@ using namespace tiledbpy::common;
...
+              return py::cast(dim.tile_extent<int8_t>());
+            }
+            case TILEDB_FLOAT64: {
+              return py::cast(dim.tile_extent<double>());
+            }
+            case TILEDB_FLOAT32: {
+              return py::cast(dim.tile_extent<float>());
+            }
+            case TILEDB_STRING_ASCII: {
+              return py::cast("");
```
ihnorton, line 161: double-check this - the kwarg for the constructor requires `None`

tiledb/dimension.py
```diff
@@ -0,0 +1,358 @@
...
+
+            # if the tile extent is specified, cast
+            if tile is not None:
+                tile_size_array = _tiledb_cast_tile_extent(tile, domain_dtype)
+                if tile_size_array.size != 1:
+                    raise ValueError("tile extent must be a scalar")
+
+        super().__init__(self._ctx, name, dim_datatype, domain_array, tile_size_array)
+
+        if filters is not None:
```
ihnorton, line 174: see if we can move this to the constructor

ihnorton (approved):
LGTM with some minor nits. Thanks

tiledb/dimension.py
nguyenv: I think it is easier to just leave it as is since the create function does not take in a filter list argument.

https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/cpp_api/dimension.h#L430-L478

PR #1333, opened 2022-09-27
Do Not Require Attribute in Query Condition To Be Passed to Query

nguyenv:
For example,

```
qc = tiledb.QueryCondition("U < 3")
u_result = A.query(attr_cond=qc, attrs=["I"])[:]
```

The query condition is applied to `U` but we no longer have to pass `attrs=["U"]`.

### DIFF
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -321,10 +321,6 @@ class QueryConditionTree(ast.NodeVisitor):
 
         if att not in self.query_attrs:
             self.query_attrs.append(att)
-        #     raise tiledb.TileDBError(
-        #         f"Attribute `{att}` given to filter in query's `attr_cond` "
-        #         "arg but not found in `attr` arg."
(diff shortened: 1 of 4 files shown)


tiledb/query_condition.py
```diff
@@ -318,10 +320,11 @@ def get_att_from_node(self, node: QueryConditionNodeElem) -> Any:
             raise tiledb.TileDBError(f"Attribute `{att}` not found in schema.")
 
         if att not in self.query_attrs:
-            raise tiledb.TileDBError(
-                f"Attribute `{att}` given to filter in query's `attr_cond` "
-                "arg but not found in `attr` arg."
-            )
+            self.query_attrs.append(att)
+        #     raise tiledb.TileDBError(
```
johnkerl, line 324: nit: remove commented-out bits?

johnkerl (approved):
:rocket

PR #1340, opened 2022-10-05
Move `ArraySchema` to Pure Python (3/3)

nguyenv:


ihnorton (approved):

tiledb/array_schema.py
```diff
@@ -0,0 +1,442 @@
...
+    @property
+    def tile_order(self) -> str:
+        """The tile order layout of the array.
+
+        :rtype: str
+        :raises: :py:exc:`tiledb.TileDBError`
+
+        """
+        layout_string = tiledb_layout_string(self._cell_order)
+        if self.cell_order == "hilbert":
```
gsakkis, line 230: - `self._cell_order` should almost certainly be `self._tile_order` 
- is the comparison correct as is or should it be `if layout_string == "hilbert"`?
cc @nguyenv @ihnorton

PR #1341, opened 2022-10-06
Add Support For `QueryCondition` on Sparse Dimensions

nguyenv:
* Deprecate `attr_cond` in favor of `cond`
* Allow `cond` to take in a string directly instead of using `tiledb.QueryCondition`
* Add `dim()` cast that behaves similarly to `attr()` but specifically for dimensions
* Add tests for overlapping ranges


ihnorton (approved):
💯

PR #1347, opened 2022-10-11
Smidge more highlighting for `attr` for special chars in QCs

johnkerl:


nguyenv:
I would move this down to line 85 

```
    TileDB attribute names are Python valid variables or a ``attr()`` casted string. ``attr("name here")`` should be used if `<str>` contains special characters such as spaces.

        ``attr ::= <variable> | attr(<str>)``
```

PR #1351, opened 2022-10-13
Addition of Utility Function `get_last_ctx_err_str()` for C API

nguyenv:
* Replaces `throw_tiledb_error()` and `_throw_ctx_err()` that was previously in `schema_evolution.cc`

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,9 @@
 ## Bug Fixes
 * Correct writing empty/null strings to array. `tiledb.main.array_to_buffer` needs to resize data buffer at the end of `convert_unicode`; otherwise, last cell will be store with trailing nulls chars [#1339](https://github.com/TileDB-Inc/TileDB-Py/pull/1339)
 
+## Improvements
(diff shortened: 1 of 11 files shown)


HISTORY.md
```diff
@@ -3,6 +3,9 @@
 ## Bug Fixes
 * Correct writing empty/null strings to array. `tiledb.main.array_to_buffer` needs to resize data buffer at the end of `convert_unicode`; otherwise, last cell will be store with trailing nulls chars [#1339](https://github.com/TileDB-Inc/TileDB-Py/pull/1339)
 
+## Improvements
+* Addition of Utility Function `get_last_ctx_err_str()` for C API [#1351](https://github.com/TileDB-Inc/TileDB-Py/pull/1351)
```
ihnorton, line 7: IMO we can omit internals like this unless they are high-impact

ihnorton (approved):
👍

HISTORY.md
ihnorton: (in the future, no need to change here)

PR #1389, opened 2022-10-25
Add support for readinto in FileIO

Shelnutt2:
The readinto is needed for a complete implemention of the `io.RawIOBase` interface.

### DIFF
--- a/tiledb/vfs.py
+++ b/tiledb/vfs.py
@@ -1,3 +1,4 @@
+from copy import copy
 import io
 from typing import List, Optional, Type, TYPE_CHECKING, Union
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


tiledb/vfs.py
```diff
@@ -495,3 +496,28 @@ def write(self, buff: bytes):
...
+        Read bytes into a pre-allocated, writable bytes-like object b, and return the number of bytes read.
+
+        :param buff bytes:
+        """
+        size = len(buff)
+        if not self.readable():
+            raise IOError("Cannot read from write-only FileIO handle")
+        if self.closed:
+            raise IOError("Cannot read from closed FileIO handle")
+
```
ihnorton, line 511: Technically this is also [supposed to fail](https://github.com/python/cpython/blob/ab8a5beb5faf23795ede1e6304aebbcf2e20e0aa/Lib/test/test_file.py#L57-L59) unless the mode is binary, but I don't think it is possible to open a VFS without the `b` flag.

tiledb/vfs.py
```diff
@@ -495,3 +496,28 @@ def write(self, buff: bytes):
...
+        nbytes_left = self._nbytes - self._offset
+        nbytes = nbytes_left if size < 0 or size > nbytes_left else size
```
ihnorton, line 513: I don't think `size < 0` is possible for `len(buff)`?

tiledb/vfs.py
Shelnutt2: Doesn't `readable()` check that its `rb` mode?

ihnorton (approved):

PR #1390, opened 2022-10-25
Release 0.17.6

nguyenv:
* This is a release branch that will not be merged into dev
* This is still linked against 2.11.3
* It includes bug fixes for moving `Attr` out of Cython and issues with writing empty/null strings to arrays

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,8 +1,8 @@
-# In Progress
+# TileDB-Py 0.17.6 Release Notes
 
 ## Bug Fixes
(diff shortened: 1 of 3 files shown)


ihnorton:
If the windows builds fail here, we might need to target [this commit](https://github.com/TileDB-Inc/TileDB/commit/502b0dda8e28f736f3ea6abadbf24002e740c738) with the curl download fix, on the core release-2.11 branch.

nguyenv:
No merge

PR #1398, opened 2022-10-28
Fix Whitespace Handling in Query Conditions

nguyenv:
- Wrap the query condition in parenthesis prior to parsing with AST to support multi-line continuation and whitespace padding that would be valid in Python if-cond statements


johnkerl (commented):
can we also check for embedded newline?

johnkerl (approved):
✨

PR #1402, opened 2022-10-29
Resizing issues

johnkerl:
Findings from debug on https://github.com/TileDB-Inc/TileDB/pull/3620

### DIFF
--- a/tiledb/core.cc
+++ b/tiledb/core.cc
@@ -1036,7 +1036,7 @@ public:
     tiledb_query_get_status_details(ctx_.ptr().get(), query_.get()->ptr().get(),
                                     &status_details);
 
-    if (status_details.incomplete_reason != TILEDB_REASON_MEMORY_BUDGET) {
+    if (status_details.incomplete_reason == TILEDB_REASON_USER_BUFFER_SIZE) {
 #else
     if (true) {
 #endif
--- a/tiledb/core.cc
+++ b/tiledb/core.cc
@@ -988,13 +988,16 @@ public:
 
       if ((Py_ssize_t)(buf.data_vals_read * buf.elem_nbytes) >
           (Py_ssize_t)buf.data.size()) {
-        throw TileDBError("After read query, data buffer out of bounds: " + name);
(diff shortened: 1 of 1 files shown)


tiledb/core.cc
```diff
@@ -801,9 +801,6 @@ class PyQuery {
     tiledb_datatype_t type;
     uint32_t cell_val_num;
     std::tie(type, cell_val_num) = buffer_type(name);
-    uint64_t cell_nbytes = tiledb_datatype_size(type);
```
johnkerl, line 804: Unused, revealed by a compiler warning -- not related to this PR

tiledb/core.cc
```diff
@@ -988,6 +985,17 @@ class PyQuery {
       buf.data_vals_read += data_vals_num;
       buf.offsets_read += offset_elem_num;
       buf.validity_vals_read += validity_elem_num;
+
+      if ((ssize_t)(buf.data_vals_read * buf.elem_nbytes) >
+          (ssize_t)buf.data.size()) {
+        throw TileDBError("data buffer out of bounds: " + name);
```
ihnorton, line 991: Would you mind to add `After read query, ` before `data buffer` here and below?

ihnorton (approved):
Wording update request for the error message, then LGTM. Thanks

tiledb/core.cc
ihnorton: Thanks

tiledb/core.cc
```diff
@@ -988,6 +985,17 @@ class PyQuery {
       buf.data_vals_read += data_vals_num;
       buf.offsets_read += offset_elem_num;
       buf.validity_vals_read += validity_elem_num;
+
+      if ((ssize_t)(buf.data_vals_read * buf.elem_nbytes) >
```
nguyenv, line 989: Use `Py_ssize_t` instead of `ssize_t`. This should fix the build errors on Windows, Python 3.10.

nguyenv (changes requested):

PR #1411, opened 2022-11-01
Move `Attr` From Cython to Pure Python

nguyenv:
* Revert "Revert "Move `Attr` From Cython to Pure Python (#1326)" (#1372)"
* This reverts commit ec35345b628a473aafb9f3c9907924762a88aa39
* No longer errors out with running Dask unit tests: `pytest -v --doctest-modules --ignore-glob='*/test_*.py' dask -k "diagnostics or tiledb"`


tiledb/tests/test_attribute.py
```diff
@@ -0,0 +1,179 @@
...
+
+import numpy as np
+import pytest
+from numpy.testing import assert_array_equal
+
+import tiledb
+from tiledb.tests.common import assert_captured, DiskTestCase, has_pandas
+
+
+class AttributeTest(DiskTestCase):
```
ihnorton, line 11: Would you mind to add a check for whether we can modify an attribute that is already in an ArraySchema? I think that should be blocked, but not 100% (it _will_ be blocked by libtiledb in the future)

ihnorton (approved):
Some minor nits from a look-over (even though this is a re-re-revert)

PR #1412, opened 2022-11-02
Passing Empty List to `multirange_index` and `df` Should Return Empty Results

nguyenv:
The goal of this PR is to ensure that empty results are returned from `multirange_index` and `df` when an empty list or `tiledb.EmptyRange` is passed.

### DIFF
--- a/tiledb/tests/test_multi_index.py
+++ b/tiledb/tests/test_multi_index.py
@@ -903,6 +903,7 @@ class TestMultiRange(DiskTestCase):
         with tiledb.open(uri, mode="r") as A:
             assert all(A.query(use_arrow=True).df[:][""] == data)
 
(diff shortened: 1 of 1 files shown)


tiledb/multirange_indexing.py
```diff
@@ -32,10 +32,17 @@
...
     DataFrame = None
 
-
-# sentinel value to denote selecting an empty range
 EmptyRange = object()
 
+
+def is_empty_range(idx: Any) -> bool:
+    if idx is not EmptyRange:
+        if hasattr(idx, "__len__") and len(idx) == 0 and idx != "":
```
ihnorton, line 40: Please double-check numpy semantics here -- is it any empty collection, or only `list`?

kounelisagis: `Union[EmptyRange, List, Tuple]` works fine after changing type of `EmptyRange` from `object` to `TypeVar`.

PR #1425, opened 2022-11-03
Explicitly Use Python 3.10 For Nightly Builds

nguyenv:
* Fix for #1418
* Explicitly use Python 3.10. Otherwise, default is now to use Python 3.11 which we (and PyArrow) do not yet fully support.


nguyenv:
Proof of working: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/3388820293/jobs/5631235544

ihnorton (approved):

PR #1428, opened 2022-11-07
[2.13] Addition of `Array.delete`

nguyenv:


tiledb/libtiledb.pyx
```diff
@@ -2827,6 +2827,42 @@ cdef class Array(object):
         )
         if rc != TILEDB_OK:
             _raise_ctx_err(ctx_ptr, rc)
+    
+    def delete(self):
```
ihnorton, line 2831: I think this should be a `@staticmethod` which takes a URI and opens the array internally. We cannot enforce usage of the `open` context manager by library users, so IMO it would be better to do the open-and-delete internally so that we don't allow operations on an initialized `Array` object which has been deleted (those _should_ fail, but I'd rather prevent the situation altogether).

ihnorton (approved):

PR #1430, opened 2022-11-08
Addition of `ArraySchema.has_dim`

nguyenv:


tiledb/libtiledb.pyx
```diff
@@ -2743,10 +2743,18 @@ cdef class ArraySchema(object):
 
         return bool(has_attr)
 
+    def has_dim(self, name):
+        """Returns true if the given name is a Dime of the ArraySchema
```
ihnorton, line 2747: nit: `DiMe` -> `Dim`

ihnorton (approved):

PR #1433, opened 2022-11-10
Fix `.df` and `.multi_index` Returning Attrs Used In Query Condition

nguyenv:
When using query conditions with `.df` and `.multi_index`, the attribute that was applied in condition would be returned in the result regardless of what was passed to the `attr` argument.

For example,
```
with open.tiledb(uri, "r") as A:
   cond = "att < 10"

   result = A.query(cond=cond, attr=[])[:]
   # with the regular indexer, this correctly did not return any attributes back
   assert "att" not in result 

   result = A.query(cond=cond, attr=[]).df[:]
   # this PR's bug fix: despite asking for no attributes back, the results would still 
   # return an `att` column which would cause this assertion to fail
   assert "att" not in result
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,7 @@
 # In Progress
 
 ## Bug Fixes
-* Fix `.df` and `.multi_index` always returning attributes applied in query conditions []()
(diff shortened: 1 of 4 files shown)


ihnorton (approved):
👍 thanks

aaronwolen (approved):
Thank you!

johnkerl (approved):

PR #1435, opened 2022-11-10
Remove Vendored `string_view` Implementation

nguyenv:


tiledb/core.cc
```diff
@@ -29,7 +29,6 @@
 #include <tiledb/tiledb_serialization.h> // C
 #endif
 
-#include "../external/string_view.hpp"
```
ihnorton, line 32: Let's remove the file from `external` too

ihnorton (approved):
Thanks!

PR #1483, opened 2022-12-01
Check for uninitialized query state after submit, and error out

ihnorton:


nguyenv (approved):

tiledb/core.cc
```diff
@@ -1361,14 +1370,18 @@ class PyQuery {
...
+    if (array_->query_type() == TILEDB_READ) {
       submit_read();
-    else if (array_->query_type() == TILEDB_WRITE)
+    } else if (array_->query_type() == TILEDB_WRITE) {
       submit_write();
-    else if (array_->query_type() == TILEDB_DELETE)
+    } else if (array_->query_type() == TILEDB_DELETE) {
       query_->submit();
-    else
+      if (query_->query_status() == Query::Status::UNINITIALIZED) {
```
nguyenv, line 1379: Is this placed only in the `else if (array_->query_type() == TILEDB_DELETE)` block on purpose?

ihnorton: The other branches don’t call submit directly right here

Shelnutt2 (approved):

PR #1496, opened 2022-12-07
Use `winmode=0` For Loading Full Path

nguyenv:


XanthosXanthopoulos:
I believe the CDLL call `ctypes.CDLL(lib_name)` at line 39 must change since this is the one that will most probably check the `PATH` environment variable. Thats is how I fixed it on my machine. The CDLL call at line 36 is succeeding even without the `winmode=0` since the direct paths to dlls are allowed (at least for now). 

This however causes another import error when trying to import `tiledb` and the working directory is within the TileDB-Py forlder. 

> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "C:\Users\root\Desktop\repos\TileDB-Py\tiledb\__init__.py", line 36, in <module>
>     from .cc import TileDBError
> ImportError: cannot import name 'TileDBError' from 'tiledb.cc' (unknown location)

PR #1503, opened 2022-12-09
Fix webp filter options

shaunrd0:
Fixed by updating `filter.py` to remove use of `__capsule__` and `lt.Context`. Added some new unit tests to protect against this in the future. 

Also updated type of `input_format` to use the `WebpInputFormat` enum, but we still support setting input format using integral values and a test for this was also added in this PR.


nguyenv (approved):
Thank you!

PR #1512, opened 2022-12-12
Install Using `python setup.py develop` In Tests

nguyenv:
* On macOS with Python 3.11, using `python setup.py install` to build from source errors out. Replace with using `develop` instead


nguyenv:
Proof of working on macos with Python 3.11 when building from source: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/3697448698/jobs/6262461644

ihnorton (approved):

PR #1520, opened 2022-12-14
Use subarray API to set ranges

jp-dark:
Remove deprecated query C-API functions and replace them with the equivalent subarray C-API functions.


ihnorton (approved):
LGTM, @nguyenv please take a look and merge if you ✅

nguyenv (approved):

PR #1528, opened 2022-12-16
Remove Fragment Info `load` With Encryption

nguyenv:
This has now been removed from core. We can remove from our code without breaking anything because this was only used internally.


nguyenv:
Now part of #1529

PR #1514, opened 2022-12-13
Ensure NumPy Array Matches ArraySchema Dims For Dense Writes

nguyenv:

PR #1529, opened 2022-12-16
Modify `WebP` Test Inputs To Match Dims With `ArraySchema`

nguyenv:
* We now check to ensure NumPy array inputs match the dimensions in the
  `ArraySchema` for dense writes as of https://github.com/TileDB-Inc/TileDB-Py/pull/1514

### DIFF
--- a/tiledb/fragment.cc
+++ b/tiledb/fragment.cc
@@ -122,15 +122,6 @@ private:
     }
   }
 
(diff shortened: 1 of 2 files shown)


shaunrd0 (approved):

ihnorton (approved):
Thanks

PR #1545, opened 2023-01-05
Use Safe `repr` If Parent Pybind11 Constructor Fails

nguyenv:
* This fixes a larger scale issue that was originally caught in the SOMA unit tests https://github.com/single-cell-data/TileDB-SOMA/pull/653
* When the parent Pybind11 constructor fails, Pytest prints a backtrace of all the arguments in `super().__init__()` which includes the `self` object. Since the object does not exist, using the Pybind11 wrapped properties, namely `_filters`, results in a segfault
* We use `self._ctx`, which is a property that only exists on the pure Python object, to check if the object has been properly instantiated

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+# In Progress
+
+## Bug Fixes
+* Introduce safe repr for classes for `Filter`, `FilterList`, `Dim`, `Domain`, and `Attr`
+
 # Release 0.19.1
 
 ## TileDB Embedded updates:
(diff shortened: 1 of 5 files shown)


nguyenv:
A few additional comments:

- We do actually use `Dim._ctx` [here](https://github.com/TileDB-Inc/TileDB-Py/blob/viviannguyen/segfault-in-dim-constructor/tiledb/domain.py#L61)
- Updating `self._ctx` after the `super()` ended up making the code more confusing to follow, so I've only set it to `None` when it errors out when calling the parent constructor

johnkerl (commented):
Wow thanks @nguyenv !! This is "super"!!!! :D

PR #1548, opened 2023-01-06
CtxMixin

gsakkis:
- DRY-fy https://github.com/TileDB-Inc/TileDB-Py/pull/1545
- Apply the fix to two more classes with the same issue (`Domain`, `Group`)

### DIFF
--- a/tiledb/attribute.py
+++ b/tiledb/attribute.py
@@ -232,7 +232,7 @@ class Attr(CtxMixin, lt.Attribute):
         """
         return self._tiledb_dtype == lt.DataType.STRING_ASCII
 
-    def _repr(self):
+    def __repr__(self):
         filters_str = ""
         if self.filters:
             filters_str = ", filters=FilterList(["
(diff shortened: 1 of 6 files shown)


tiledb/attribute.py
```diff
@@ -46,15 +46,13 @@ def __init__(
         :raises: :py:exc:`tiledb.TileDBError`
 
         """
-        self._ctx = ctx or default_ctx()
-        _dtype = None
-
         if _capsule is not None:
-            return super().__init__(self._ctx, _capsule)
+            return super().__init__(ctx, _capsule)
```
nguyenv, line 50: Shouldn't these still be `self._ctx`?

nguyenv (commented):
Thank you for the refactoring and clean up. This looks really good. I should have looked at this before I merged in my other PR.

I just have one question to clarify about using `ctx` instead of `self._ctx` below.

tiledb/attribute.py
gsakkis: There is no `self._ctx` yet, it is set in `super().__init__`

ihnorton (approved):

PR #1560, opened 2023-01-10
Add test for SC-23827: default override w/ AWS_{DEFAULT_}REGION

ihnorton:
[ci skip] until other side merged in libtiledb.

### DIFF
--- a/tiledb/tests/test_fixes.py
+++ b/tiledb/tests/test_fixes.py
@@ -146,6 +146,10 @@ class FixesTest(DiskTestCase):
         with tiledb.open(uri) as arr:
             tm.assert_frame_equal(arr.df[:], df)
 
+    @pytest.mark.skipif(
+        tiledb.libtiledb.version() < (2, 14, 0),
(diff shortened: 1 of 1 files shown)


tiledb/tests/test_fixes.py
```diff
@@ -143,3 +145,25 @@ def test_py1078_df_all_empty_strings(self):
         tiledb.from_pandas(uri, df)
         with tiledb.open(uri) as arr:
             tm.assert_frame_equal(arr.df[:], df)
+
+    def test_sc23827_aws_region(self):
```
johnkerl, line 149: Sorry, I don't understand

This changes a unit-test file without adding any non-test code; PR description is skip until non-test code is added; I don't see any `skip` syntax here ...

ihnorton: Apologies for the confusion, see https://github.com/TileDB-Inc/TileDB/pull/3788 (intended to open in immediate succession, but saw a nit there)

I'm testing it here because I don't have a good way to test this in C++ (no process library).

johnkerl (approved):

PR #1565, opened 2023-01-11
Use subarray to set ranges in PyQuery

jp-dark:
Use a subarray to add ranges in PyQuery instead of deprecated `Query::add_range` calls.


ihnorton (approved):
LGTM, thank you!

PR #1580, opened 2023-01-18
Add python/pybind Subarray classes

jp-dark:
Add PyBind Subarray class and use when querying on ranges.


tiledb/subarray.py
```diff
@@ -0,0 +1,80 @@
...
+        ctx: "Ctx" = None,
+    ):
+        """Class representing a subarray of a TileDB Array.
+
+        :param array: tiledb.Array the subarray is defined on
+        :param ctx: A TileDB Context
+        """
+        self._array = array
+        self._cpp_array = ArrayImpl(array)
+        super().__init__(ctx, self._cpp_array)
```
gsakkis, line 41: The `_cpp_array` attribute is referenced only here, any reason to store it instead of `super().__init__(ctx, ArrayImpl(array))`?

tiledb/_array.py
```diff
@@ -0,0 +1,22 @@
...
+from .ctx import CtxMixin
+
+if TYPE_CHECKING:
+    from libtiledb import Ctx
+
+
+class ArrayImpl(CtxMixin, lt.Array):
+    """
+    Represents a TileDB array.
+    """
```
gsakkis, line 14: 1. What is the purpose of this class, vs the Cython `Array`? It seems it just wraps an `Array` with a `Ctx`, but `ctx` is also passed to the `super().__init__` of `Subarray` so I'm not clear why it is needed at all. 

2. Currently this is used only by `Subarray`. Unless its use is to be expanded in the future, I'd move it to `subarray.py` (assuming it is actually needed).

3. Purpose and location aside, the name `ArrayImpl` seems misleading. Unlike `DenseArrayImpl` & `SparseArrayImpl`, this doesn't implement anything. Perhaps `ArrayWrapper`?

tiledb/subarray.py
```diff
@@ -0,0 +1,80 @@
+from numbers import Real
+from typing import (
+    TYPE_CHECKING,
+    Sequence,
+    Tuple,
```
gsakkis, line 5: If you remove the trailing comma `black` will format it in one line

tiledb/_array.py
jp-dark: `ArrayWrapper` is a better name. Long-story short: I need a C++ Array object to create a C++ Subarray object. This is a placeholder until we convert the rest of the Cython to PyBind. Depending on the order of conversion, it may get used for other PyBind classes or may be subsumed into the PyBind Array class (classes?) when they are fully implemented.

If you have a better way to do this, I am open to alternatives.

tiledb/cc/query.cc
```diff
@@ -16,237 +16,70 @@ using namespace tiledb;
...
-             auto ctx = q.ctx(); // NB this requires libtiledb >=2.6
-             tiledb_subarray_t *subarray = nullptr;
-             ctx.handle_error(tiledb_query_get_subarray_t(
-                 ctx.ptr().get(), q.ptr().get(), &subarray));
-             ctx.handle_error(tiledb_subarray_set_subarray(
-                 ctx.ptr().get(), subarray, const_cast<void *>(a.data())));
-             ctx.handle_error(tiledb_query_set_subarray_t(
-                 ctx.ptr().get(), q.ptr().get(), subarray));
+      .def("set_subarray",
+           [](Query &query, const Subarray &subarray) {
```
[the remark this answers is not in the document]
jp-dark: I'm pretty sure we perform a deep copy when we set the subarray.

tiledb/_array.py
gsakkis: You can just instantiate directly the underlying C++ Array; all `CtxMixin` is buying you here is the `default_ctx()` call if `ctx` is None, which you can do in one line:
```diff
diff --git a/tiledb/subarray.py b/tiledb/subarray.py
index 9b77939..b0071fa 100644
--- a/tiledb/subarray.py
+++ b/tiledb/subarray.py
@@ -5,8 +5,7 @@ import numpy as np
 
 import tiledb.cc as lt
 
-from ._array import ArrayWrapper
-from .ctx import Ctx, CtxMixin
+from .ctx import Ctx, CtxMixin, default_ctx
 from .libtiledb import Array
 
 Scalar = Real
@@ -18,18 +17,13 @@ class Subarray(CtxMixin, lt.Subarray):
     Represents a TileDB subarray.
     """
 
-    def __init__(
-        self,
-        array: Array,
-        ctx: Ctx = None,
-    ):
+    def __init__(self, array: Array, ctx: Ctx = None):
         """Class representing a subarray of a TileDB Array.
 
         :param array: tiledb.Array the subarray is defined on
         :param ctx: A TileDB Context
         """
-        self._array = array
-        super().__init__(ctx, ArrayWrapper(array))
+        super().__init__(ctx, lt.Array(ctx or default_ctx(), array))
 
     def add_dim_range(self, dim_idx: int, dim_range: Range):
         """Add a range to a dimension of the subarray.
```

tiledb/subarray.py
```diff
@@ -29,24 +21,33 @@ class Subarray(CtxMixin, lt.Subarray):
...
+        """Add a range to a dimension of the subarray.
 
-        :param key: dimension index (int) or name (str)
-        :rtype: np.uint64
+        :param dim_idx: dimension index (int) of the dimension to add the range to
+        :param range: tuple containing the inclusive range to query on
+        :raises: :py:exc:`tiledb.TileDBError`
         """
-        return self._range_num(key)
+        self._add_dim_range(dim_idx, dim_range)
```
gsakkis, line 41: Except for `__init__` and `add_ranges`, every other method in this class just delegates to the respective PyBind11 method. Is there a convention to have all public methods implemented in Python (or equivalently, all C++ methods be private), e.g for easier documentation?

tiledb/_array.py
jp-dark: Done.

tiledb/cc/query.cc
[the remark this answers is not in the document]
jp-dark: I double checked in core. We do perform a deep copy.

tiledb/subarray.py
gsakkis: AFAIK there's nothing that requires or encourages having "private" parent class/methods in PyBind11 and public child ones in pure Python. I'm not too keen on extra levels of indirection without a concrete benefit but I'll defer to @nguyenv and @ihnorton.

jp-dark: It looks like you can actually [add docstrings to the pybind class](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#generating-documentation-using-sphinx). I'll remove the wrappers functions and make the pybind methods public in a future PR.

tiledb/subarray.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        array: Array,
+        ctx: Ctx = None,
+    ):
+        """Class representing a subarray of a TileDB Array.
+
+        :param array: tiledb.Array the subarray is defined on
+        :param ctx: A TileDB Context
+        """
+        self._array = array
```
gsakkis, line 30: (nit: unused attribute)

ihnorton (approved):
Nice work, thank you!

PR #1583, opened 2023-01-19
Bump setup.py to 2.14.0 too

ihnorton:


tiledb/libtiledb.pxd
```diff
@@ -18,6 +18,7 @@ cdef extern from "tiledb/tiledb.h":
     enum: TILEDB_VAR_NUM
     unsigned int tiledb_var_num()
 
+    # TODO: remove after TileDB 2.15
```
nguyenv, line 21: Is there a way to do add a statement like `assert tiledb.libtiledb.version() < (2, 15, 0)` so that this TODO doesn't slip unnoticed?

ihnorton: I couldn't find a way to do that in .pxd

PR #1600, opened 2023-01-24
Replace Deprecated libtiledb Array Open Functions

nguyenv:
* Set encryption key in config
* Use `tiledb_array_set_open_timestamp_start` to set timestamp
* `tiledb_array_consolidate_metadata_with_key` --> `tiledb_array_consolidate_with_key`
* `tiledb_array_reopen_at` -> `tiledb_array_reopen`

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -1607,7 +1607,7 @@ cdef ArrayPtr preload_array(uri, mode, key, timestamp, ctx=None):
     # key
     cdef bytes bkey
     cdef tiledb_encryption_type_t key_type = TILEDB_NO_ENCRYPTION
-    cdef void* key_ptr = NULL
+    cdef const char* key_ptr = NULL
     cdef unsigned int key_len = 0
 
     # convert python mode string to a query type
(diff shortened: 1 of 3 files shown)


ihnorton (approved):

tiledb/libtiledb.pyx
```diff
@@ -1659,6 +1660,25 @@ cdef ArrayPtr preload_array(uri, mode, key, timestamp, ctx=None):
     rc = tiledb_array_alloc(ctx_ptr, uri_ptr, &array_ptr)
     if rc != TILEDB_OK:
         _raise_ctx_err(ctx_ptr, rc)
+    
+    cdef tiledb_config_t* config_ptr = NULL
+    cdef tiledb_error_t* err_ptr = NULL
+    if key is not None:
+        rc = tiledb_config_alloc(&config_ptr, &err_ptr)
```
ihnorton, line 1667: This context is leaking, we need to free it after `tiledb_array_set_config`. Will push a fix in a sec.

nguyenv: Thank you for catching and fixing that.

PR #1603, opened 2023-01-26
Extend DenseArray/SparseArray from CloudArray at tiledb import time

gsakkis:
Replace the dynamic injection of `CloudArray` as a base of `DenseArray/SparseArray` the first time an array instance of the respective class is created with a static definition at tiledb import time. The trick was to do it at the end of `tiledb/__init__.py`, after `tiledb` is registered as a (`pkgutil`-style) namespace package (`__path__ = __import__('pkgutil').extend_path(__path__, __name__)`).

Output of `python -c "import tiledb; print(tiledb.DenseArray.mro())"`
- Without `tiledb.cloud` installed:
```
$ python -c "import tiledb; print(tiledb.DenseArray.mro())"
[<class 'tiledb.libtiledb.DenseArrayImpl'>, <class 'tiledb.libtiledb.Array'>, <class 'object'>]
```
- With `tiledb.cloud` installed:
```
[<class 'tiledb.DenseArray'>, <class 'tiledb.libtiledb.DenseArrayImpl'>, <class 'tiledb.libtiledb.Array'>, <class 'tiledb.cloud.cloudarray.CloudArray'>, <class 'object'>]
```


tiledb/__init__.py
```diff
@@ -114,3 +115,17 @@
 #
 # Note: 'pip -e' in particular will not work without this declaration:
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)
+
+try:
```
thetorpedodog, line 119: Could you add a comment here explaining why we're doing this? (e.g. "If tiledb.cloud is installed, add CloudArray methods to TileDB arrays")

gsakkis: :heavy_check_mark:

thetorpedodog (approved):
You may wish to get an OK from somebody on the core libraries team, but this is good as far as I can tell.

ihnorton (approved):

PR #1616, opened 2023-02-01
Update stats test for 2.15, fix SC-24954

ihnorton:
Also re-fix SC-24863


Shelnutt2 (approved):
Changes look good. The referenced shortcut story seems to be the wrong one.

PR #1621, opened 2023-02-02
Datatype refactoring (1/2)

gsakkis:
The handful of functions still left in `tiledb.util`  after the https://github.com/TileDB-Inc/TileDB-Py/pull/1609 cleanup are related to conversions between TileDB and Numpy dtypes, as well as extracting related type properties (min, max, number of cells, is_datetime). 

This PR introduces a new `DataType` dataclass that provides the same functionality in a more unified and consistent API, cleaner internals and (at least theoretically) better efficiency as it uses an LRU cache to store the 128 most recently created `DataType` instances.

To minimize the risk surface, the `tiledb.util` functions are maintained in this PR, delegating the actual work to `DataType`. A subsequent PR will replace `tiledb.util` with direct calls to `DataType`.


nguyenv (approved):
Thanks for cleaning everything up. Looks great.

PR #1623, opened 2023-02-03
Also set ReadTheDocs build.tools

ihnorton:
RTD `build.os` config option requires setting `build.tools` simultaneously.


ihnorton:
RTD PR builds were off, turning back on to make sure this works.

nguyenv (approved):

PR #1644, opened 2023-02-09
Datatype refactoring (2/2)

gsakkis:
Follow up to https://github.com/TileDB-Inc/TileDB-Py/pull/1621; 274 fewer LOC to maintain.

- Replace `util` module with `datatypes`
- Extend `DataType.from_numpy()` to support "ascii" & "blob":
-  Simplify `Attr`, `Dim` and `DimLabelSchema` constructors using `DataType`; likewise for `GroupMetadata.__getitem__`
- Refactor `dim_for_column`, rename to `create_dim` and reuse it in `_schema_like_numpy`.


ihnorton (approved):

ihnorton:
Thanks @gsakkis.

PR #1684, opened 2023-03-07
Use architecture-specific libtiledb install path to get correct libti…

ihnorton:
…ledb on macOS-arm64

Fixes (correctly) the warning [here](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=30693&view=logs&j=3094b6d3-484e-574d-e9ba-d762c1161906), which leads to an unusable wheel:
```
2023-03-07T17:50:21.2339550Z   ld: warning: ignoring file /Users/runner/work/1/.libtiledb_dist/1fb59c4410823c5f615661997021437d7223041a/lib/libtiledb.dylib, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
```
when the x86_64 artifact gets pulled into the arm64 build variant due to identical cache key.


davisp (approved):
+1

PR #1700, opened 2023-03-13
Restore `__version__` variable.

thetorpedodog:
Someone *cough* accidentally deleted the `__version__` string variable when they refactored the way setuptools-scm handled versions. This brings it back while also subjecting the reader to the word “version” enough to reach semantic satiation.

### DIFF
--- a/tiledb/__init__.py
+++ b/tiledb/__init__.py
@@ -111,6 +111,7 @@ from .subarray import Subarray
 from .version_helper import version
 from .vfs import VFS, FileIO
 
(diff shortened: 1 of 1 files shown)


tiledb/__init__.py
```diff
@@ -111,6 +111,7 @@
 from .version_helper import version
 from .vfs import VFS, FileIO
 
+__version__ = version.version
```
johnkerl, line 114: ❤️

ihnorton (approved):

PR #1715, opened 2023-03-30
Addition of `Group(config=...)` and `Group.set_config`

nguyenv:
* Add `config` parameter to `Group` constructor and `set_config` to the high-level `Group` API. This will help relieve the excessive allocation of `Ctx` objects in the TileDB-SOMA-Py codebase as described in https://github.com/single-cell-data/TileDB-SOMA/issues/1169)
* Group unit tests now use timestamps instead of `time.sleep()`


johnkerl (approved):

johnkerl (approved):
✨

PR #1736, opened 2023-04-15
Fix sc-27374: default order mapping fallback

ihnorton:
Restores prior behavior changed in https://github.com/TileDB-Inc/TileDB-Py/pull/1609.


nguyenv (approved):

johnkerl (approved):
🚀

PR #1748, opened 2023-04-23
[WIP] Use certifi certs if available

ihnorton:
This avoids use of old certificates, matching behavior of many other packages. Fixes runtime certificate error in macOS-arm64 wheels for S3 backend.


ihnorton:
Fixed by https://github.com/TileDB-Inc/TileDB/pull/4059

Issue #1730, opened 2023-04-11
"unexpected object type in string conversion" when writing variable size attribute whose all items have size = 1


rouault:
Given
```python
import tiledb
import numpy as np
attr = tiledb.Attr("bool_field", dtype = np.bool_, var = True)
dimX = tiledb.Dim(name="_X", domain=(1, 4), tile=2, dtype=np.float64)
dimY = tiledb.Dim(name="_Y", domain=(1, 4), tile=2, dtype=np.float64)
dom = tiledb.Domain(dimX, dimY)
schema = tiledb.ArraySchema(domain=dom, sparse=True, attrs=[attr])
tiledb.Array.create("test_bool.tiledb", schema)
d1 = np.array([3,4,1,2])
d2 = np.array([3,4,2,1])
a = np.array(list([np.array([True], dtype=np.bool_), np.array([True], dtype=np.bool_), np.array([True], dtype=np.bool_), np.array([True], dtype=np.bool_)]), dtype=object,)
# a = np.array(list([np.array([True, False], dtype=np.bool_), np.array([True], dtype=np.bool_), np.array([True], dtype=np.bool_), np.array([True], dtype=np.bool_)]), dtype=object,)
with tiledb.open("test_bool.tiledb", 'w') as A:
    A[d1, d2] = {"bool_field": a}
```

this fails with:
```
Traceback (most recent call last):
  File "tiledb/libtiledb.pyx", line 151, in tiledb.libtiledb._write_array
tiledb.cc.TileDBError: Unexpected object type in string conversion (tiledb/npbuffer.cc:288)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "tiledb/libtiledb.pyx", line 2692, in tiledb.libtiledb.SparseArrayImpl.__setitem__
  File "tiledb/libtiledb.pyx", line 2632, in tiledb.libtiledb._setitem_impl_sparse
  File "tiledb/libtiledb.pyx", line 153, in tiledb.libtiledb._write_array
tiledb.cc.TileDBError: Failed to convert buffer for attribute: 'bool_field'
```

This is due to all items in a[] having size 1. Uncommenting the commented line to have a[0] having two items makes it work

Tested on latest tiledb==0.21.2 and several earlier versions

PR #1767, opened 2023-05-04
Fix "unexpected object type in string conversion" error when using booleans

juliusgeo:
This fixes https://github.com/TileDB-Inc/TileDB-Py/issues/1730


juliusgeo:
@ihnorton I believe this is ready for review, it's a bit hard for me to tell which of the CI failures are expected, but it passes locally.

nguyenv (approved):
Thank you for the bug fix. I have rebased against current dev and made slight modifications to the unit test so that it runs properly against our linter.

PR #1774, opened 2023-05-26
Handle empty string passed to Query condition

Shelnutt2:
This treats an empty string `""` the same as None when applying a query condition. This is useful as users are passing strings for query condition and avoids the user needing to validate they've successfully built a condition before passing it in.


ihnorton (approved):
Lgtm as long as @nguyenv is ok with it.

ihnorton (approved):
Lgtm as long as @nguyenv is ok with it.

nguyenv:
I am still seeing `FAILED tiledb/tests/test_query_condition.py::QueryDeleteTest::test_qc_dense_empty - AttributeError: 'list' object has no attribute 'nbytes' `

nguyenv (approved):

PR #1795, opened 2023-06-24
[bugfix] `Group.close` now takes in `should_throw` argument

nguyenv:
* Set default to True


nguyenv:
Opps didn't notice #1793 was already up.

PR #1798, opened 2023-06-26
Fix macOS deployment target mismatch in nightly test build

ihnorton:


Shelnutt2 (approved):

nguyenv:
This is causing CI to fail due to a syntax error in the yaml file.

https://github.com/TileDB-Inc/TileDB-Py/actions/runs/5378233280
```
The workflow is not valid. .github/workflows/ci.yml (Line: 22, Col: 33): Unrecognized named-value: 'os'. Located at position 1 within expression: os == 'macos-11' && python-version in ['3.7', '3.8'] && '11.7' || '10.15'
```
I am taking a look ow

PR #1801, opened 2023-06-26
Correct Syntax Errors in `ci.yaml`

nguyenv:
https://github.com/TileDB-Inc/TileDB-Py/pull/1798#issuecomment-1607612327


nguyenv:
The error on macos-11, Python 3.7 is unrelated to these changes. It will be fixed in a subsequent PR.

PR #1802, opened 2023-06-27
Support variable length dimension label reads

shaunrd0:
Adds support for variable length dimension label reads. This change requires TileDB-Inc/TileDB#4142 to add dimension label support for [result_buffer_elements_nullable](https://github.com/TileDB-Inc/TileDB-Py/blob/dev/tiledb/core.cc#L717) to size read offset buffers. The experimental dimension label variant of this function is defined as `QueryExperimental::result_buffer_elements_nullable_labels` and used in this PR to determine the number of offsets read by the var-size dimension label query.


jp-dark (approved):
Looks good to me!

ihnorton (approved):

ihnorton:
Cherry-picked and merged in https://github.com/TileDB-Inc/TileDB-Py/pull/1805. Thanks @shaunrd0 @jp-dark!

PR #1803, opened 2023-07-03
Fix an online-help typo, and clarify

johnkerl:
Raised by @bkmartinjr on Slack -- it needs to be clearer what version this is. Also, we have a typo.


ihnorton (approved):

ihnorton:
Thanks

PR #1813, opened 2023-08-17
Append nightly failure updates if an open Issue exists

jdblischak:
This PR updates all the nightly GitHub Actions workflows to append to an existing open Issue instead of always creating a new one.

The action [JasonEtco/create-an-issue](https://github.com/JasonEtco/create-an-issue) requires that the existing open Issue has the [exact same title](https://github.com/JasonEtco/create-an-issue/blob/main/action.yml#L20):

>  Update an open existing issue with the same title if it exists

I haven't tested this on my fork yet. If you approve of the strategy, I'll generate some spurious build errors on my fork to confirm it works as expected.

If you'd prefer to adopt this appending behavior but still keep the date of the initial build failure in the Issue title, we could switch to using the `gh` cli approach I have used in the past (eg [in TileDB-MariaDB](https://github.com/TileDB-Inc/TileDB-MariaDB/blob/master/scripts/nightly/open-issue.sh)). I'm able to allow different titles because I instead search for open Issues with the specific label [nightly-failure](https://github.com/TileDB-Inc/TileDB-MariaDB/issues?q=is%3Aissue+label%3Anightly-failure+is%3Aclosed)

### DIFF
--- a/.github/workflows/daily-test-build-issue-template.md
+++ b/.github/workflows/daily-test-build-issue-template.md
@@ -1,8 +1,9 @@
 ---
-title: Nightly GitHub Actions Build Fail on {{ date | date('ddd, MMMM Do YYYY') }}
+title: Nightly GitHub Actions Build Fail
 assignees: nguyenv, ihnorton
 labels: bug
 ---
 
+The nightly GitHub Actions build failed on {{ date | date('ddd, MMMM Do YYYY') }}.
 See run for more details:
 https://github.com/{{ env.GITHUB_REPOSITORY }}/actions/runs/{{ env.GITHUB_RUN_ID }}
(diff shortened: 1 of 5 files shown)


nguyenv:
Yes, I definitely approve of this update. Thanks for working on it.

ihnorton:
If possible, I'd like to have separate comments like we do in mariadb, so that a new notification is generated for each failure. I think it would be worth the time to make a parametrized action for this pattern, because we will want to expand the use of such test jobs with notification. I did find [one existing action](https://github.com/jayqi/failed-build-issue-action) which almost does what we want, but the input and the existing issue title are not customizable.

Thanks for working on this.

jdblischak:
> I think it would be worth the time to make a parametrized action for this pattern, because we will want to expand the use of such test jobs to other repos (with the same notification pattern).

Good idea. I've already copy-pasted this Issue-opening code to various repos, so it's better to consolidate in one place. I'll add it to [TileDB-Inc/github-actions](https://github.com/TileDB-Inc/github-actions)

Creating a reusable, parametrized action will take more effort than it took to open this PR. I propose that we merge this PR as an incremental improvement, and then I'll send a follow-up PR once the action is ready.

ihnorton (approved):

jdblischak:
Note that I can't self-merge because I'm not a collaborator on this repo

Issue #1816, opened 2023-08-17
Feature: Support `pathlib.Path`


Nicholas-Schaub:
### Summary

When reading/writing arrays, `tiledb` does not support `pathlib.Path` for file URIs.

### Details

When using a `pathlib.Path` to open a new array, the first write triggers an array creation event that fails with the following error (truncated):

```bash
  File "tiledb/libtiledb.pyx", line 962, in tiledb.libtiledb.Array.create
  File "tiledb/libtiledb.pyx", line 527, in tiledb.libtiledb.unicode_path
  File "tiledb/libtiledb.pyx", line 520, in tiledb.libtiledb.ustring
TypeError: ustring() must be a string or a bytes-like object, not <class 'pathlib.PosixPath'>
```

### What I Expect to Happen

When opening a `tiledb.Array`, I should be able to use a `pathlib.Path` object. Additionally, using the `os` pathing objects should also be possible.

### Implementation

The stack trace seems to indicate the code is written in Cython. This should be a relatively easy feature to implement with two lines of code to type check if the input object is `pathlib.Path`, then casting with `str(pathlib.Path("path/to/file.tiledb"))`.

If you're open to taking PRs, I am happy to contribute.

nguyenv:
Hi @Nicholas-Schaub,

Thank you for reporting this issue. Yes, we are open to contributions! Please feel free to put in a PR.

Thanks.

thetorpedodog:
The VFS change addresses this in part; I will probably expand essentially the same thing to the rest of the TileDB library in a follow-up.

PR #1818, opened 2023-08-22
Make VFS accept path-like objects to refer to files.

thetorpedodog:
When working on a change in tiledbsoma [1], we ran into somewhat mysterious error messages where things had worked before. This was because, unlike `open` (and other libraries which accept path-likes), `tiledb.VFS` only accepts `str` (and undocumented `bytes`) paths. Accepting path-like objects (like `pathlib.Path`s) lets us play better with other libraries and user code.

[1]: https://github.com/single-cell-data/TileDB-SOMA/pull/1629

### DIFF
--- a/tiledb/vfs.py
+++ b/tiledb/vfs.py
@@ -1,10 +1,13 @@
 import io
+import os
 from types import TracebackType
 from typing import List, Optional, Type, Union
 
 import tiledb.cc as lt
 
 from .ctx import Config, Ctx, default_ctx
+
+_AnyPath = Union[str, bytes, os.PathLike]
 
 
 class VFS(lt.VFS):
@@ -29,7 +32,7 @@ class VFS(lt.VFS):
             else:
                 try:
                     config = dict(config)
-                except:
+                except Exception:
                     raise ValueError("`config` argument must be of type Config or dict")
 
             ccfg = lt.Config(config)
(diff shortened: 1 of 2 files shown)


nguyenv:
Related to #1816.

tiledb/vfs.py
```diff
@@ -1,11 +1,14 @@
 import io
+import os
 from types import TracebackType
 from typing import List, Optional, Type, Union
 
 import tiledb.cc as lt
 
 from .ctx import Config, Ctx, default_ctx
 
+AnyPath = Union[str, bytes, os.PathLike]
```
nguyenv, line 10: More of a note to self: We'll want to expand support `pathlib.Path` / path-like objects for other functions like `tiledb.open`, etc. eventually. OK to keep this typing definition in here for now but should probably be moved to a new `_types.py` in the future.

nguyenv (approved):
Looks good to me.

ihnorton (approved):
💯 thanks!

tiledb/vfs.py
thetorpedodog: In order to make this not look like it is intended to be used by any external users, I renamed this to `_AnyPath`, with no other changes.

thetorpedodog:
I don’t have the power to merge changes, so one of you will have to do it once the tests go through.

PR #1820, opened 2023-08-28
Fix nightly build syntax error

ihnorton:


.github/workflows/daily-test-build-numpy.yml
```diff
@@ -69,7 +69,6 @@ jobs:
 
       - name: Test without pandas
         run: python -m pip uninstall -y pandas
-        if: ${{ matrix.uninstall_pandas }}
```
nguyenv, line 72: I don't think this line should be deleted. We want to run the unit tests with and without pandas.

ihnorton: This file is not parameterized on that AFAICT...

PR #1824, opened 2023-09-06
Add `read_subarray` and `write_subarray` methods to arrays

jp-dark:


jp-dark:
I'm going to add more tests to this (lots of edge cases), but here is a preview of the `read_subarray` and `write_subarray` methods.

PR #1825, opened 2023-09-07
Avoid importing Pandas until we actually use it.

thetorpedodog:
Importing Pandas takes a significant amount of time; nearly half a second in UDFs. Previously, we imported it eagerly if it was present, even though outside of a couple functions it was only used for type annotations. This change ensures that we only import it exactly when we actually use it.


ihnorton (approved):
Thanks

PR #1826, opened 2023-09-08
Make pandas lazy import check conditional on tiledb.cloud version

ihnorton:


tiledb/tests/test_basic_import.py
```diff
@@ -2,12 +2,32 @@
...
 
+import importlib_metadata
+from packaging.version import Version
+
 
 def test_dont_import_pandas() -> None:
     """Verifies that when we import TileDB, we don't import Pandas eagerly."""
+
+    # If tiledb.cloud < 0.10.21 is installed, we should remove it from the module list
+    # before running the test, because it eagerly imported pandas before that version.
```
thetorpedodog, line 13: maybe phrase this as "prevent it from being imported" rather than "remove it from the module list"

tiledb/tests/test_basic_import.py
```diff
@@ -2,12 +2,32 @@
...
+    # Note that we import tiledb.cloud within tiledb-py, if available, in order to hook
+    # Array.apply and other functionality.
+    try:
+        ver = importlib_metadata.version("tiledb.cloud")
+    except importlib_metadata.PackageNotFoundError:
+        ver = None
```
thetorpedodog, line 19: Since it's OK to import `tiledb.cloud` in this part of the test, I would do that and look at `tiledb.cloud.__version__` rather than going through `importlib_metadata`.

Note that if you *did* want to continue using `importlib_metadata`, you would need to refer to `tiledb-cloud` in place of `tiledb.cloud`, since `importlib_metadata` is looking at the metadata of the *package* rather than the *module* `tiledb.cloud`. As it stands I don’t think this code will *ever* suppress the import, because `tiledb.cloud` will never be found.

tiledb/tests/test_basic_import.py
```diff
@@ -2,12 +2,32 @@
...
+    if ver and Version(ver) < Version("0.10.21"):
+        maybe_cloud = "sys.modules['tiledb.cloud'] = None;"
```
thetorpedodog, line 21: "suppress_cloud" would make this clearer imo

ihnorton:
Addressed, thanks!

thetorpedodog (approved):

PR #1828, opened 2023-09-14
Document "m" For Deleting Groups

nguyenv:


tiledb/group.py
```diff
@@ -22,7 +22,7 @@ class Group(CtxMixin, lt.Group):
 
     :param uri: The URI to the Group
     :type uri: str
-    :param mode: Read mode ('r') or write mode ('w')
+    :param mode: Read mode ('r'), write mode ('w'), or modify exclusive ('m')
```
aaronwolen, line 25: Maybe explicitly mention that `'m'` is for deletes, similar to [delete_fragments](https://tiledb-inc-tiledb.readthedocs-hosted.com/projects/tiledb-py/en/stable/python-api.html#tiledb.libtiledb.Array.delete_fragments)?

aaronwolen (approved):
Had one minor suggestion but thanks for adding this!

Shelnutt2 (approved):

johnkerl (approved):

PR #1832, opened 2023-09-17
Support categoricals in from_pandas

Shelnutt2:
This adds support in from_pandas for converting and handling categorical datatypes.

Notes: There are some deficiencies in pandas pre-2.0, namely they do not store a dtype of the enumeration/categorical values. Instead they are always treated as an object array. As such we need to infer the type and this has limitations, such as being able to easily detect signed vs unsigned integers without manually checking all values. This PR takes a first-pass best effort at dtype conversion for the pandas categorical values. Future PRs are likely to improve this.

Example usage:
```
df = pd.DataFrame(
    {
        "int": [0, 1, 2, 3],
        "categorical_string": pd.Series(["A", "B", "A", "B"], dtype="category"),
        "categorical_int": pd.Series(
            np.array([1, 2, 3, 4], dtype=np.int8), dtype="category"
        ),
    }
)

tiledb.from_pandas("my_array", df)
```

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -3,7 +3,7 @@ import json
 import os
 import warnings
 from dataclasses import dataclass
-from typing import List, Optional, Union
+from typing import Any, List, Optional, Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


ihnorton (approved):

leipzig:
when this is available i will test the open datasets again since categorical variables was a major offender there https://cloud.tiledb.com/notebooks/details/leipzig/ab803874-bccd-4e10-8dd8-40cf633171a2/preview

PR #1841, opened 2023-10-04
[?] Correct Subarray With Negative Domains on Dense Arrays

nguyenv:
* We currently support Python negative indexing ie. where -1 returns the last element in the sequence. However, this conflicts with reading and writing with negative subarray values -- indexing with negative values errors out.
* This PR deprecates the usage of Pythoh negative indexing so that dense arrays with negative domains can be indexed by TileDB dimensions correctly.


tiledb/libtiledb.pyx
```diff
@@ -648,7 +650,9 @@ def index_domain_subarray(array: Array, dom, idx: tuple):
                     raise IndexError("cannot index integral domain dimension with non-integral slice (dtype: {})".format(type(start)))
             # apply negative indexing (wrap-around semantics)
```
nguyenv, line 651: need to correct documentation

PR #1842, opened 2023-10-05
Update clang-format to version 16

jp-dark:
Update clang-format to version 16.


jp-dark:
Note: The previous GitHub action we were using for checking the formatting appears to be broken for clang-format version 15+ so I copied the GitHub action and scripts from the core TileDB repo.

ihnorton (approved):
Thanks!

PR #1843, opened 2023-10-06
Add sparse dimension label example

jp-dark:
Add an example of how to use dimension labels on sparse data.


nkules (approved):
Good examples with the subarray. It would be good to document the differences between reading with subarray and the more traditional query. But this probably isn't the right place for that anyways.

ihnorton (approved):
Thanks

PR #1846, opened 2023-10-20
Allow delete mode for attributes on dense arrays

johnkerl:
Needs unit-test coverage


nguyenv:
Closing as query conditions are not supported for dense arrays.

johnkerl:
@nguyenv do we not want to support delete-by-attribute filter for dense arrays?

nguyenv:
Oh that's a good point. We should support it for attributes. I'll reopen the PR.

kounelisagis:
Query deletes aren't supported for dense arrays in core. Will reopen if needed.

PR #1847, opened 2023-10-24
[MERGE 2.17.3] Addition of `extend_enumeration`, test creating empty enumeration

nguyenv:
Depends on https://github.com/TileDB-Inc/TileDB/pull/4445


nguyenv:
Building against `dev` in `setup.py` is erroring out due to the missing `external/vcpkg` when downloading source from git. There is a known workaround by downloading  and installing `vcpkg` from git. Otherwise, this branch passes locally when using a pre-built version of dev.

tiledb/tests/test_schema_evolution.py
```diff
@@ -170,3 +170,27 @@ def test_schema_evolution_with_enmr(tmp_path):
...
+    schema = tiledb.ArraySchema(domain=domain, attrs=attrs, enums=[enmr])
+    tiledb.Array.create(uri, schema)
+
+    with tiledb.open(uri) as A:
+        assert A.schema.has_attr("a")
+        assert A.attr("a").enum_label == "e"
+        assert A.enum("e") == enmr
+
+    se = tiledb.ArraySchemaEvolution()
+    updated_enmr = enmr.extend(["a", "b", "c"])
```
ihnorton, line 189: Can we add a bad case here (mismatched data type) and assert an error is thrown?

ihnorton (approved):

PR #1848, opened 2023-10-25
Correct Nullable String Attributes and All NULL Data

nguyenv:
* `np.nan_to_num` cannot be used on string arrays
* We should hold off on merging this PR until we get https://github.com/single-cell-data/TileDB-SOMA/pull/1834 passing, and the confirm that [this](https://github.com/TileDB-Inc/TileDB-Py/pull/1848#issuecomment-1781413424) repro script passes

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -2319,14 +2319,17 @@ cdef class DenseArrayImpl(Array):
                                             "typed attribute '{}'!".format(name))
                         
                         if attr.isnullable and name not in nullmaps:
-                            nullmaps[name] = ~np.ma.masked_invalid(attr_val).mask
-                            
-                            if (np.issubdtype(attr.dtype, np.unicode_) 
-                                or np.issubdtype(attr.dtype, np.string_)):
-                                attr_val = np.array(
-                                    ["" if v is None else v for v in attr_val])
+                            try:
+                                nullmaps[name] = ~np.ma.masked_invalid(attr_val).mask
+                            except Exception as exc:
+                                nullmaps[name] = np.array(
+                                    [int(v is not None) for v in attr_val], dtype=np.uint8)
+
+                            if np.issubdtype(attr.dtype, np.string_):
+                                attr_val = np.array(["" if v is None else v for v in attr_val])
                             else:
                                 attr_val = np.nan_to_num(attr_val)
+                                attr_val = np.array([0 if v is None else v for v in attr_val])
                         attr_val = np.ascontiguousarray(attr_val, dtype=attr.dtype)
                 except Exception as exc:
                     raise ValueError(f"NumPy array conversion check failed for attr '{name}'") from exc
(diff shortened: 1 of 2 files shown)


johnkerl:
@nguyenv in the application-level use case I have this:

```
anndata = ...
anndata.obs['new_col'] = pd.Series(data=np.nan, dtype=np.dtype(str))
if ...:
    anndata.obs['new_col'][10] = 'abc'

# Initial
soma_uri = tiledbsoma.io.from_anndata(
    experiment_uri=soma_uri,
    anndata=anndata,
    measurement_name='RNA'
)

# Update
anndata.obs['new_col'] = 'abc'

# saving updated obs
with tiledbsoma.open(soma_uri, 'w') as soma_exp:
    tiledbsoma.io.update_obs(soma_exp, anndata.obs)
```

When that first `if` is `True` -- the `new_col` column is nine NaNs and one `"abc"` on initial write -- all is well

When that first `if` is `False` -- the `new_col` column is all NaNs on initial write -- I still get

```
  File "/Users/johnkerl/./doronanx.py", line 279, in <module>
    tiledbsoma.io.update_obs(soma_exp, anndata.obs)
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/io/ingest.py", line 1408, in update_obs
    _update_dataframe(
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/io/ingest.py", line 1529, in _update_dataframe
    raise ValueError(f"unsupported type updates: {msg}")
ValueError: unsupported type updates: new_col type double != string
```

and I'm still unsure how to solve this at the `tiledbsoma` level (in an experimental branch).

I'm debating whether to accept this PR, with the caveat that we will need another, or, to request changes on this PR ...

johnkerl (changes requested):
As discussed

nguyenv:
I've now corrected the two issues mentioned. 

I do not think these changes will make it into the current release though.

nguyenv:
Thanks for providing the full script. I was able to repro the issue.

nguyenv:
This seems to be a problem residing in TileDB-SOMA's `df_to_arrow` converter. It's actually once again an issue where a column with all `pd.NA` is always [inferred](https://github.com/single-cell-data/TileDB-SOMA/blob/bd13a68667a84b068f4ca27152c527f27652e810/apis/python/src/tiledbsoma/_arrow_types.py#L200) as type `double`, so we not be using `pa.infer_type` there and instead specify an explicit type. There doesn't seem to be a convenient Pandas to Pyarrow dtype converter available (only the other way around), so I think I'll add a dtype mapping in `_arrow_types.py` and open a PR in TileDB-SOMA.

johnkerl:
Ok @nguyenv I'll triple check

I should have mentioned I did this on a soma branch with attempts to handle that

Let me put up the branch for your reference ...

johnkerl (approved):

nguyenv:
I haven't done what I think should be the corrections yet, so I can't say with 100% confidence that the problem is only residing in SOMA. I'm going to make sure we can get the repro script to work before I merge this PR.

johnkerl:
@nguyenv https://github.com/single-cell-data/TileDB-SOMA/pull/1834

johnkerl:
@nguyenv https://github.com/single-cell-data/TileDB-SOMA/pull/1834 passes with this PR installed locally

nguyenv:
I've confirmed both this PR and the PR put forth in TileDB-SOMA are working.

PR #1850, opened 2023-10-30
Correct `Enumeration.extend` to handle Boolean types

nguyenv:


ihnorton (commented):
It seems like this change, we don't get any type checking of the input?

nguyenv:
The type checking still happens in core. It makes sure the size of the passed in values matches the size of the enumeration's dtype. Also in the unit tests, the check for passing in integers when the enumeration type is a string still fails as expected.

nguyenv:
I just updated to cast to the extent values to the dtype of the enumeration. Also, in the unit tests, the check for passing in integers when the enumeration type is a string still fails as expected.

johnkerl (approved):

ihnorton (approved):

PR #1856, opened 2023-11-01
Add `filter_name` to Filter class

jp-dark:
This returns the underlying TileDB string name for the filter types.


ihnorton (approved):

nguyenv:
Can you rebase against `dev`? The unit tests should be fixed by https://github.com/TileDB-Inc/TileDB-Py/pull/1858.

PR #1858, opened 2023-11-02
Use `with_metadata` instead of argument to `pa.schema` constructor

nguyenv:
* Pandas 2.0 has deprecated many functions including passing the metadata as an argument to the `pa.schema` constructor. This has been replaced with a separate call to `with_metadata`


johnkerl (approved):

ihnorton:
Is this backwards-compatible?

nguyenv:
> Is this backwards-compatible?

This passes with Pandas 1.5.3.

johnkerl:
[sc-35946]

ihnorton (approved):
THanks

PR #1864, opened 2023-11-09
Release the GIL while consolidating

davisp:
I stumbled over this while attempting to log stats during consolidation. Turns out when we don't release the gil the background thread can't run.

### DIFF
--- a/tiledb/cc/array.cc
+++ b/tiledb/cc/array.cc
@@ -57,12 +57,12 @@ void init_array(py::module &m) {
       .def("consolidate",
            py::overload_cast<const Context &, const std::string &,
                              Config *const>(&Array::consolidate),
-                             py::call_guard<py::gil_scoped_release>())
+           py::call_guard<py::gil_scoped_release>())
       .def("consolidate",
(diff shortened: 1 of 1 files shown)


ihnorton (approved):

ihnorton:
Cherry-picked to https://github.com/TileDB-Inc/TileDB-Py/pull/1869

tiledb/cc/array.cc
```diff
@@ -56,11 +56,13 @@ void init_array(py::module &m) {
...
            py::overload_cast<const Context &, const std::string &,
-                             Config *const>(&Array::consolidate))
+                             Config *const>(&Array::consolidate),
+           py::call_guard<py::gil_scoped_release>())
       .def("consolidate",
            py::overload_cast<const Context &, const std::string &,
                              tiledb_encryption_type_t, const std::string &,
-                             Config *const>(&Array::consolidate))
+                             Config *const>(&Array::consolidate),
+           py::call_guard<py::gil_scoped_release>())
```
lust4life, line 65: Do we still release the GIL in latest impl? Didn't see related changes

PR #1866, opened 2023-11-14
Fix domain equality check

jp-dark:
Fixes:
* Add checks for matching filters and variable length in `Dim.__eq__`.
* Fix bug where error is thrown when using `Domain.__eq__` on a non-integral sparse domain.
* Fix bug where filters are dropped from dimensions in a domain with all anonymous dimensions.


jp-dark:
[sc-36840] [sc-37022]

ihnorton (approved):
Thanks!

PR #1875, opened 2023-12-01
Add all array properties to ArraySchema.__eq__

jp-dark:


nguyenv (approved):

tiledb/tests/test_pandas_dataframe.py
```diff
@@ -823,7 +823,7 @@ def test_dataframe_csv_schema_only(self):
                 tile_order="row-major",
                 capacity=1001,
                 sparse=True,
-                allows_duplicates=False,
+                allows_duplicates=True,
```
ihnorton, line 826: Is this a leftover?

jp-dark: Previously, the state of `allow_duplicates` was not being checked in this test. The "expected schema" was set to `allows_duplicates=False`, but the default value in the code base for `from_csv` is `allow_duplicates=True`. I am assuming the error is in the test function, not the code base.

PR #1876, opened 2023-12-01
Warn when a process that has imported tiledb calls fork()

sgillies:
TileDB uses multiple threads and it's easy to run into deadlocks on Linux where `fork` is the default for `multiprocessing` until Python 3.14. We want to see a warning like Python 3.12 emits when fork() is called with multiple threads detected (see the implementation in posixmodule.c). I see three different ways to do this.

1. Do nothing other than document the situation and wait for users to migrate to Python 3.12.
2. Warn about fork() immediately on import of tiledb.
3. Warn when `os.fork()` is called and tiledb has been imported.  

The pros for number one: only needing to reference the change coming in 3.12. The cons: this project currently supports 4 versions of Python before 3.12 and it'll be years before new users are entirely on 3.12. Also, to a first approximation, nobody reads documentation. Perhaps better said as: we often try to read as little documentation as we can get away with.

The pros for number two: super easy to implement and all tiledb users see it. The cons: users would see the warning even if they had no intention of using `fork()`. It's noisy and would have to be actively silenced.

The pros for number three: all tiledb users who call `os.fork()`, intentionally or not, get the warning and no one else. It's not noisy. The cons: it requires monkeypatching`os.fork()` or, suboptimally, using `os.register_at_fork()`.

The only problem with `os.register_at_fork()` is that exceptions and warnings can't be raised from registered callables. We're limited to printing to stderr/stdout. This is a less good experience for developers and the project also currently checks in tests that tiledb doesn't print such output.

Monkeypatching methods of the standard library isn't a perfect solution. We'd all prefer that modules we import incidentally didn't change things in the standard library, right? And there's no guarantee that a module imported after tiledb won't patch `os.fork()` again.

On the other hand, users import tiledb deliberately, to use it. And it's mostly incompatible with `fork()`. The monkeypatch I'm suggesting for `os.fork()` doesn't change behavior of the stdlib function, it only wraps it in a warning. It's as innocuous as such a patch can be.


tiledb/tests/test_multiprocessing.py
```diff
@@ -24,10 +27,80 @@ def test_warn_on_fork():
...
+                UserWarning,
+            )
+            return os_fork()
+
+        return wrapper
+
+    import tiledb  # noqa
+
+    if sys.version_info < (3, 12):
+        os.fork = wrap_fork()
```
sgillies, line 51: These lines would be moved into the tiledb package if we agree to go this route.

tiledb/tests/test_multiprocessing.py
```diff
@@ -24,10 +27,80 @@ def test_warn_on_fork():
...
+        @wraps(os_fork)
+        def wrapper():
+            warnings.warn(
+                "TileDB is a multithreading library and deadlocks are "
+                "likely if fork() is called after a TileDB array has "
+                "been created or accessed. "
+                "To safely use TileDB with multiprocessing or "
+                "concurrent.futures, choose 'spawn' as the start "
+                "method for child processes.",
+                UserWarning,
```
sgillies, line 42: Here's my first take on what a user should see printed in their terminal or notebook if they unsafely combine tiledb and fork().

sgillies:
@ihnorton can you do a preliminary review of this PR?

tiledb/tests/test_multiprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+    # Linux.
+
+    def wrap_fork():
+        os_fork = os.fork
+
+        @wraps(os_fork)
+        def wrapper():
+            warnings.warn(
+                "TileDB is a multithreading library and deadlocks are "
+                "likely if fork() is called after a TileDB array has "
```
ihnorton, line 37: nit: array -> context

tiledb/tests/test_multiprocessing.py
```diff
@@ -0,0 +1,106 @@
...
+                "been created or accessed. "
+                "To safely use TileDB with multiprocessing or "
+                "concurrent.futures, choose 'spawn' as the start "
+                "method for child processes.",
```
ihnorton, line 41: Consider adding a suggestion for how to do this? If we're adding a nice detailed message here, might be worth also adding a three-line snippet for how to do it safely.

ihnorton (commented):
Looks great, thanks for the nice detailed message. Two minor suggestions inline.

Also, it _might_ be worth running the test in a separate process for full control over the import order (maybe test both before and after some other user of `os.fork`). In another import-related test [here](https://github.com/TileDB-Inc/TileDB-Py/blob/22ff29a6a69a8c5bb4aced03f8a8b70133b3ab46/tiledb/tests/test_basic_import.py#L26-L47), we spawn a subprocess to validate the postcondition in a separate interpreter environment.

sgillies:
I overlooked a 4th approach: patching `os.fork()` not on import of tiledb, but on initialization of a `Ctx`. It keeps the solution closer to the problem and is much easier to test in isolation. A fixture like https://github.com/TileDB-Inc/TileDB-Py/pull/1876/commits/8cfe60d6c2be0ba772bf05bc31e9c91078bd76b2#diff-f31ed6f6e26b774c62db3c48ad14bc06543b7d6fc81d6e3890ea4f7722261a2fR104-R111 is enough to ensure that all tests run with the `os` module in its original pristine state.

tiledb/ctx.py
```diff
@@ -521,7 +531,46 @@ def default_ctx(config: Union["Config", dict] = None) -> "Ctx":
...
+        _ensure_os_fork_wrap()
     except LookupError:
         ctx = tiledb.Ctx(config)
         _ctx_var.set(ctx)
     return ctx
+
+
+def _ensure_os_fork_wrap():
+    global _needs_fork_wrapper
+    if _needs_fork_wrapper:
```
sgillies, line 552: Again, wrap and patch no more than once.

sgillies:
@ihnorton I've finally got the wrap/patch and fixture to isolate it into a readable and maintainable state.

ihnorton (approved):
This LGTM! Thanks for building this, and especially for the detailed comments.

PR #1877, opened 2023-12-04
Error out if query condition given empty set

nguyenv:


ckrilow (approved):
Thank you for the quick turnaround with this PR.

ihnorton (approved):

Issue #1879, opened 2023-12-14
[bug] Enumeration attribute incorrectly writes Pandas categorical column


atolopko-czi:
**Describe the bug**
A Pandas categorical column is incorrectly encoded to a TileDB Enumeration column if the Pandas categorical’s `categories` (i.e., the global set of possible values) is not isomorphic to the TileDB Enumeration’s `values`. In this case, `tiledb.from_pandas(..., mode="append")` fails to write the data properly.

**To Reproduce**
https://gist.github.com/atolopko-czi/5ca7d848d0a4155fd377444c7ecd4d14

**Versions (please complete the following information):**
 - TileDB version: 0.24.0
 - Python version: 3.11.6
 - OS: MacOS

**Additional context**
At a minimum, an error should be raised indicating the lack of agreement in the definitions of the categorical and the enumeration. Optionally, it would be nice if a mismatched Panda’s categorical was re-encoded to the the Enumeration’s encoded values, effectively ignoring the Panda’s categorical encoding.  Re-encoding behavior could enabled/disabled with a `strict_enumeration` option, potentially.

The workaround is to explicitly create Pandas Categorical columns with the same set of `categories` as the Enumeration’s `values`.

johnkerl:
Needs triaging for any analgous R behavior

eddelbuettel:
Fix is in, just need to add some more tests.

```sh
edd@rob:~/git/tiledb-r(de/sc-38457/relevel_factor)$ ../tiledb-adhoc/bugreports/issue1879py.R
    dim1  attr1
   <int> <fctr>
1:     0      b
2:     1      c
    dim1  attr1
   <int> <fctr>
1:     0      b
2:     1      c
edd@rob:~/git/tiledb-r(de/sc-38457/relevel_factor)$ 
```

Before the fix the second write would (as in the initial Python repro) write 'a', 'b' in error.

eddelbuettel:
- In TileDB-SOMA that is commit c404e90c3 from Dec 16 and this line https://github.com/single-cell-data/TileDB-SOMA/blob/37298c63e65015c899a11492d42a7a7bdc7ed676/apis/r/R/SOMADataFrame.R#L192
- In TileDB-R that is commit 82c4abd2c from Dec 15 and this line https://github.com/TileDB-Inc/TileDB-R/blob/a5bde207b213674bf736f129698666b800abe539/R/TileDBArray.R#L1427 (plus the two before)
[...]

PR #1881, opened 2023-12-15
Use enum in `ArraySchema` for to-be-written Pandas category

nguyenv:
Fixes issue #1879 for Python

As reported by @atolopko-czi

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -693,8 +693,7 @@ def _write_array(
     row_start_idx=None,
     timestamp=None,
 ):
-    import pandas as pd 
-    
+
     with tiledb.open(uri, "w", timestamp=timestamp) as A:
         for j in range(A.schema.nattr):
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
🚢

tiledb/dataframe_.py
```diff
@@ -693,7 +693,15 @@ def _write_array(
     row_start_idx=None,
     timestamp=None,
 ):
+
     with tiledb.open(uri, "w", timestamp=timestamp) as A:
+        for j in range(A.schema.nattr):
+            attr = A.schema.attr(j)
+            if attr.enum_label is not None:
+                enmr = A.enum(attr.enum_label).values()
+                df[attr.name] = df[attr.name].cat.set_categories(enmr)
```
ihnorton, line 702: We should probably do this at the arrow level eventually, this is going to be expensive. (sc-38461)

ihnorton (approved):

Issue #1880, opened 2023-12-15
[feature request] Query conditions have unexpected behavior with enum attributes


bkmartinjr:
I am in process of converting a bunch of `string` attributes to use the new enum capability.  I assumed the filter/query condition equality (`==` and `in [...]`) behavior would match that of the original string columns, but there is a significant change in behavior.

When a query condition on a string attribute fails to match, it returns an empty result.  When a query condition is applied to a enum column, it generates an exception if the filter tries to test for equality with a string not present in the enum value list.

This effectively forces query condition users to know ahead of time what enum values are present, even if the end result of the operation is the same (an empty result).

At least in the Pandas ecosystem, equality operations on categoricals act as if they are operating on the value, e.g., an `==` operation is supported against a string, regardless if the dataframe column is a categorical-of-string or just plain string.  String ordered comparison operations, such as `<`, are not supported in this manner for unoredered categories - just equality and related ops (e.g., `in`).

What I'd like to advocate for:
* for unordered enums:  support equality-like ops vs any string (e.g,   "enum_col == 'foobar'"), returning an empty result if the string value is not in the defined enum values.
* for ordered enums: also support lt, gt, etc. in the filters

Example of the unordered case:

```
In [16]: obs.schema
Out[16]: 
ArraySchema(
  domain=Domain(*[
    Dim(name='soma_joinid', domain=(0, 2147483646), tile=2048, dtype='int64', filters=FilterList([DoubleDeltaFilter(reinterp_dtype=None), ZstdFilter(level=19), ])),
  ]),
  attrs=[
    Attr(name='assay', dtype='int8', var=False, nullable=False, enum_label='assay', filters=FilterList([ZstdFilter(level=19), ])),
    Attr(name='observation_joinid', dtype='<U0', var=True, nullable=False, enum_label=None, filters=FilterList([ZstdFilter(level=19), ])),
  ],
  cell_order='row-major',
  tile_order='row-major',
  capacity=65536,
  sparse=True,
  allows_duplicates=True,
)

In [17]: obs.enum('assay')
Out[17]: Enumeration(name='assay', cell_val_num=4294967295, ordered=False, values=["10x 3' v2", "10x 3' v3", "10x 5' v1", 'BD Rhapsody Targeted mRNA', 'Smart-seq2'])

In [18]: obs.query(cond="assay == 'foobar'").df[:]
---------------------------------------------------------------------------
TileDBError                               Traceback (most recent call last)
[...]

PR #1882, opened 2023-12-15
[2.21.0] Add tests to ensure empty result on query condition for invalid enum

nguyenv:
https://github.com/TileDB-Inc/TileDB-Py/issues/1880

This is now fixed in core. This PR just adds unit tests to ensure query conditions return an empty result, instead of erroring out, for invalid enumeration values.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ from setuptools import Extension, find_packages, setup
 # - this is for builds-from-source
 # - release builds are controlled by `misc/azure-release.yml`
 # - this should be set to the current core release, not `dev`
-TILEDB_VERSION = "2.21.0-rc1"
+TILEDB_VERSION = "2.21.0"
 
 # allow overriding w/ environment variable
 TILEDB_VERSION = (
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ from setuptools import Extension, find_packages, setup
 # - this is for builds-from-source
 # - release builds are controlled by `misc/azure-release.yml`
 # - this should be set to the current core release, not `dev`
-TILEDB_VERSION = "2.20.1"
+TILEDB_VERSION = "2.21.0-rc1"
 
 # allow overriding w/ environment variable
 TILEDB_VERSION = (
--- a/tiledb/tests/test_query_condition.py
+++ b/tiledb/tests/test_query_condition.py
@@ -858,6 +858,15 @@ class QueryConditionTest(DiskTestCase):
             result = A.query(cond="attr1 < 2", attrs=["attr1"])[:]
             assert all(self.filter_dense(result["attr1"], mask) < 2)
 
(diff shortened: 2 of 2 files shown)


nguyenv:
This is also needs to be ported over to TileDB-SOMA-Py.

ihnorton (changes requested):
I don't think this is the right way to solve this issue. Let's see if we can address this in libtiledb.

nguyenv:
[sc-43002]

After discussing with @kounelisagis, we are just going to use this pre-existing PR to add the unit tests instead of opening a new one.

kounelisagis:
Error comes from libtiledb. Should we address it there?

`FAILED tiledb/tests/test_query_condition.py::QueryConditionTest::test_qc_enumeration - tiledb.cc.TileDBError: TileDB internal: Enumeration value not found for field 'attr2'`

nguyenv:
We need to wait for the libtiledb 2.21.0 release for this fix and then we can update `setup.py`. Although in the meantime we can test this against 2.21.0-rc2 to make sure CI is all passing.

setup.py
```diff
@@ -20,7 +20,7 @@
 # - this is for builds-from-source
 # - release builds are controlled by `misc/azure-release.yml`
 # - this should be set to the current core release, not `dev`
-TILEDB_VERSION = "2.20.1"
+TILEDB_VERSION = "2.21.0-rc1"
```
nguyenv, line 23: Reminder to revert before merging.

nguyenv:
@kounelisagis I'm so sorry, when I merged main into this branch I accidentally removed your changes. Can you push your changes again.

ihnorton (commented):
LGTM but please double-check that these all _fail_ without 2.21.

kounelisagis:
It fails without 2.21 (or at least with 2.20...)

---

```
git checkout viviannguyen/sc-38443/dataframe-value-filter-on-enum-dict-column
pip install -r requirements_dev.txt
python setup.py develop
python -c "import tiledb; print(tiledb.libtiledb.version())"
```
**(2, 21, 0)**
```
pytest -o log_cli=true --log-cli-level=10 tiledb/tests/test_query_condition.py
```
Passes

---

```
git checkout dev setup.py
python setup.py develop
python -c "import tiledb; print(tiledb.libtiledb.version())"
```
**(2, 20, 1)**

```
pytest -o log_cli=true --log-cli-level=10 tiledb/tests/test_query_condition.py
```
Gives the following error:
https://gist.github.com/kounelisagis/7c9d8803b24f92335b2c71d58bc7f9c8

kounelisagis:
Also tested it with 2.21.0 (not pre-release) and it works.

kounelisagis (approved):

nguyenv:
Thanks for doing the work of double checking @kounelisagis.

nguyenv:
And also for reverting the libtiledb version in setup.py.

PR #1886, opened 2024-01-10
Release 0.25.1 Against 2.19.0

nguyenv:
* 0.25.0 will be need to be deprecated as the conda package depends on the wrong version of libtiledb

```
(tiledb-clean) vivian@mangonada:~/drop$ python
Python 3.10.13 (main, Sep 11 2023, 13:44:35) [GCC 11.2.0] on linux
>>> import tiledb
>>> tiledb.version()
(0, 25, 1)
>>> tiledb.libtiledb.version()
(2, 19, 0)
```

- [ ] Upload new PyPI wheels
- [ ] Update libtiledb dependency in `meta.yaml`
- [ ] Mark broken conda package: https://github.com/nguyenv/admin-requests/compare/main...mark-tiledb-py-0.25.0-as-broken
- [ ] Yank on PyPI

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,12 +1,14 @@
-# Next
+# Release 0.25.1
+
+* TileDB-Py 0.25.1 includes TileDB Embedded [2.19.0](https://github.com/TileDB-Inc/TileDB/releases/tag/2.19.0)
 
 ## Improvements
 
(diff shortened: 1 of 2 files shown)


jdblischak (commented):
I don't think this PR is necessary. We can fix the conda problem in the feedstock repo

PR #1887, opened 2024-01-11
Fix ambiguous capsule calls with pybind11

robertbindar:
I got multiple errors like this:
```error: call of overloaded ‘capsule(std::__shared_ptr<tiledb_subarray_t, __gnu_cxx::_S_atomic>::element_type*, const char [9], std::nullptr_t)’ is ambiguous```
when building fresh with latest pybind on ubuntu22, `m6id.4xlarge` ec2 machine.

Not sure this is the correct fix, just fixed all the errors to get myself a successful build and thought effort shouldn't go to waste in case it's the correct approach.

### DIFF
--- a/tiledb/cc/attribute.cc
+++ b/tiledb/cc/attribute.cc
@@ -69,7 +69,7 @@ void init_attribute(py::module &m) {
 
       .def("__capsule__",
            [](Attribute &attr) {
-             return py::capsule(attr.ptr().get(), "attr", nullptr);
+             return py::capsule(attr.ptr().get(), "attr");
            })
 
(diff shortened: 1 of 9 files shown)


nguyenv:
This looks correct to me, but I swear we already made these changes before to fix these errors. Maybe it was in different files, or maybe even in the TileDB-SOMA Pybind11 code. I want to check again tomorrow morning. For now, can you please run `pre-commit run -a -v` to correct the formatting?

robertbindar:
@nguyenv I saw a fix related in a SOMA pybind11 PR, I'll uldate formatting in a bit, thanks.

ihnorton (approved):

PR #1890, opened 2024-01-18
Safe repr for `Group`

nguyenv:
See https://github.com/TileDB-Inc/TileDB-Py/pull/1545/ for a full explanation. This just needed to be added for the `Group` class as well.


jparismorgan (approved):
Thanks for fixing this! I would wait for another approve though, as I don't know this code well.

PR #1902, opened 2024-02-26
Add method aliases to tiledb.VFS

sgillies:
Three aliases in `VFS`, three tests. Not making any deprecations or changes to the original methods. We could add a note that in the docstrings that aliases exist, maybe.


tiledb/tests/test_vfs.py
```diff
@@ -287,3 +287,26 @@ def test_open_with(self):
...
+    assert fs.isfile(tmp_file.as_posix())
+
+
+def test_vfs_size(tmp_path):
+    """size is an alias for file_size."""
+    tmp_file = tmp_path.joinpath("foo")
+    buffer = b"0123456789"
+    tmp_file.write_bytes(buffer)
+    fs = tiledb.VFS()
+    assert fs.size(tmp_file.as_posix()) == len(buffer)
```
sgillies, line 312: I'm TDD'ing. 3 failing tests, all using pytest and its `tmp_path` fixture instead of creating resources using the package's API.

tiledb/vfs.py
```diff
@@ -304,6 +304,11 @@ def touch(self, uri: _AnyPath):
         """
         return self._touch(_to_path_str(uri))
 
+    # Aliases for compatibility with fsspec's AbstractFileSystem.
+    isdir = is_dir
+    isfile = is_file
+    size = file_size
```
sgillies, line 310: That's all it takes to make the tests green. If you `help(isdir)` you get the docstring from `is_dir`, but that's not terrible.

sgillies:
The failing test looks unrelated.

nguyenv:
> The failing test looks unrelated.

Yeah this test is known to intermittenly fail. We should be passing in an explicit datetime format into this converter.

nguyenv:
https://www.notion.so/Deprecation-Cycles-in-TileDB-Py-96924f154dc34910a4c61a59b596749a

I think that we could start deprecating the old names and transition into these new ones. In the warning message we would note that we are doing this to comply with the fsspec.

sgillies:
@nguyenv I like the version assertion check 👍 So, do you suggest making, for example, `is_dir()` a deprecated alias of the new `isdir()`?

nguyenv:
I should have mentioned - I talked to Isaiah about this yesterday. For now don't worry about deprecating anything. We use `VFS` in so many things that it'll be more work than it's worth to change things over right now.

PR #1904, opened 2024-02-27
Add mem:// backend example

kounelisagis:
Add an example of how to use new in-memory URI base **mem://** for TileDB VFS operations, provided by backend.

### DIFF
--- a/examples/in_memory_vfs.py
+++ b/examples/in_memory_vfs.py
@@ -1,0 +1,45 @@
+# in_memory_vfs.py
+#
+# LICENSE
+#
+# The MIT License
+#
+# Copyright (c) 2024 TileDB, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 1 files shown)


examples/in_memory_vfs.py
```diff
@@ -0,0 +1,45 @@
...
+#
+
+import numpy as np
+
+import tiledb
+
+data = np.random.rand(10, 10)
+
+with tiledb.from_numpy("mem://myarray", data):
+    pass
```
nguyenv, line 42: I think you'll want to check `VFS.is_file(uri)` before creating the array. Otherwise, you'll error out with a message saying that there's already something at the given URI. Another option would be to use the `tempfile` library to create a temporary directory that cleans itself up after running this example.

kounelisagis: don't you believe it would be better to encapsulate everything in a try...catch?

nguyenv: You could do that as well

kounelisagis: I was following similar examples like query_condition_datetime.py:
- array creation: https://github.com/TileDB-Inc/TileDB-Py/blob/c0e243eb12a0883f1b86bba65ff3767276068c36/examples/query_condition_datetime.py#L48
- array opening: https://github.com/TileDB-Inc/TileDB-Py/blob/c0e243eb12a0883f1b86bba65ff3767276068c36/examples/query_condition_datetime.py#L54

In any case `with` statement handles the case when there is already an array at the given URI. What do you think?

nguyenv: Those examples should be modified as well. I think some of the examples do this check and the others do not, but we should start being more careful about ensuring we don't error out creating example arrays that already exist. 

Do you mean `with tiledb.from_numpy` should handle the case of an already existing object?

```
>>> with tiledb.from_numpy("mem://myarray", data):
...     pass
...
>>> with tiledb.from_numpy("mem://myarray", data):
...     pass
...
  File "/home/vivian/TileDB-Py/tiledb/highlevel.py", line 125, in from_numpy
  File "tiledb/libtiledb.pyx", line 985, in tiledb.libtiledb.Array.create
  File "tiledb/libtiledb.pyx", line 351, in tiledb.libtiledb._raise_ctx_err
  File "tiledb/libtiledb.pyx", line 336, in tiledb.libtiledb._raise_tiledb_error
tiledb.cc.TileDBError: [TileDB::StorageManager] Error: Cannot create array; Array 'mem://myarray' already exists
```

kounelisagis: I see. In case there is already an array, should the program continue with this array or exit?

nguyenv: Yes it should

nguyenv (approved):
Thanks!

ihnorton (approved):

PR #1905, opened 2024-03-01
Specify datetime format in tests

kounelisagis:
The `tiledb.from_csv` and `tiledb.from_pandas` functions have been causing intermittent errors when running the unit tests with the message:

```
UserWarning: Could not infer format, so each element will be parsed individually, falling back to `dateutil`. To ensure parsing is consistent and as-expected, please specify a format.
```
Example failure: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/7911328196/job/21595351207#step:10:1029

This is solved by explicitly passing `date_format` argument into each `from_{csv,pandas}` call.

Please note that the `date_format` argument is not available for Python 3.7 and below, so tests containing it should be skipped.

### DIFF
--- a/tiledb/tests/test_pandas_dataframe.py
+++ b/tiledb/tests/test_pandas_dataframe.py
@@ -3,6 +3,7 @@ import glob
 import os
 import random
 import string
+import sys
 import uuid
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


ihnorton (approved):
LGTM, thanks. I just restarted the CI job, just to see if we can get another clean run. Merge when you want but please check the job when it finishes.

ihnorton:
I'm going to go ahead and merge, and then rebase 1906 on top of this.

PR #1906, opened 2024-03-01
Use Ninja in Windows CI and explicitly set the VS toolset version.

teo-tsirpanis:
Use Ninja in Windows CI and explicitly set the VS toolset version.


teo-tsirpanis:
[Nightly builds succeed on Windows.](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/8111146558)

Some other jobs fail intermittently but all succeed at least once.

ihnorton (approved):

PR #1907, opened 2024-03-01
Fix Filestore unable to read sizes smaller than offset

kounelisagis:
In `Filestore` class, in the `read()` method this line `size = max(size - offset, 0)` does not cover the case of trying to read a size that is smaller than the offset. Consider a 4 byte file. If we want to read the last byte (offset 3, size 1) this check sets the size to 0 incorrectly.

Changing it to `size = min(size, len(self) - offset)` ensures that we read either the specified size or the remaining length of the file store starting from the offset, whichever is smaller.

### DIFF
--- a/tiledb/filestore.py
+++ b/tiledb/filestore.py
@@ -67,7 +67,7 @@ class Filestore:
 
         if size == -1:
             size = len(self)
-        size = max(size - offset, 0)
+        size = min(size, len(self) - offset)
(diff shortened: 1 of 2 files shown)


ihnorton (changes requested):
Is it possible to add a test?

ihnorton (approved):
Thanks!

PR #1910, opened 2024-03-04
Update daily test builds to use single source of truth for libtiledb target versions

kounelisagis:
Currently, we have two different daily ("nightly") build jobs run on a schedule:

1. https://github.com/TileDB-Inc/TileDB-Py/blob/dev/.github/workflows/daily-test-build.yml
2. https://github.com/TileDB-Inc/TileDB-Py/blob/dev/.github/workflows/daily-test-build-numpy.yml

These jobs have different specifications for where they get the libtiledb version:
- (1) specifies a [version matrix](https://github.com/TileDB-Inc/TileDB-Py/blob/dev/.github/workflows/daily-test-build.yml#L38)
- (2) uses the [version in setup.py](https://github.com/TileDB-Inc/TileDB-Py/blob/dev/setup.py#L23)

We would like to change this to use github actions `workflow_call` from a single list of versions.

See the following examples in libtiledb repo:
- https://github.com/TileDB-Inc/TileDB/blob/dev/.github/workflows/full-ci.yml#L49-L54
- https://github.com/TileDB-Inc/TileDB/blob/dev/.github/workflows/ci-linux_mac.yml#L3-L9


jdblischak (approved):

dudoslav (approved):
Looks good to me, did you have a chance to test this on your local fork for example?

kounelisagis:
@dudoslav I run it manually, as I wrote in the description of the PR.

PR #1912, opened 2024-03-05
Add support for creating WKB/WKT attributes

jp-dark:
Features:
* Add WKB/WKT as possible attribute types

Fixes:
* Fix getting fill value for BLOB, WKB, and WKT types

---

- closes https://github.com/TileDB-Inc/TileDB-Py/pull/1912
- [sc-41135]

### DIFF
--- a/tiledb/cc/attribute.cc
+++ b/tiledb/cc/attribute.cc
@@ -19,28 +19,42 @@ void set_fill_value(Attribute &attr, py::array value) {
 }
 
 py::array get_fill_value(Attribute &attr) {
+  // Get the fill value from the C++ API as a void* value.
   const void *value;
   uint64_t size;
-
   attr.get_fill_value(&value, &size);
 
+  // If this is a string type, we want to return each value as a single cell.
+  if (is_tdb_str(attr.type())) {
+    auto value_type = py::dtype("|S1");
+    return py::array(value_type, size, value);
+  }
+
+  // If this is a record type (void), return a single cell.
+  // If this is a blob-like type, we want to return each value as a single byte
+  // cell.
+  auto tdb_type = attr.type();
+  if (tdb_type == TILEDB_BLOB
+#if TILEDB_VERSION_MAJOR >= 2 && TILEDB_VERSION_MINOR >= 21
+      || tdb_type == TILEDB_GEOM_WKB || tdb_type == TILEDB_GEOM_WKT
+#endif
+  ) {
+    auto value_type = py::dtype("S");
+    return py::array(value_type, size, value);
+  }
+
+  // Get the number of values in a cell and the Python datatype.
   auto value_num = attr.cell_val_num();
   auto value_type = tdb_to_np_dtype(attr.type(), value_num);
 
-  if (is_tdb_str(attr.type())) {
-    value_type = py::dtype("|S1");
-    value_num = size;
-  }
-
-  // record type
   if (py::str(value_type.attr("kind")) == py::str("V")) {
-    value_num = 1;
+    return py::array(value_type, 1, value);
   }
 
-  // complex type - both cell values fit in a single complex element
+  // If this is a complex type both cell values fit in a single complex element.
   if (value_type == py::dtype("complex64") ||
       value_type == py::dtype("complex128")) {
-    value_num = 1;
+    return py::array(value_type, 1, value);
   }
 
   return py::array(value_type, value_num, value);
--- a/tiledb/datatypes.py
+++ b/tiledb/datatypes.py
@@ -24,6 +24,13 @@ class DataType:
         if dtype == "blob":
             return cls(np.dtype("S"), lt.DataType.BLOB, 1)
 
+        if hasattr(lt.DataType, "GEOM_WKB"):
+            if dtype == "wkb":
+                return cls(np.dtype("S"), lt.DataType.GEOM_WKB, 1)
+
+            if dtype == "wkt":
+                return cls(np.dtype("S"), lt.DataType.GEOM_WKT, 1)
+
         dtype = np.dtype(dtype)
         if dtype.kind == "V":
             # fixed-size record dtypes
@@ -179,10 +186,13 @@ _NUMPY_TO_TILEDB = {n: t for n, t in _COMMON_DATATYPES}
 _NUMPY_TO_TILEDB[np.dtype("complex64")] = lt.DataType.FLOAT32
 _NUMPY_TO_TILEDB[np.dtype("complex128")] = lt.DataType.FLOAT64
 
-# tiledb has STRING_ASCII and BLOB, numpy doesn't
+# tiledb has STRING_ASCII, BLOB, WKB and WKT types, numpy doesn't
 _TILEDB_TO_NUMPY = {t: n for n, t in _COMMON_DATATYPES}
 _TILEDB_TO_NUMPY[lt.DataType.STRING_ASCII] = np.dtype("S")
 _TILEDB_TO_NUMPY[lt.DataType.BLOB] = np.dtype("S")
+if hasattr(lt.DataType, "GEOM_WKB"):
+    _TILEDB_TO_NUMPY[lt.DataType.GEOM_WKB] = np.dtype("S")
+    _TILEDB_TO_NUMPY[lt.DataType.GEOM_WKT] = np.dtype("S")
 
 # pre-populate the LRU caches with all ncell=1 datatypes
 list(map(DataType.from_numpy, _NUMPY_TO_TILEDB.keys()))
--- a/tiledb/datatypes.py
+++ b/tiledb/datatypes.py
@@ -28,6 +28,7 @@ class DataType:
             if dtype == "wkb":
                 return cls(np.dtype("S"), lt.DataType.GEOM_WKB, 1)
 
+        if hasattr(lt.DataType, "GEOM_WKT"):
             if dtype == "wkt":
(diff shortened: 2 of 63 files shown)


tiledb/datatypes.py
```diff
@@ -24,6 +24,13 @@ def from_numpy(cls, dtype: np.dtype) -> DataType:
         if dtype == "blob":
             return cls(np.dtype("S"), lt.DataType.BLOB, 1)
 
+        if hasattr(lt.DataType, "GEOM_WKB"):
+            if dtype == "wkb":
+                return cls(np.dtype("S"), lt.DataType.GEOM_WKB, 1)
+
+            if dtype == "wkt":
+                return cls(np.dtype("S"), lt.DataType.GEOM_WKT, 1)
+
```
perrygeo, line 33: Should we have a separate branch for WKT? 

```suggestion
        if hasattr(lt.DataType, "GEOM_WKB"):
            if dtype == "wkb":
                return cls(np.dtype("S"), lt.DataType.GEOM_WKB, 1)

        if hasattr(lt.DataType, "GEOM_WKT"):
            if dtype == "wkt":
                return cls(np.dtype("S"), lt.DataType.GEOM_WKT, 1)

```

perrygeo:
I've been testing locally and this works great. This unblocks a whole bunch of testing I need to do! Thanks @jp-dark 

 I'm using the code below to create a geometry array compatible with GDAL/OGR [metadata conventions](https://gdal.org/drivers/vector/tiledb.html#supported-datasets). This is probably too application-specific for this repo's test suite but I'll leave it here for reference.

```python
import os

import numpy as np
from shapely import box

import tiledb

path = "/tmp/geom_array"
schema = tiledb.ArraySchema(
    domain=tiledb.Domain(
        # Globe in Longitude, Latitude coordinates, 1 degree square tiles
        tiledb.Dim(name="_X", domain=(-180.0, 180.0), tile=1.0, dtype=np.float64),
        tiledb.Dim(name="_Y", domain=(-90.0, 90.0), tile=1.0, dtype=np.float64),
    ),
    attrs=[
        tiledb.Attr(name="fid", dtype=np.int64),
        # GDAL currently only supports blob
        # tiledb.Attr(name="wkb_geometry", dtype="blob"),
        tiledb.Attr(name="wkb_geometry", dtype="wkb"),
    ],
    allows_duplicates=True,
    cell_order="row-major",
    tile_order="row-major",
    capacity=100_000,
    sparse=True,
)

geoms = [
    box(2, 2, 4, 4),
    box(1, 1, 3, 3),
    box(0, 0, 2, 2),
]
wkbs = [g.wkb for g in geoms]
bboxes = [g.bounds for g in geoms]
fids = np.array([i for i, _ in enumerate(geoms)])
n = len(geoms)

# use the centroid of the bounding box as dimensions
xs = [(b[0] + b[2]) / 2 for b in bboxes]
ys = [(b[1] + b[3]) / 2 for b in bboxes]

# use half the max dimensions for padding
max_width = max(b[2] - b[0] for b in bboxes)
max_height = max(b[3] - b[1] for b in bboxes)
padx = max_width / 2
pady = max_height / 2

tiledb.Array.create(path, schema)

# Write OGR Metadata
with tiledb.open(path, "w") as arr:
    # arr.meta["CRS"] = CRS
    arr.meta["GeometryType"] = "Polygon"
    arr.meta["GEOMETRY_ATTRIBUTE_NAME"] = "wkb_geometry"
    arr.meta["LAYER_EXTENT_MINX"] = -180.0
    arr.meta["LAYER_EXTENT_MINY"] = -90.0
    arr.meta["LAYER_EXTENT_MAXX"] = 180.0
    arr.meta["LAYER_EXTENT_MAXY"] = 90.0
    arr.meta["PAD_X"] = padx
    arr.meta["PAD_Y"] = pady
    arr.meta["FEATURE_COUNT"] = n

# Write attributes themselves
with tiledb.open(path, mode="w") as arr:
    data = {
        "fid": fids,
        "wkb_geometry": np.array(
            wkbs,
            dtype="O",
        ),
    }
    arr[xs, ys] = data
```
Using the `blob` type results in a valid geometry array according to `ogrinfo`. Using the `wkb` type does not, yet.

```
$ ogrinfo -al /tmp/geom_array
INFO: Open of `/tmp/geom_array'
      using driver `TileDB' successful.
Layer name: geom_array
Geometry: Polygon
Feature Count: 3
Extent: (-180.000000, -90.000000) - (180.000000, 90.000000)
fid: Integer64 (0.0) NOT NULL
wkb_geometry: String (0.0) NOT NULL
ERROR 1: std::get: wrong index for variant
```

I'll start looking into the GDAL code to see what it will take to support reading/writing the WKB_GEOMETRY type instead of BLOB - It should be doable in a backwards compatible way.

jp-dark:
Note: This PR requires an update to how WKT is handled (should support strings -> binary blob) and tests for round tripping data for both WKT and WKB.

tiledb/cc/attribute.cc
```diff
@@ -19,28 +19,42 @@ void set_fill_value(Attribute &attr, py::array value) {
...
+    return py::array(value_type, size, value);
   }
 
-  // record type
+  // If this is a record type (void), return a single cell.
+  // If this is a blob-like type, we want to return each value as a single byte
+  // cell.
+  auto tdb_type = attr.type();
+  if (tdb_type == TILEDB_BLOB
+#if TILEDB_VERSION_MAJOR >= 2 && TILEDB_VERSION_MINOR >= 21
```
rouault, line 38: if there's a tiledb 3 some day:
```suggestion
#if TILEDB_VERSION_MAJOR > 2 || (TILEDB_VERSION_MAJOR == 2 && TILEDB_VERSION_MINOR >= 21)
```

tiledb/py_arrow_io_impl.h
```diff
@@ -199,6 +199,10 @@ ArrowInfo tiledb_buffer_arrow_fmt(BufferInfo bufferinfo, bool use_list = true) {
   case TILEDB_FLOAT64:
     return ArrowInfo("g");
   case TILEDB_BLOB:
+#if TILEDB_VERSION_MAJOR >= 2 && TILEDB_VERSION_MINOR >= 21
```
rouault, line 202: if there's a tiledb 3 some day:
```suggestion
#if TILEDB_VERSION_MAJOR > 2 || (TILEDB_VERSION_MAJOR == 2 && TILEDB_VERSION_MINOR >= 21)
```

kounelisagis (changes requested):

tiledb/py_arrow_io_impl.h
kounelisagis: Nice observation. It's acceptable for now, but it will change throughout the entire project in the future.

ihnorton (approved):
This LGTM, but I'd like to get a ✅ from @jp-dark (time-permitting) before we merge. In particular, is there anything we should re-test [here](https://github.com/TileDB-Inc/TileDB-Py/pull/1912#issuecomment-2004618669), and do you know if we need to build ogrinfo from source to be able to repro that?

tiledb/tests/test_attribute.py
```diff
@@ -252,3 +252,52 @@ def test_modify_attribute_in_schema(self):
...
+        schema = tiledb.ArraySchema(dom, (att,))
+
+        tiledb.DenseArray.create(self.path("foo"), schema)
+        with tiledb.DenseArray(self.path("foo"), mode="w") as T:
+            T[:] = A
+
+        # read back the data
+        with tiledb.DenseArray(self.path("foo"), mode="r") as T:
+            for i in range(2):
+                assert_array_equal(T[:][i].tobytes(), A[i])
```
jp-dark, line 274: I think from the Python API, I would expect "WKT" to behave more like a string type than a blob type. For example, I would expect back the same thing we get from an ascii datatype:
```
array([b'POINT (30 10)', b'POLYGON ((3 1, 4 5, 2 2, 1 2, 3 1))'],
      dtype=object)
```
instead of 
```
array([array([b'P', b'O', b'I', b'N', b'T', b' ', b'(', b'3', b'0', b' ', b'1',
              b'0', b')'], dtype='|S1')                                        ,
       array([b'P', b'O', b'L', b'Y', b'G', b'O', b'N', b' ', b'(', b'(', b'3',
              b' ', b'1', b',', b' ', b'4', b' ', b'5', b',', b' ', b'2', b' ',
              b'2', b',', b' ', b'1', b' ', b'2', b',', b' ', b'3', b' ', b'1',
              b')', b')'], dtype='|S1')                                        ],
      dtype=object)
```

That being said, it might be worth asking the geospatial team to confirm this is the behavior they would expect as well.

jp-dark:
> In particular, is there anything we should re-test [here](https://github.com/TileDB-Inc/TileDB-Py/pull/1912#issuecomment-2004618669), and do you know if we need to build ogrinfo from source to be able to repro that?

Those test were all @perrygeo. I haven't been directly involved in any of the GDAL support or testing. Maybe @rouault can answer about the orginfo changes?

tiledb/tests/test_attribute.py
kounelisagis: Solved.

---
- **np.dtype("S")** represents a string of fixed length but without a specified length. Length of the string is determined by the data provided.
- **np.dtype("S1")** This represents a string of fixed length where the length is explicitly set to 1 byte.

jp-dark (commented):
lgtm

rouault (approved):

ihnorton (approved):

kounelisagis (approved):

PR #1921, opened 2024-03-14
Minor fix docs

kounelisagis:
Fixes readinto() description for the docs and moves more metadata into pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,9 @@ dependencies = {file = "requirements_wheel.txt"}
 homepage = "https://github.com/TileDB-Inc/TileDB-Py"
 
 [tool.setuptools_scm]
+version_scheme = "guess-next-dev"
+local_scheme = "dirty-tag"
+version_file = "tiledb/_generated_version.py"
 
 [tool.pytest.ini_options]
 python_classes = "*Test*"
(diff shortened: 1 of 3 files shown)


dudoslav (commented):
I am getting lost in what file is used for building this repo. Is it `setup.py` or `pyproject.toml`? Or both and it is then combined? What would happen if `setup.py` would define the same variables as `pyproject.toml`? like `version_scheme` for example.

kounelisagis:
I'm trying to move everything to .toml. Some things depend on Python functions, so it's more complex.

---
"While it is not technically necessary yet, it is STRONGLY RECOMMENDED for a project to have a pyproject.toml file at the root of its source tree"
"One way to look at it is that the scope of setuptools has now been reduced to the role of a build backend."
https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#modernize-setup-py-project

dudoslav (approved):

ihnorton (commented):
Doc changes are fine to merge with ✅ from anyone. Just make sure to put build system change in the title too in order to get attention on the scope of the change - ie I would suggest titling this PR something like "pyproject metadata updates, and minor docs fix".

PR #1922, opened 2024-03-15
Fix missing dev requirements

kounelisagis:
Requirements such as `pandas` are missing when trying to run `pytest`, even after running `pip install -r requirements_dev.txt`, leading to skipped tests. This fixes it.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -13,3 +13,4 @@ wheel >= 0.30
 contextvars ;python_version<"3.7"
 dataclasses ;python_version<"3.7"
 pytest
+-r misc/requirements_test.txt


ihnorton (approved):

nguyenv:
Which tests are failing with Pandas uninstalled? We should be checking if the unit test requires Pandas (or Pyarrow) and skipping it if it is not installed. Here is an example: https://github.com/TileDB-Inc/TileDB-Py/blob/dev/tiledb/tests/test_query_condition.py#L597

nguyenv:
Well actually now I'm not sure if we should be doing what I said or just use `requirements_test.txt`...

ihnorton:
The tests are skipped correctly, but anyone who is developing needs sufficient dependencies to run full tests. We handle the "no failures if pandas is missing" case by [manually uninstalling](https://github.com/TileDB-Inc/TileDB-Py/blob/5cfefdd0d72a90fa4e16fb2f1107cb09a86b2958/.github/workflows/daily-test-build-numpy.yml#L90).

ihnorton (approved):

nguyenv:
Oh ok I see.

PR #1933, opened 2024-03-22
Add wrapping for ls_recursive

kounelisagis:
- Add pybind11 wrapping for ls_recursive
- Connect to the high-level VFS.ls API as a keyword argument
- Test

### DIFF
--- a/tiledb/cc/vfs.cc
+++ b/tiledb/cc/vfs.cc
@@ -43,7 +43,25 @@ void init_vfs(py::module &m) {
       .def("_copy_file", &VFS::copy_file)
 
       .def("_ls", &VFS::ls)
-      .def_static("_ls_recursive", &VFSExperimental::ls_recursive)
+      .def("_ls_recursive",
+        // 1. the user provides a callback, the user callback is passed to the C++ function. Return an empty vector.
+        // 2. no callback is passed and a default callback is used. The default callback just appends each path gathered. Return the paths.
+        [](VFS& vfs, std::string uri, py::object callback) {
+          tiledb::Context ctx;
+          if (callback.is_none()) {
+            std::vector<std::string> paths;
+            tiledb::VFSExperimental::ls_recursive(ctx, vfs, uri, [&](const std::string_view path, uint64_t object_size) -> bool {
+              paths.push_back(std::string(path));
+              return true;
+            });
+            return paths;
+          } else {
+            tiledb::VFSExperimental::ls_recursive(ctx, vfs, uri, [&](const std::string_view path, uint64_t object_size) -> bool {
+              return callback(path, object_size).cast<bool>();
+          });
+          return std::vector<std::string>();
+        }
+      })
       .def("_touch", &VFS::touch);
 }
 
--- a/tiledb/cc/vfs.cc
+++ b/tiledb/cc/vfs.cc
@@ -1,11 +1,11 @@
 #include <tiledb/tiledb>
 #include <tiledb/tiledb_experimental>
 
+#include <pybind11/functional.h>
 #include <pybind11/numpy.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/pytypes.h>
 #include <pybind11/stl.h>
-#include <pybind11/functional.h>
 
 #include "common.h"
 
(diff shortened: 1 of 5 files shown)


kounelisagis:
Works when using libtiledb built from dev branch.
Requires this to be released: https://github.com/TileDB-Inc/TileDB/pull/4820

nguyenv:
As discussed in the meeting, lets test temporily change the `LIBTILEDB_VERSION` in `setup.py` to `release-2.21` to make sure this passes in CI, so by the time this gets released in core, we can just merge it without any potential issues. I also recommend making a note to revert the version to `2.21.2` before merging.

tiledb/tests/test_vfs.py
```diff
@@ -291,6 +291,72 @@ def test_ls(self):
             ),
         )
 
+    @pytest.mark.skipif(pytest.tiledb_vfs != "s3", reason="S3 specific test")
+    def test_ls_recursive(self):
+        basepath = self.path("test_vfs_ls_recursive")
+        self.vfs.create_dir(basepath)
+        # Create a nested directory structure to test recursive listing
+        for id in (1, 2, 3):
+            dir = os.path.join(basepath, "dir" + str(id))
```
nguyenv, line 300: nit
```suggestion
            dir = os.path.join(basepath, f"dir{id}")
```

tiledb/tests/test_vfs.py
```diff
@@ -291,6 +291,72 @@ def test_ls(self):
...
+            self.vfs.create_dir(dir)
+            for id2 in (1, 2, 3):
+                dir2 = os.path.join(dir, "dir" + str(id2))
```
nguyenv, line 303: nit 
```suggestion
                dir = os.path.join(basepath, f"dir{id2}")
```

tiledb/tests/test_vfs.py
```diff
@@ -291,6 +291,72 @@ def test_ls(self):
...
+                self.vfs.create_dir(dir2)
+                fname = os.path.join(dir, "file_" + str(id2))
```
nguyenv, line 305: nit 
```suggestion
                dir = os.path.join(basepath, f"dir{id2}")
```

tiledb/libtiledb.pxd
```diff
@@ -1067,8 +1067,14 @@ cdef extern from "tiledb/tiledb.h":
...
-        int (*callback)(const char *, void *) noexcept,
-        void * data)
+        void * data) nogil
+
+    int tiledb_vfs_ls_recursive(
+        tiledb_ctx_t * ctx,
+        tiledb_vfs_t * vfs,
+        const char * uri,
+        int (*callback)(const char *, uint64_t, void *),
+        void * data) nogil
```
nguyenv, line 1077: Is this Cython addition necessary?

nguyenv (approved):

kounelisagis:
Update: Optimized performance for recursive calls. The default wrapper now uses a pure C++ callback to avoid any overhead from Python transition.

---
[sc-45109]

tiledb/cc/vfs.cc
```diff
@@ -41,6 +43,34 @@ void init_vfs(py::module &m) {
...
+            if (callback.is_none()) {
+              std::vector<std::string> paths;
+              tiledb::VFSExperimental::ls_recursive(
+                  ctx, vfs, uri,
+                  [&](const std::string_view path,
+                      uint64_t object_size) -> bool {
+                    paths.push_back(std::string(path));
+                    return true;
+                  });
+              return paths;
```
ihnorton, line 63: For awareness, please note that this will be expensive for large result sets, because pybind11 will generate code to convert it to a native py::list (each item has to be copied into the list before returning - see discussion here: https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#binding-stl-containers). The only way to avoid that is to a create a holder class like this:
```
class PbStdVectorString {
    std::vector<string> m_vec;
    PbStdVectorString(std::string&& v) : m_vec(v) {}
    ...
```
and then wrap `PbStdVectorString` class in pybind11 to provide getitem/slice/iter for views on the underlying data. Not a blocker but might be necessary when you benchmark.

ihnorton (approved):
LGTM, nice work!

nguyenv (approved):

kounelisagis:
Let's merge. Extra performance tests will be part of another story.

PR #1935, opened 2024-03-26
Add wrapping for new consolidation plan API

kounelisagis:


tiledb/consolidation_plan.py
```diff
@@ -0,0 +1,59 @@
+import tiledb.cc as lt
+
+from .ctx import Ctx, CtxMixin, default_ctx
+from .libtiledb import Array
+
+
+class ConsolidationPlan(CtxMixin, lt.ConsolidationPlan):
+    """Represents TileDB ConsolidationPlan class"""
```
nguyenv, line 8: Can we add a more detailed description of what the `ConslidationPlan` actually is?

tiledb/consolidation_plan.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self, ctx: Ctx, array: lt.Array, fragment_size: int):
+        """Class representing the consolidation plan for an array. The plan divides the fragments of the array into different nodes that need to be consolidated together.
+
+        :param ctx: A TileDB Context
+        :param array: The array
+        :param fragment_size: The desired fragment size
+        :raises TypeError: All dimensions must have the same dtype
```
ypatia, line 16: this needs an update?

kounelisagis: Why? I have already added a more detailed description as Vivian proposed. Arguments seem fine.

ypatia: "TypeError: All dimensions must have the same dtype" doesn't seem to me like an error Consolidation Plan would throw.

ihnorton (approved):
This LGTM for wrapping, but let's also add (here or follow-up):
1. repr
2. pythonic interface (`__len__`, `__getitem__` etc. for traversing nodes and URIs)

PR #1915, opened 2024-03-07
Fix vfs readinto when buff is not bytes

kounelisagis:
Trying to ingest an Hamamatsu CMU-1.ndpi https://openslide.cs.cmu.edu/download/openslide-testdata/Hamamatsu/ image using `tiledb.bioimg`:

```
from tiledb.bioimg import from_bioimg
import tiledb
vfs = tiledb.VFS()
with vfs.open('CMU-1.ndpi') as src:
    from_bioimg(src, "test_bioimg")
```

was giving the following error:

```
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/.venv/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "&lt;ipython-input-5-f900874cbef5&gt;", line 2, in &lt;module&gt;
    from_bioimg(src, "/Users/agis/tiledb_dev/TileDB-BioImaging/tests/data/cmu1_failed_v2.ndpi.tdb")
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/tiledb/bioimg/wrappers.py", line 35, in from_bioimg
    return OMETiffConverter.to_tiledb(
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/tiledb/bioimg/converters/base.py", line 357, in to_tiledb
    reader = cls._ImageReaderType(
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/tiledb/bioimg/converters/ome_tiff.py", line 34, in __init__
    self._tiff = tifffile.TiffFile(input_path, is_ndpi=True)
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/.venv/lib/python3.9/site-packages/tifffile/tifffile.py", line 4301, in __init__
    self.pages = TiffPages(self)
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/.venv/lib/python3.9/site-packages/tifffile/tifffile.py", line 7420, in __init__
    page = TiffPage(self.parent, index=pageindex)
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/.venv/lib/python3.9/site-packages/tifffile/tifffile.py", line 8407, in __init__
    mcustarts = tags.valueof(65426)
  File "/Users/agis/tiledb_dev/TileDB-BioImaging/.venv/lib/python3.9/site-packages/tifffile/tifffile.py", line 11655, in valueof
    return tag.value
[...]

PR #1937, opened 2024-03-29
Update type signature for VFS::readinto

kounelisagis:
- Solves this: https://github.com/TileDB-Inc/TileDB-Py/pull/1915#discussion_r1518052243
- In the future, `collections.abc.Buffer` will be used for Python 3.12+


tiledb/vfs.py
```diff
@@ -1,7 +1,7 @@
 import io
 import os
 from types import TracebackType
-from typing import List, Optional, Type, Union
+from typing import ByteString, List, Optional, Type, Union
```
sgillies, line 4: @kounelisagis `ByteString` is deprecated. What would you think about using `bytes | bytearray | memoryview` as suggested at https://docs.python.org/3/library/typing.html#typing.ByteString ?

sgillies (changes requested):
I suggest that we not introduce deprecated usage.

sgillies (approved):
Excellent!

ihnorton (approved):
LGTM, thanks @kounelisagis and @sgillies!

PR #1938, opened 2024-04-01
Remove modular building option

kounelisagis:
Removes the `--modular` option when building TileDB-Py.


kounelisagis:
> I understand the `IF` keyword is deprecated in Cython now, but does that mean we need to get rid of the `TILEDBPY_MODULAR` option? I also have never toggled `TILEDBPY_MODULAR` so don't have any context for how that macro is used.

After discussed with @ihnorton, we will remove the whole `TILEDBPY_MODULAR` option.

ihnorton:
@nguyenv this is untested/unmaintained, and was always kind of brittle; not necessary now with pybind work.

ihnorton (approved):
Thanks

PR #1941, opened 2024-04-03
Add test for Group constructor invalid uri object type

kounelisagis:
Checks for cases where the `uri` passed in the constructor is not of type `Group`.

### DIFF
--- a/tiledb/tests/test_group.py
+++ b/tiledb/tests/test_group.py
@@ -171,7 +171,7 @@ class GroupTest(GroupTestCase):
 
         grp = tiledb.Group(grp_path, "w")
         assert os.path.basename(grp.uri) == os.path.basename(grp_path)
-        array_path = self.path("test_group_members")
+        array_path = self.path("test_group_members_array")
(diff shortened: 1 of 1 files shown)


nguyenv (approved):
Thanks for correcting these!

tiledb/group.py
```diff
@@ -264,6 +265,11 @@ def __init__(
         config: Config = None,
         ctx: Optional[Ctx] = None,
     ):
+        if uri is not None and tiledb.object_type(uri) != "group":
```
ihnorton, line 268: Can you clarify why we need this - the `tiledb::Group` constructor should throw if the uri is not a group?

We should avoid extra calls to introspect like this, if we can (they are expensive on remote/object stores).

ihnorton: If we need this, it seems like we are missing something either in the wrapping or in core error handling/propagation.

kounelisagis: I agree 100%.

ihnorton (approved):

ihnorton:
Thanks

PR #1942, opened 2024-04-03
Improve documentation of from_numpy function

kounelisagis:
List all the keyword arguments for `from_numpy` function.


kounelisagis:
Here we have a conditional pass of `kwargs`.
- [1st case](https://github.com/TileDB-Inc/TileDB-Py/pull/1942/files#diff-5161466dcbd3b3dd74549a439e965eb650fbe6506dcf15cc8ed3c1007901410eL124)
- [2nd case](https://github.com/TileDB-Inc/TileDB-Py/pull/1942/files#diff-5161466dcbd3b3dd74549a439e965eb650fbe6506dcf15cc8ed3c1007901410eL134)

Do we need to list all the arguments for both cases without specifying anything further, right? @ihnorton, @nguyenv

ihnorton (approved):

PR #1948, opened 2024-04-11
Wrap tiledb_array_consolidate_fragments from pybind11

kounelisagis:
+test

### DIFF
--- a/tiledb/tests/cc/test_cc.py
+++ b/tiledb/tests/cc/test_cc.py
@@ -161,7 +161,6 @@ def test_array():
     arr.close()
 
 
-@pytest.mark.xfail
 def test_consolidate_fragments():
     uri = tempfile.mkdtemp()
     ctx = lt.Context()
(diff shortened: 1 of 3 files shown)


tiledb/tests/cc/test_cc.py
```diff
@@ -160,6 +161,39 @@ def test_array():
...
+    # https://github.com/TileDB-Inc/TileDB-Py/pull/1946
+    uris = [uri.split("/")[-1] for uri in uris]
+
+    arr = lt.Array(ctx, uri, lt.QueryType.WRITE)
+    arr.consolidate_fragments(ctx, uris, config)
+    arr.close()
+
+    fragment_info = PyFragmentInfo(uri, schema, False, ctx)
+    # This xfail will be resolved after SC-10782
+    assert fragment_info.get_num_fragments() == 4
```
ihnorton, line 194: Since it is expected right now, let's remove the x-fail, set `assert .. == 1`, and just make a note that fragmentinfo doesn't see the consolidated range.

ihnorton (approved):
LGTM after addressing the one comment. (will need to add to high-level API as well, but that can be done as follow-up)

PR #1953, opened 2024-04-16
Make timestamp overrides optional in tests and add faketime test

kounelisagis:
Currently in TileDB-Py we have a number of tests which override the timestamp for writes in order to avoid test failures when writes happen within the same millisecond, e.g.:

https://github.com/TileDB-Inc/TileDB-Py/blob/ff2149738a5daeee2c11d7660b3c7631fe1a293e/tiledb/tests/cc/test_cc.py#L124
https://github.com/TileDB-Inc/TileDB-Py/blob/ff2149738a5daeee2c11d7660b3c7631fe1a293e/tiledb/tests/test_fragments.py#L100

In libtiledb [2.22.0](https://github.com/TileDB-Inc/TileDB/releases/tag/2.22.0) a sub-millisecond temporal disambiguation of random labels from a single-process is implemented (serialized ordering for writes).

To test this from Python, tests that use timestamp ordering overrides are modified. They now run both with and without overriding and they should succeed in both cases. This solution is temporary; in the future, code that uses ordering override will be removed, ensuring a robust solution.

Additionally, to _directly_ check that the new feature is working as expected, [libfaketime](https://github.com/wolfcw/libfaketime) is utilized. More specifically the [faketime wrapper](https://github.com/wolfcw/libfaketime/blob/master/README#L569-L589) comes in handy in our case to decrease the complexity and have a unified solution for both Linux and macOS.
By using faketime and setting the time multiplier to zero (`faketime -f '+x0' python time_test.py`), we are able to "freeze" time for the whole duration of the test which runs in a subprocess.
Installation of libfaketime means a new CI step. A new test file, `test_timestamp_overrides` is added to run write tests under faketime and check that the expected outcome holds.

Checking if libfaketime exists, means a system call to skip test.

### DIFF
--- a/tiledb/tests/test_timestamp_overrides.py
+++ b/tiledb/tests/test_timestamp_overrides.py
@@ -8,8 +8,7 @@ import pytest
 
 import tiledb
 from tiledb.main import PyFragmentInfo
-
-from .common import DiskTestCase
+from tiledb.tests.common import DiskTestCase
 
 
 def has_libfaketime():
@@ -20,7 +19,7 @@ def has_libfaketime():
         return False
 
 
-class TimestampOverridesTest(DiskTestCase):
+class TestTimestampOverrides(DiskTestCase):
     @pytest.mark.skipif(
         sys.platform == "win32",
         reason="libfaketime is not supported on Windows",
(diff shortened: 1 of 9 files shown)


bekadavis9 (approved):
LGTM!

tiledb/tests/test_timestamp_overrides.py
```diff
@@ -0,0 +1,161 @@
...
+        not has_libfaketime(),
+        reason="libfaketime not installed.",
+    )
+    def test_timestamp_overrides(self):
+        uri_fragments = self.path("time_test_fragments")
+        uri_group_metadata = self.path("time_test_group_metadata")
+
+        python_exe = sys.executable
+        cmd = (
+            f"from tiledb.tests.test_timestamp_overrides import TimestampOverridesTest; "
```
ihnorton, line 38: these have to change to match the class name change above

tiledb/tests/test_timestamp_overrides.py
```diff
@@ -0,0 +1,161 @@
...
+import subprocess
+import sys
+
+import numpy as np
+import pytest
+
+import tiledb
+from tiledb.main import PyFragmentInfo
+
+from .common import DiskTestCase
```
ihnorton, line 12:
```suggestion
from tiledb.tests.common import DiskTestCase
```
so that this file can easily be copied out and run against an old version of tiledb-py

ihnorton (changes requested):
A few suggested changes, but overall great work on the `test_timestamp_overrides`. When I apply the changes here, copy just that file out
- run w/ pytest under tiledb-py<0.28, it fails (as expected)
- run with tiledb-py==0.28, it succeeds. Excellent!

ihnorton (approved):

Issue #1865, opened 2023-11-14
Installation fails on python 3.12.0


endast:
Trying to install tiledb-py on python 3.12.0 fails:
```shell
*[~]$ python -m venv venv312
*[~]$ source venv312/bin/activate
(venv312) *[~]$ python --version
Python 3.12.0
(venv312) *[~]$ python -m pip install tiledb
Collecting tiledb
  Downloading tiledb-0.23.4.tar.gz (326 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 326.3/326.3 kB 5.6 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
ERROR: Exception:
Traceback (most recent call last):
  File "/Users/endast/.pyenv/versions/3.12.0/lib/python3.12/site-packages/pip/_internal/cli/base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
             ^^^^^^^^^^^^^^^
  File "/Users/endast/.pyenv/versions/3.12.0/lib/python3.12/site-packages/pip/_internal/cli/req_command.py", line 245, in wrapper
    return func(self, options, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/endast/.pyenv/versions/3.12.0/lib/python3.12/site-packages/pip/_internal/commands/install.py", line 377, in run
    requirement_set = resolver.resolve(
                      ^^^^^^^^^^^^^^^^^
  File "/Users/endast/.pyenv/versions/3.12.0/lib/python3.12/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 95, in resolve
    result = self._result = resolver.resolve(
                            ^^^^^^^^^^^^^^^^^
  File "/Users/endast/.pyenv/versions/3.12.0/lib/python3.12/site-packages/pip/_vendor/resolvelib/resolvers.py", line 546, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/endast/.pyenv/versions/3.12.0/lib/python3.12/site-packages/pip/_vendor/resolvelib/resolvers.py", line 397, in resolve
    self._add_to_criteria(self.state.criteria, r, parent=None)
[...]

PR #1959, opened 2024-05-01
Enable python 3.12

dudoslav:
Enable python 3.12 for this repo

---
[sc-37726]

### DIFF
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -15,9 +15,7 @@ A high level wrapper around the Pybind11 query_condition.cc implementation for
 filtering query results on attribute and dimension values.
 """
 
-QueryConditionNodeElem = Union[
-    ast.Name, ast.Constant, ast.Call, ast.Num, ast.Str, ast.Bytes
-]
+QueryConditionNodeElem = Union[ast.Name, ast.Constant, ast.Call]
 
 
 @dataclass
@@ -280,8 +278,8 @@ class QueryConditionTree(ast.NodeVisitor):
 
             return (
                 isinstance(variable.args[0], ast.Constant)
-                or isinstance(variable.args[0], ast.Str)
-                or isinstance(variable.args[0], ast.Bytes)
+                or isinstance(variable.args[0], ast.Constant)
+                or isinstance(variable.args[0], ast.Constant)
             )
 
         return isinstance(variable, ast.Name)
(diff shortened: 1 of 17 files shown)


johnkerl:
See also #1865

kounelisagis:
Reasons why tests were failing:
- tarfile supports extraction filters, a mechanism to mitigate some of the security issues: https://docs.python.org/3.12/library/tarfile.html#extraction-filters. Solved it using `filter="fully_trusted"`argument for `tf.extractall()`
- `__repr__` of `collections.OrderedDict` for py3.12 has changed: https://github.com/python/cpython/pull/101661. This means we are getting different expected output for 3.12. Solved it using ellipsis marker: https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS

kounelisagis (approved):
LGTM, maybe wait for another approval to check my changes.

dudoslav:
Please include pytest outputs for 3.12 executions

tiledb/tests/test_compat.py
```diff
@@ -46,7 +46,10 @@ def test_compat_tiledb_py_0_5_anon_attr_dense(self):
 
         path = self.path("tiledb_py_0_6_anon_attr")
         with tarfile.open(fileobj=io.BytesIO(base64.b64decode(array_tgz))) as tf:
-            tf.extractall(path)
+            try:
+                tf.extractall(path, filter="fully_trusted")
```
ihnorton, line 50: Should this order be inverted? ie only use `fully_trusted` if it fails?

ihnorton: Hmm, actually looking at the docs I don't think we need both - we control these entirely, so just use `fully_trusted` by default (unless there's some way it could fail...)

ihnorton (approved):
Minor comment about `fully_trusted`; otherwise LGTM. Thanks @kounelisagis @dudoslav!

tiledb/tests/test_compat.py
ihnorton: Oh I see - probably the keyword argument is not valid in those versions. Disregard.

kounelisagis:
> Please include pytest outputs for 3.12 executions

```
Python 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:20:11) [MSC v.1938 64 bit (AMD64)] on win32
>>> import tiledb
>>> tiledb.version()
(0, 28, 1, 'dev10')
>>> tiledb.libtiledb.version()
(2, 20, 1)
```

tiledb/query_condition.py
```diff
@@ -280,8 +278,8 @@ def is_variable_node(self, variable: QueryConditionNodeElem) -> bool:
 
             return (
                 isinstance(variable.args[0], ast.Constant)
-                or isinstance(variable.args[0], ast.Str)
-                or isinstance(variable.args[0], ast.Bytes)
+                or isinstance(variable.args[0], ast.Constant)
+                or isinstance(variable.args[0], ast.Constant)
```
teo-tsirpanis, line 282: Aren't these redundant?

kounelisagis: ohh wow

PR #1963, opened 2024-05-10
Set an upper bound for numpy to dodge 2.0

sgillies:
Set an upper bound for numpy to dodge 2.0 and avoid incompatible upcoming release until we support it. x-ref sc-46950


ihnorton (approved):

sgillies:
@kounelisagis does this look okay to you?

kounelisagis (approved):

PR #1965, opened 2024-05-13
Add Array.query in docs and improve docs in general

kounelisagis:
- Added `Array.query` in docs, both `SparseArray` and `DenseArray`.
- Fixed some minor errors/warnings while building docs like importing modules from `tiledb.libtiledb` instead of `tiledb`.
- Setting default language for docs - it was causing warning.
- Fixed some typos, like double 'a'
- Addressed the issue with new `__repr__` of `OrderedDict` in Python 3.12 using [np.testing.assert_equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html).


kounelisagis:
- Dense Array: https://tiledb-inc-tiledb--1965.com.readthedocs.build/projects/tiledb-py/en/1965/python-api.html#dense-array
- Sparse Array: https://tiledb-inc-tiledb--1965.com.readthedocs.build/projects/tiledb-py/en/1965/python-api.html#sparse-array

`query()` is included for both.

doc/source/python-api.rst
```diff
@@ -115,23 +115,26 @@ Array
 Dense Array
 -----------
 
-.. autoclass:: tiledb.DenseArray
+.. autoclass:: tiledb.libtiledb.DenseArrayImpl
```
ihnorton, line 118: Rather than importing the internal class `DenseArrayImpl` (which will be confusing) we can use this technique to make sphinx use the documentation from the source of the alias (but keep the aliased name): https://stackoverflow.com/questions/38765577/overriding-sphinx-autodoc-alias-of-for-import-of-private-class#comment115374850_58982001

ihnorton (approved):
LGTM after addressing the one comment inline about doc for aliased classes.

doc/source/python-api.rst
kounelisagis: Modifying https://github.com/TileDB-Inc/TileDB-Py/blob/7b5a4004f0ea6c1ec495e6aa15e58fb02726e6b1/tiledb/__init__.py#L106-L107 to:

```
from .libtiledb import DenseArrayImpl as DenseArray
DenseArray.__module__ = __name__
DenseArray.__name__ = 'DenseArray'
from .libtiledb import SparseArrayImpl as SparseArray
SparseArray.__module__ = __name__
SparseArray.__name__ = 'SparseArray'
```

gives `TypeError: cannot set '__module__' attribute of immutable type 'tiledb.libtiledb.DenseArrayImpl'`.

Having the same problem with all the similar methods I found online.
I also found no way to disable aliases for the whole project.
Any ideas?

ihnorton: Try in conf.py ([like this](https://github.com/shap/shap/blob/6af9e1008702fb0fab939bf2154bbf93dfe84a16/docs/conf.py#L380-L394) from that comment)

kounelisagis: Getting the same error.

kounelisagis: Doesn't work because classes generated by Cython are immutable.
Solve it the way that works for cloud, by adding a wrapper class with basically the same name.

ihnorton (approved):
Nice fix, LGTM!

PR #1973, opened 2024-05-20
Fix CompressionFilter docstring example

kounelisagis:
The provided example in the `CompressionFilter` class docstring incorrectly used `GzipFilter` instead of `CompressionFilter`. This PR fixes the issue, ensuring that the example accurately reflects the intended use of the base class for compression filters, rather than a specific filter type.


ihnorton:
Can you please clarify the fix here in the PR description?

ihnorton (approved):

PR #1975, opened 2024-05-20
Fix schema.dump() so that it prints out consistently in notebooks

kounelisagis:
Jupyter doesn't capture the C stdout of `a.schema.dump()` and instead prints `None`. This PR uses `ArraySchema::dump(FILE*)` to capture the output string, and then print it in Python.


nguyenv (approved):

kounelisagis:
Waiting for https://github.com/TileDB-Inc/TileDB-Py/pull/1984.

johnkerl:
@kounelisagis thanks for the PR!

Good code, wrong title I think -- ? This isn't 2.25.0, it's `dev`; and this is about _restoring_ the `dump()` calls to use the new `operator <<` ...

kounelisagis:
> @kounelisagis thanks for the PR!
> 
> Good code, wrong title I think -- ? This isn't 2.25.0, it's `dev`; and this is about _restoring_ the `dump()` calls to use the new `operator <<` ...

You are right, I am trying to figure out. In any case, this is not finalized yet and I should wait for https://github.com/TileDB-Inc/TileDB/pull/5179 to be merged.

johnkerl (approved):
🔥

ihnorton (approved):

kounelisagis:
Reminder for myself to:
- squash all the commits to two; one for version bump and one for the rest
- merge without squashing those two

ihnorton (approved):
👍 Thanks!

PR #1976, opened 2024-05-20
Fix Query constructor to return error for dense arrays with return_incomplete=True

kounelisagis:
Incomplete queries are only supported for sparse arrays in TileDB-Py at this time. Dense reads will internally reallocate buffers and resubmit the query until successful completion. Docs: https://docs.tiledb.com/main/how-to/arrays/reading-arrays/incomplete-queries

Additionally, create a test to verify the expected behavior for both sparse and dense arrays.


ihnorton (approved):
LGTM, thanks!

PR #1979, opened 2024-05-21
Fix ModuleNotFoundError: No module named 'numpy' on build

kounelisagis:
The following error occurs when trying to build. Fixed by adding `numpy` in build dependencies.

```
  File "/home/vsts/.local/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py", line 373, in <module>
    main()
  File "/home/vsts/.local/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py", line 357, in main
    json_out["return_val"] = hook(**hook_input["kwargs"])
  File "/home/vsts/.local/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py", line 308, in get_requires_for_build_sdist
    return hook(config_settings)
  File "/tmp/build-env-dg88woc1/lib/python3.10/site-packages/setuptools/build_meta.py", line 328, in get_requires_for_build_sdist
    return self._get_build_requires(config_settings, requirements=[])
  File "/tmp/build-env-dg88woc1/lib/python3.10/site-packages/setuptools/build_meta.py", line 295, in _get_build_requires
    self.run_setup()
  File "/tmp/build-env-dg88woc1/lib/python3.10/site-packages/setuptools/build_meta.py", line 311, in run_setup
    exec(code, locals())
  File "<string>", line 9, in <module>
ModuleNotFoundError: No module named 'numpy'
```

Also update `HISTORY`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,11 @@
 [build-system]
-requires = ["setuptools>=64", "wheel", "pybind11", "Cython",
-    "numpy==1.17.*,<2.0 ; python_version == '3.8' and platform_machine != 'aarch64'",
-    "numpy>=1.25 ; python_version >= '3.9"
+requires = [
+    "setuptools>=64",
+    "wheel",
+    "pybind11",
+    "Cython",
+    "numpy== 1.17.*,<2.0 ; python_version == '3.8' and platform_machine != 'aarch64'",
+    "numpy>=1.25 ; python_version >= '3.9'",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,8 @@
 [build-system]
-requires = ["setuptools>=64", "wheel", "numpy", "pybind11", "Cython"]
+requires = ["setuptools>=64", "wheel", "pybind11", "Cython",
+    "numpy==1.17.*,<2.0 ; python_version == '3.8' and platform_machine != 'aarch64'",
+    "numpy>=1.25 ; python_version >= '3.9"
+]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=64", "wheel", "pybind11", "Cython"]
+requires = ["setuptools>=64", "wheel", "numpy", "pybind11", "Cython"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -10,6 +10,7 @@
 
 ## Build system changes
 
+* Fix ModuleNotFoundError: No module named 'numpy' on build by @kounelisagis in https://github.com/TileDB-Inc/TileDB-Py/pull/1979
 * Add support for numpy2 by @kounelisagis in https://github.com/TileDB-Inc/TileDB-Py/pull/1969
 * Fix syntax error in nightly build workflow by @ihnorton in https://github.com/TileDB-Inc/TileDB-Py/pull/1970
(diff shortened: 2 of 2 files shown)


ihnorton (changes requested):
In order for this to work correctly, we need to use `numpy >= 1.25` to ensure that we build against an ABI-compatible version of NumPy (otherwise will get ABI mismatch errors at import time)
- see discussion here: https://github.com/scipy/oldest-supported-numpy/issues/76
- and in README for [this deprecated package](https://pypi.org/project/oldest-supported-numpy/)

I believe we _also_ need to make it conditional for python 3.8 like this:
https://github.com/TileDB-Inc/TileDB-Py/blob/e89521b6392b68c06d22f1d2299ce86793e14e75/pyproject.toml#L36

so the `requires` would look like:
```
"numpy==1.17.*,<2.0 ; python_version == '3.8' and platform_machine != 'aarch64'",
"numpy>=1.25 ; python_version >= '3.9"
```

ihnorton:
cc @sgillies in case any comments, since you worked on cleaning up pyproject deps here in https://github.com/TileDB-Inc/TileDB-Py/pull/1963/files

kounelisagis:
> cc @sgillies in case any comments, since you worked on cleaning up pyproject deps here in https://github.com/TileDB-Inc/TileDB-Py/pull/1963/files

It seems like that PR broke the build procedure. Maybe the problem is the removal of:

```
[tool.setuptools.dynamic]
dependencies = {file = "requirements_wheel.txt"}
```

But just adding `numpy` in `pyproject.toml` doesn't seem to fix it. It then requires `cmake` and adding cmake in `pyproject.toml` won't fix the error: https://gist.github.com/kounelisagis/28f520e52d98e1c04f4f00573b310ace

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=64", "wheel", "pybind11", "Cython"]
+requires = ["setuptools>=64", "wheel", "numpy", "pybind11", "Cython"]
```
ihnorton, line 2:
```suggestion
requires = ["setuptools>=64", "wheel", "pybind11", "Cython",
    "numpy==1.17.*,<2.0 ; python_version == '3.8' and platform_machine != 'aarch64'",
    "numpy>=1.25 ; python_version >= '3.9"
]
```

ihnorton: ^ added suggestion to hopefully clarify what I meant

sgillies:
Sorry about the breakage. Maybe we can work on having our tests build and install a wheel instead of using `python setup.py develop`. I've started doing that on some of my open source projects.

I'll check this out. We may want to start requiring numpy 2 in pyproject.toml for forward compatibility.

sgillies:
@kounelisagis I'm on the dev branch, commit 8355035, and if I add numpy and cmake to pyproject.toml

```
[build-system]
requires = ["setuptools>=64", "wheel", "pybind11", "Cython", "cmake", "numpy<2"]
build-backend = "setuptools.build_meta"
```

Then I get an sdist and a wheel if I run `python -m build`.

```
(venv) $ python -m build
...
Successfully built tiledb-0.23.5.dev95.tar.gz and tiledb-0.23.5.dev95-cp311-cp311-macosx_11_0_universal2.whl
```

I think that should be all that's needed. @ihnorton's comments about Python specific numpy versions look correct.

ihnorton:
> I'm sorry about the breakage.

I don't think there was breakage in any release, slightly unclear to me what that's referring to - no worries.

> Maybe we can work on having our tests build and install a wheel

We do run tests on the wheels with cibuildwheel (but that probably won't catch a numpy mismatch).

kounelisagis:
@ihnorton's suggestion works after @dudoslav's help. NumPy was needed. No need to add cmake. Again, there was a problem with the windows wheel setup. A fix is coming for all the CI.

ihnorton (approved):

PR #1980, opened 2024-05-22
[2.24] Document Azure, GCS and local support for `VFS.ls_recursive`

kounelisagis:
Following https://github.com/TileDB-Inc/TileDB/pull/4981 and https://github.com/TileDB-Inc/TileDB/pull/4997, this PR updates the docstring of `ls_recursive` and the skip condition for its test.

### DIFF
--- a/tiledb/tests/test_vfs.py
+++ b/tiledb/tests/test_vfs.py
@@ -292,8 +292,8 @@ class TestVFS(DiskTestCase):
         )
 
     @pytest.mark.skipif(
-        pytest.tiledb_vfs not in ["file", "s3", "azure"],
-        reason="Only test on local, S3 and Azure",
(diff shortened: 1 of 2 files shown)


tiledb/tests/test_vfs.py
```diff
@@ -292,7 +292,8 @@ def test_ls(self):
         )
 
     @pytest.mark.skipif(
-        pytest.tiledb_vfs not in ["s3", "file"], reason="Only test on S3 and local"
+        pytest.tiledb_vfs not in ["file", "s3", "azure"],
```
ihnorton, line 295: I don't think we have a harness set up to be able to test against these other ones?

ihnorton (approved):

PR #1981, opened 2024-05-24
Update numpy versions everywhere to fix "Some build dependencies conflict with the backend dependencies"

kounelisagis:
The problem we were getting with Azure pipeline when trying to create wheels: https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=39688&view=results

All artifacts are built after this change: https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=39692&view=results

### DIFF
--- a/misc/requirements_wheel.txt
+++ b/misc/requirements_wheel.txt
@@ -1,12 +1,8 @@
-# numpy pinning for ABI forward-compatibility
-numpy==1.16.5 ; python_version < "3.8" and platform_machine !='aarch64'
-numpy==1.17.* ; python_version == "3.8" and platform_machine !='aarch64'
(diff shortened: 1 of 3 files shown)


sgillies (commented):
Excellent @kounelisagis !

ihnorton:
For Python >= 3.9 we can use numpy >= 1.25 unconditionally: https://numpy.org/doc/stable/dev/depending_on_numpy.html#build-time-dependency

For Python 3.8, we have to pin an older version because Numpy 1.25 doesn't support Python 3.8

kounelisagis:
After discussing with @ihnorton, we will wait for https://github.com/TileDB-Inc/TileDB-Py/pull/1971 and not merge breaking the backwards compatibility.

PR #1982, opened 2024-05-28
Deprecate support for `[]` indexing with floats

kounelisagis:
Deprecate support for [] indexing with floats. NumPy no longer allows this (as of [1.12](https://numpy.org/devdocs/release/1.12.0-notes.html#deprecationwarning-to-error)), and it can cause confusion because of the wrap-around semantics.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -2335,7 +2335,8 @@ cdef class DenseArrayImpl(Array):
         idx = replace_ellipsis(self.schema.domain.ndim, selection)
         if check_for_floats(selection):
             warnings.warn(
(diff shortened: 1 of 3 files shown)


nguyenv:
https://www.notion.so/Deprecation-Cycles-in-TileDB-Py-96924f154dc34910a4c61a59b596749a?pm=c

I'd like to be more explicit about what version we will be deprecating this (and it doesn't need to be in two minor release as notion.so doc has written), but wait for @ihnorton to weigh in before making changes.

ihnorton (approved):
LGTM. Let's declare a target version before merging, as @nguyenv suggested (probably 0.31 is fine, since this is already removed in numpy).

PR #1985, opened 2024-06-02
Add tests for blob type attributes

kounelisagis:
Adds tests to ensure that `.dtype` and `.dump()` of `Attribute` class returns what expected.


ihnorton (approved):
Thanks

PR #1988, opened 2024-06-05
Scikit-build-core build system rework

dudoslav:
This PR reworks build system for TileDB-Py to `scikit-build-core` + `CMake`. To comply with our packaging requirements I made sure that the `TILEDB_PATH` environment variable is properly handled. In all cases `RPATH` is set to the location where TileDB core library is present.

Additionally, Github workflows were added to create, test and upload PyPI wheels and sdist packages.

---
[sc-49832]


.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,146 @@
...
+        env:
+          CIBW_BUILD_VERBOSITY: 3
+          CIBW_ENVIRONMENT_MACOS: >
+            CC=clang
+            CXX=clang++
+          MACOSX_DEPLOYMENT_TARGET: "11.0"
+          CIBW_ARCHS: all
+          CIBW_PRERELEASE_PYTHONS: True
+          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
+          # __init__.py interferes with the tests and is included as local file instead of
```
ihnorton, line 49: x-ref `https://github.com/pypa/cibuildwheel/issues/1534` probably

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,146 @@
...
+    name: Wheel ${{ matrix.buildplat[0] }}-${{ matrix.buildplat[1] }}-${{ matrix.python }}
+    runs-on: ${{ matrix.buildplat[0] }}
+    strategy:
+      matrix:
+        buildplat:
+          - [ubuntu-22.04, manylinux_x86_64]
+          - [macos-13, macosx_x86_64]
+          - [macos-14, macosx_arm64]
+          - [windows-2022, win_amd64]
+        python: ["cp39", "cp310", "cp311", "cp312"]
```
kounelisagis, line 26: Do we have a problem with cp38? Or just considered deprecated?

dudoslav: I forgot why, I can try it out if you want

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,146 @@
...
+          name: sdist
+          path: dist
+
+      - name: Install sdist artifact
+        run: pip install --verbose dist/${{ needs.build_sdist.outputs.sdist_name }}
+
+      - uses: actions/checkout@v4
+
+      - name: Install test dependencies
+        run: pip install pytest hypothesis psutil pyarrow
```
kounelisagis, line 111: Maybe `pip install -r misc/requirements_test.txt` would be better. This is the one we are suggesting in the docs. Let's change `requirements_test.txt` if requirements are not up to date.

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,146 @@
...
+        id: get_sdist_name
+        run: |
+          echo "sdist_name=$(ls dist/ | head -n 1)" >> "$GITHUB_OUTPUT"
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: sdist
+          path: dist/*.tar.gz
+
+  test_sdist:
```
kounelisagis, line 81: why are we testing sdist but not wheels?

.github/workflows/ci.yml
```diff
@@ -14,27 +14,29 @@ jobs:
...
       fail-fast: false
     env:
-      # 11.7 necessary due to: https://github.com/actions/setup-python/issues/682#issuecomment-1604261330
-      MACOSX_DEPLOYMENT_TARGET: ${{ matrix.os == 'macos-12' && matrix.python-version == '3.8' && '11.7' || '11' }}
-      #MACOSX_DEPLOYMENT_TARGET: "10.11"
-      # On windows-2019 we are using the Visual Studio generator, which is multi-config and places the build artifacts in a subdirectory
+      MACOSX_DEPLOYMENT_TARGET: "11"
     steps:
       - name: Checkout TileDB-Py `dev`
         uses: actions/checkout@v3
```
kounelisagis, line 29: we should use `actions/checkout@v4` wherever it appears: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

.github/workflows/build-wheels.yml
dudoslav: https://github.com/dudoslav/TileDB-Py/actions/runs/9580398914

.github/workflows/ci.yml
kounelisagis, line 29: we should use `actions/checkout@v4` wherever it appears: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

.github/workflows/build-wheels.yml
dudoslav: We are also testing wheels, but its automatically done by cibuildwheel step

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,146 @@
...
+        os:
+          - macos-13
+          - macos-14
+          - windows-2022
+          - ubuntu-22.04
+        python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v4
```
kounelisagis, line 95: `actions/setup-python@v5`

tiledb/__init__.py
```diff
@@ -19,23 +19,6 @@
...
-
-    del Ctx
-except:
-    try:
-        lib_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "native")
-        ctypes.CDLL(os.path.join(lib_dir, lib_name))
-    except OSError:
-        # Otherwise try loading by name only.
-        ctypes.CDLL(lib_name)
-
```
kounelisagis, line 38: Not sure if this code is supposed to be removed. @ihnorton there is a reason for this here, right?

.github/workflows/build-wheels.yml
kounelisagis: I not see that this file is deleted. I don't believe we can delete this since it contains the dependencies for someone who wants to run the tests.

doc/local-build.sh
kounelisagis, line 1: we need it back

kounelisagis:
Let's not forget to update the docs: https://docs.tiledb.com/main/how-to/installation/building-from-source/python
It would be good to have a draft PR here: https://github.com/TileDB-Inc/TileDB-Developer-Docs before merging this PR.

kounelisagis:
> Let's not forget to update the docs: https://docs.tiledb.com/main/how-to/installation/building-from-source/python It would be good to have a draft PR here: https://github.com/TileDB-Inc/TileDB-Developer-Docs before merging this PR.

same for this: https://github.com/dudoslav/TileDB-Py/blob/db/scikit-build-core/CONTRIBUTING.md

doc/local-build.sh
dudoslav: returned

.github/workflows/ci.yml
```diff
@@ -57,79 +59,28 @@ jobs:
...
           git clone https://github.com/wolfcw/libfaketime/
           cd libfaketime
           sudo make install
           cd ..
 
+      - name: "Build TileDB-Py"
+        run: pip install --verbose .
+
+      - name: "Install test dependencies"
+        run: pip install pytest hypothesis psutil pyarrow
```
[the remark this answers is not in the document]
dudoslav: I decided to put all dependencies into pyproject.toml under optional-dependencies

dudoslav:
I also re-implemented daily tests CI/CD: https://github.com/dudoslav/TileDB-Py/actions/runs/9596846267

tiledb/__init__.py
ihnorton: No, this should not be removed.

pyproject.toml
```diff
@@ -49,22 +48,52 @@ dependencies = [
...
+    "tiledb/*.cc",
+    "tiledb/*.cmake",
+    "tiledb/*.pyx",
+    "tiledb/*.pxd",
+    "tiledb/*.pxi",
+]
+
+[tool.scikit-build.cmake.define]
+TILEDB_PATH = {env="TILEDB_PATH"}
+TILEDB_VERSION = {env="TILEDB_VERSION"}
```
kounelisagis, line 86: Let's also add an option to choose SHA from terminal.

dudoslav: Done

tiledb/__init__.py
dudoslav: As far as I know, right now the produced wheel works on Windows, but I don't understand the incentive behind this code snippet. Why do we need it? I will put it back and test everything, but I am curious why it is there.

jdblischak: Similar looking code was removed from TileDB-VCF when it was migrated to scikit-build-core:

https://github.com/TileDB-Inc/TileDB-VCF/pull/701/files#diff-08ebe1f990ba71b94e5815abc464c931de9c8ffaba809aaa5768ede105399c9b

ihnorton: Ok, please test that editable installs work on Windows with this change -- if they do, then we can proceed.

dudoslav:
Latest diff with Licences:
https://www.diffchecker.com/SxLbtGmC/

dudoslav:
Latest Diff:
https://www.diffchecker.com/eFdvkppd/

ihnorton:
Thanks
- we can exclude all of `doc` and `ci`
- seems like this is a local thing that should not have been included?
image

dudoslav:
> Thanks
> 
>     * we can exclude all of `doc` and `ci`
> 
>     * seems like this is a local thing that should not have been included?
> 
> 
> image

I removed `doc` and `ci` folders and made sure that I create wheel in a clean environment. This is the result right now:
https://www.diffchecker.com/9i4Ge9l2/

Another difference that I found is that the original `setup.py` approach puts the `libtiledb.so` in a different place. I did not notice any difference when running the wheel or testing it, but I want to mention it:

1720088026_grim
1720088016_grim

dudoslav:
I have noticed the following packaging behavior:

When running `pipx run build` if TileDB core library is found in the system it gets picked up and used as the core TileDB library for the wheel. This however never happens in our CI/CD when deploying wheels, so it should be fine, but it can lead to incorrect versions of TileDB included inside wheel for packages produced manually, for example in a Conda environment with TileDB installed in it.

dudoslav:
Testing if Conda feedstock will work:
https://github.com/conda-forge/tiledb-py-feedstock/pull/224

dudoslav:
Seems like conda packages were able to produce:
https://github.com/conda-forge/tiledb-py-feedstock/pull/224

jdblischak:
I tested the shared object copying behavior similar to how I did for TileDB-VCF (https://github.com/TileDB-Inc/TileDB-VCF/issues/717).

I ran into 2 issues:

* When bundling `libtiledb.so` into the installed Python package, I get the error `AttributeError: module 'tiledb.cc' has no attribute 'Config'` when running `import tiledb`
* When using an external `libtiledb.so`, the compilation fails


```sh
docker run --rm -it ubuntu:22.04

# Setup
apt-get update
apt-get install --yes git python-is-python3 python3 python3-pip python3-venv unzip wget
cd
mkdir downloads install-libtiledb

# Install nightly binary
the_date=2024-04-08
wget --quiet -P downloads/ \
  https://github.com/jdblischak/centralized-tiledb-nightlies/releases/download/$the_date/libtiledb-$the_date.tar.gz
tar -C install-libtiledb -xzf downloads/libtiledb-$the_date.tar.gz
export LD_LIBRARY_PATH=$(pwd)/install-libtiledb/lib

# Clone TileDB-Py
git clone https://github.com/TileDB-Inc/TileDB-Py.git
cd TileDB-Py/
git checkout db/scikit-build-core

# Build and install TileDB-Py **with** bundled libtiledb
python -m pip install -v .
##   -- Downloading TileDB default version ...
##   -- [download 100% complete]
##   -- Adding "libtiledb" into install group
##
##   -- Installing: /tmp/tmppmr6xlbn/wheel/platlib/tiledb/libtiledb.so.2.24
##
## Successfully installed numpy-2.0.0 tiledb-0.30.3.dev109+gd18362b

ls /usr/local/lib/python3.10/dist-packages/tiledb/*.so*
## /usr/local/lib/python3.10/dist-packages/tiledb/cc.cpython-310-x86_64-linux-gnu.so
## /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.cpython-310-x86_64-linux-gnu.so
## /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.so.2.24
## /usr/local/lib/python3.10/dist-packages/tiledb/main.cpython-310-x86_64-linux-gnu.so
ldd /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.cpython-310-x86_64-linux-gnu.so | grep libtiledb
## libtiledb.so.2.24 => /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.so.2.24 (0x00007f3ac2e8a000)
readelf -d /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.cpython-310-x86_64-linux-gnu.so | grep RUNPATH
## 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]
python -c "import tiledb; tiledb.libtiledb.version()"
## Traceback (most recent call last):
##   File "<string>", line 1, in <module>
##   File "/root/TileDB-Py/tiledb/__init__.py", line 22, in <module>
##     from .array_schema import ArraySchema
##   File "/root/TileDB-Py/tiledb/array_schema.py", line 10, in <module>
##     from .attribute import Attr
##   File "/root/TileDB-Py/tiledb/attribute.py", line 9, in <module>
##     from .ctx import Ctx, CtxMixin
##   File "/root/TileDB-Py/tiledb/ctx.py", line 16, in <module>
##     class Config(lt.Config):
## AttributeError: module 'tiledb.cc' has no attribute 'Config'

python -m pip uninstall --yes tiledb

# Build and install TileDB-Py **without** bundled libtiledb
export TILEDB_PATH=${HOME}/install-libtiledb/
python -m pip install -v .
## -- Found external TileDB core library
## -- Setting RPATH for targets "main" and "libtiledb" to /root/install-libtiledb/lib
## -- Setting RPATH for target "cc" to /root/install-libtiledb/lib
##
##   In file included from /tmp/tmp8dwduzb1/build/tiledb/libtiledb.cc:1248:
##   /root/install-libtiledb/include/tiledb/tiledb.h:2973:34: note: declared here
##    2973 | TILEDB_DEPRECATED_EXPORT int32_t tiledb_array_delete_fragments(
##         |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
##   ninja: build stopped: subcommand failed.
##
##   *** CMake build failed
##   error: subprocess-exited-with-error
##
##   × Building wheel for tiledb (pyproject.toml) did not run successfully.
##   │ exit code: 1
##   ╰─> See above for output.
##
##   note: This error originates from a subprocess, and is likely not a problem with pip.
##   full command: /usr/bin/python /usr/lib/python3/dist-packages/pip/_vendor/pep517/in_process/_in_process.py build_wheel /tmp/tmpkbu_gaxq
##   cwd: /root/TileDB-Py
##   Building wheel for tiledb (pyproject.toml) ... error
##   ERROR: Failed building wheel for tiledb
## Failed to build tiledb
## ERROR: Could not build wheels for tiledb, which is required to install pyproject.toml-based projects
```

jdblischak:
@ihnorton explained to me the source of the import error I reported above. It's because I ran it within the Git repo, and by default Python looks for `./module.py` and `./module/__init__.py`. Thus it ran the local one instead of the one installed in `site-packages`. Thus the fix it to either run an editable install (`-e`) or move outside of the Git repo.

The reprex below confirms that the import works fine outside of the Git repo:

```sh
docker run --rm -it ubuntu:22.04

# Setup
apt-get update
apt-get install --yes git python-is-python3 python3 python3-pip python3-venv unzip wget
cd

# Clone TileDB-Py
git clone https://github.com/TileDB-Inc/TileDB-Py.git
cd TileDB-Py/
git checkout db/scikit-build-core

# Build and install TileDB-Py **with** bundled libtiledb
python -m pip install -v .

# Run import test outside of Git repo
cd
python -c "import tiledb; print(tiledb.libtiledb.version())"
## (2, 24, 0)
```

jdblischak:
Also, I switched to using the 2.24.2 release tarball, and I was able to compile TileDB-Py against the external libtiledb. The shared object isn't copied, and all the linking looks good.

```sh
docker run --rm -it ubuntu:22.04

# Setup
apt-get update
apt-get install --yes git python-is-python3 python3 python3-pip python3-venv unzip wget
cd
mkdir downloads install-libtiledb

# Install 2.24.2 release binary
# https://github.com/TileDB-Inc/TileDB/releases/tag/2.24.2
wget --quiet -P downloads/ \
  https://github.com/TileDB-Inc/TileDB/releases/download/2.24.2/tiledb-linux-x86_64-2.24.2-76cd03c.tar.gz
tar -C install-libtiledb -xzf downloads/tiledb-linux-x86_64-2.24.2-76cd03c.tar.gz
export LD_LIBRARY_PATH=$(pwd)/install-libtiledb/lib

# Clone TileDB-Py
git clone https://github.com/TileDB-Inc/TileDB-Py.git
cd TileDB-Py/
git checkout db/scikit-build-core

# Build and install TileDB-Py **without** bundled libtiledb
export TILEDB_PATH=${HOME}/install-libtiledb/
python -m pip install -vv .
##  -- Found external TileDB core library
##  -- Setting RPATH for targets "main" and "libtiledb" to /root/install-libtiledb/lib
##  -- Setting RPATH for target "cc" to /root/install-libtiledb/lib

ls /usr/local/lib/python3.10/dist-packages/tiledb/*.so*
## /usr/local/lib/python3.10/dist-packages/tiledb/cc.cpython-310-x86_64-linux-gnu.so
## /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.cpython-310-x86_64-linux-gnu.so
## /usr/local/lib/python3.10/dist-packages/tiledb/main.cpython-310-x86_64-linux-gnu.so

ldd /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.cpython-310-x86_64-linux-gnu.so | grep libtiledb
## libtiledb.so.2.24 => /root/install-libtiledb/lib/libtiledb.so.2.24 (0x00007f7dda6f6000)

readelf -d /usr/local/lib/python3.10/dist-packages/tiledb/libtiledb.cpython-310-x86_64-linux-gnu.so | grep RUNPATH
##  0x000000000000001d (RUNPATH)            Library runpath: [/root/install-libtiledb/lib]

cd
python -c "import tiledb; print(tiledb.libtiledb.version())"
## (2, 24, 2)
```

jdblischak:
Lastly, I got this PR working in a [branch](https://github.com/jdblischak/centralized-tiledb-nightlies/compare/tiledb-py-scikit-build-core) of my centralized-tiledb-nightlies, so I'll be able to quickly migrate the nightly TileDB-Py build after this is merged.

jdblischak (approved):

CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+        else()
+            fetch_prebuilt_tiledb(
+                    VERSION ${TILEDB_VERSION}
+            )
+        endif()
+    else()
+        message(STATUS "Downloading TileDB default version ...")
+        # Download latest release
+        fetch_prebuilt_tiledb(
+                VERSION 2.24.0
```
ihnorton, line 42: Follow-up: we need to move this out at least to the top of the cmakelist if not the workflow file.

ihnorton: Also I'm not sure if this is being validated correctly, because this is the has for 2.24.2

ihnorton (commented):
We need to provide a way to override SETUPTOOLS_SCM_PRETEND_VERSION at the workflow level in order to avoid having to make temporary tags (to get properly named wheels from a QA branch).

tiledb/__init__.py
ihnorton: I tested editable installs on Windows, works perfectly! 💯

CMakeLists.txt
ihnorton: Nevermind, it is definitely being validated 😆 (I grabbed the wrong one)

ihnorton (changes requested):
I ran this branch [here](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/9982922181) and noticed the manylinux wheels are ~20MB larger than the current ones on PyPI for the last release.

dudoslav:
The wheel size is most probably due to `strip=False` and `build_type=RelWithDebInfo`. Latest commit removed these options, lets see if the size will get smaller.

dudoslav:
1721652445_grim
Latest wheels are about 15MB in size

CMakeLists.txt
dudoslav: What about a variable named TILEDB_VERSION? This variable can set on the top of the file

ihnorton (approved):
🎉

PR #1991, opened 2024-06-06
Remove pin to MSVC toolset version in CI.

teo-tsirpanis:
[SC-48973](https://app.shortcut.com/tiledb-inc/story/48973/revert-pins-to-msvc-toolset-14-39)

GitHub Actions is starting to update the MSVC toolset version in CI, causing [occasional failures](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/9398673865/job/25888740034). This PR removes the pin to version 14.39 added in #1906.


kounelisagis (approved):
Thanks!

PR #1994, opened 2024-06-18
Wrap as_built function

kounelisagis:
- Wraps `as_built` function.
- Adds test
- Fixes disable_stats typo

The function was implemented with an optional argument `as_string=False`.
This is because it is easier for the user to use a dictionary structure rather than a string. When the argument is set to `True`, the function returns output identical to the C API.

### DIFF
--- a/tiledb/highlevel.py
+++ b/tiledb/highlevel.py
@@ -1,3 +1,5 @@
+import json
+
 import numpy as np
 
 import tiledb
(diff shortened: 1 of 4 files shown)


tiledb/highlevel.py
```diff
@@ -254,6 +256,21 @@ def is_ndarray_like(arr):
     return schema
 
 
+def as_built(as_string=False):
```
ihnorton, line 259:
```suggestion
def as_built(return_json_string=False):
```

ihnorton (changes requested):

kounelisagis:
@ihnorton FYI I checked what `json.loads()` returns and it's a `<class 'dict'>` as intended.

ihnorton (approved):

PR #1997, opened 2024-07-01
Have `from_pandas` list offending column when `_get_column_infos` fails

kounelisagis:
Adds offending column when `from_pandas` -> `_get_column_infos` fails.

e.g. `NotImplementedError: boolean inferred dtype not supported` becomes `NotImplementedError: boolean inferred dtype not supported (column foo)`.


ihnorton (approved):
Thanks!

leipzig:
good stuff

PR #1998, opened 2024-07-01
Fix `array.query()` incorrectly handling nullables

kounelisagis:
This PR fixes `array.query()` incorrectly handling nullables. It was a logic issue.

When masking an array using `np.ma.array` with a mask derived from a subarray selection, as seen here: https://github.com/TileDB-Inc/TileDB-Py/blob/8c92a30bd3d2a5340c4498f48713f75d5813549d/tiledb/libtiledb.pyx#L3383, the mask should not be inverted because it's already a mask. Invertion is only needed when the array used originates from PyArrow::results, as shown here: https://github.com/TileDB-Inc/TileDB-Py/blob/8c92a30bd3d2a5340c4498f48713f75d5813549d/tiledb/core.cc#L1367. In that case it's the inverse of a mask.

---

A minimal reproducible example:
```
import tiledb
import numpy as np
import pyarrow as pa
import tempfile

uri = tempfile.mkdtemp()

schema = tiledb.ArraySchema(
    domain=tiledb.Domain(tiledb.Dim(domain=[0, 3], dtype=np.int64, tile=4)),
    attrs=[tiledb.Attr(name="a", dtype=np.float64, nullable=True)],
)

with tiledb.open(uri, mode="w") as array:
    array[:] = pa.array([1.0, None, 1.2, None])

with tiledb.open(uri) as array:                                                                                                                                                                  
   A = array[:]
   B = array.query()[:]
```

Returns the following:

A:

```
OrderedDict([('a1',
              masked_array(data=[1.0, --, 1.2, --],
                           mask=[False,  True, False,  True],
                     fill_value=1e+20))])
```

B:

```
OrderedDict([('a1',
              masked_array(data=[--, 0.0, --, 0.0],
                           mask=[ True, False,  True, False],
                     fill_value=1e+20))])
```

A and B should be identical.

### DIFF
--- a/tiledb/libtiledb.pyx
+++ b/tiledb/libtiledb.pyx
@@ -2170,14 +2170,14 @@ cdef class DenseArrayImpl(Array):
                 if attr.isnullable:
                     data = np.array([values[idx] for idx in result[attr.name].data])
                     result[attr.name] = np.ma.array(
-                        data, mask=~result[attr.name].mask)
+                        data, mask=result[attr.name].mask)
                 else:
                     result[attr.name] = np.array(
                         [values[idx] for idx in result[attr.name]])
             else:
                 if attr.isnullable:
                     result[attr.name] = np.ma.array(result[attr.name].data, 
-                        mask=~result[attr.name].mask)
+                        mask=result[attr.name].mask)
 
         return result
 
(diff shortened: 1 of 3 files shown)


ihnorton:
@jp-dark could you confirm this fixes your original issue?

ihnorton (approved):
LGTM pending confirmation.

jp-dark (approved):
LGTM. Thanks!

PR #1999, opened 2024-07-03
Move fragment list consolidation API to pybind

kounelisagis:


tiledb/cc/array.cc
```diff
@@ -55,17 +55,57 @@ void init_array(py::module &m) {
...
+           [](Array &self) {
+             if (self.query_type() == TILEDB_READ) {
+               throw TileDBError("cannot consolidate array opened in readonly "
+                                 "mode (mode='r')");
+             }
+
+             Config config;
+             Context ctx;
+             Array::consolidate(ctx, self.uri(), &config);
+           })
```
nguyenv, line 67: I don't think we should have this one. The default context should be coming from the already existing `tiledb.default_ctx()` rather than constructing a new `tiledb::Context`.

nguyenv (approved):
lgtm. I only have one minor comment.

tiledb/cc/array.cc
kounelisagis: But if we delete this, we cannot have calls like `array.consolidate()` without any arguments.
This can be done with current implementation.

tiledb/tests/test_libtiledb.py
```diff
@@ -3206,12 +3209,41 @@ def write_fragments(target_path):
...
+
+        path = self.path("test_array_consolidate_read_mode")
+
+        create_array(path, dshape)
+        write_fragments(path, dshape, num_writes)
+
+        config = tiledb.Config()
+
+        ctx = tiledb.Ctx(config=config)
+        lt_array = lt.Array(ctx, path, lt.QueryType.READ)
```
nguyenv, line 3241: Ah ok I did not catch this before. The underlying `tiledb.cc` (`lt`) bindings are not meant to be public. The user should still be using `tiledb.Array`. [For example](https://github.com/TileDB-Inc/TileDB-Py/blob/agis/sc-44909/move-fragment-list-consolidation-api-to-pybind/tiledb/array_schema.py#L36-L75), `tiledb.ArraySchema` inherits from `lt.Schema`. And this way we can set the default ctx to `tiledb.default_ctx()`.

nguyenv (changes requested):

tiledb/tests/test_libtiledb.py
nguyenv: What I think is that the static `tiledb.consolidate` method can be in Pybind11, but we still need the Cython  `tiledb.libtiledb.Array.consolidate` method. Otherwise, to create a pure Python `Array` class as we did for `ArraySchema`, you would need to move everything in the `Array` class out of Cython and into Pybind11 which is a TON of work.

kounelisagis: That's exactly the problem. That's the reason that it wasn't clear what to do with that story.

nguyenv (approved):

PR #2000, opened 2024-07-05
Fix OverflowError: Python int too large to convert to C long

kounelisagis:
The problem is related to https://numpy.org/neps/nep-0050-scalar-promotion.html#nep50.
Especially the fact that in numpy2 this behavior is not valid anymore:
```
>>> np.uint64(5) + -1
OverflowError: Python integer -1 out of bounds for uint64
```
And we have to do this:
```
>>> np.uint64(5) + np.int64(-1)
np.float64(4.0)
```

I had made a fix when support for `numpy2` was added, converting `start`,`stop`,`step` to `np.int64`, but this caused this overflow.


kounelisagis:
@jparismorgan  could you confirm this fixes your original issue?

jparismorgan (approved):
Thanks for fixing this!

PR #2007, opened 2024-07-12
Fix malformed doc str for `tiledb.array_schema.ArraySchema`

kounelisagis:


tiledb/array_schema.py
```diff
@@ -38,6 +38,8 @@ class ArraySchema(CtxMixin, lt.ArraySchema):
     Schema class for TileDB dense / sparse array representations
 
     :param domain: Domain of schema
+    :type domain: tiledb.Domain
```
spencerseale, line 41: Can we just remove type from doc str since we're type hinting? If you insist we keep it then type is not uniformly annotated across the doc str.

kounelisagis: done! I now see your point. Thanks for your comment!

spencerseale: Thank you @kounelisagis!

spencerseale (approved):

PR #2009, opened 2024-07-17
Remove deprecated `Array.delete_fragments` code path.

teo-tsirpanis:
[SC-51240](https://app.shortcut.com/tiledb-inc/story/51240/remove-use-of-tiledb-array-delete-fragments-in-the-python-api)

The `tiledb_array_delete_fragments` API has been removed and this PR removes its last use in the Python API. Calling `Array.delete_fragments` as an instance method will now throw.

Fixes https://github.com/TileDB-Inc/conda-forge-nightly-controller/issues/108#issuecomment-2233622013

You might want to snap a branch for the Python API with 2.25 before merging this.


tiledb/libtiledb.pyx
```diff
@@ -1330,22 +1330,9 @@ cdef class Array(object):
...
-            buri = array_instance.uri.encode('UTF-8')
-
-            rc = tiledb_array_delete_fragments(
-                    ctx_ptr,
-                    array_ptr,
-                    buri,
-                    timestamp_start,
-                    timestamp_end
-            )
+            # TODO: Make this method static and entirely remove the conditional.
```
teo-tsirpanis, line 1333: How about we do this now?

kounelisagis: Do you mean to reimplement `tiledb.delete_fragments()`?
It was there before https://github.com/TileDB-Inc/TileDB-Py/pull/1966

teo-tsirpanis: Yes, we could essentially revert #1966. What do you think?

kounelisagis: But `tiledb.delete_fragments` was supposed to be deprecated and the users were forced to use `Array.delete_fragments`. cc. @nguyenv

tiledb/tests/test_fragments.py
```diff
@@ -705,9 +705,9 @@ def write_fragments(target_path, dshape, num_writes):
             assert frags.timestamp_range == ts
 
         if use_timestamps:
-            tiledb.Array.delete_fragments(path, 3, 6)
+            tiledb.delete_fragments(path, 3, 6)
```
teo-tsirpanis, line 708: This is a breaking change. I am not familiar with Python API design, but I prefer the method to be on `Array`.

kounelisagis (approved):

PR #2011, opened 2024-07-22
Update .gitignore

dudoslav:
Add `tiledb/_generated_version.py` into gitignore so that we don't commit it by mistake

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
+tiledb/_generated_version.py
 
 # PyInstaller
 #  Usually these files are written by a python script from a template


ihnorton (approved):

ihnorton (changes requested):
Hmm, we use this in several places: https://github.com/search?q=repo%3ATileDB-Inc%2FTileDB-Py%20_generated_version&type=code

dudoslav:
Ah, my bad. I want it to not get into commit by mistake, but scikit-build-core uses .gitignore to add/remove from files from package

dudoslav:
Updated pyproject.toml https://scikit-build-core.readthedocs.io/en/latest/configuration.html#dynamic-metadata

ihnorton (approved):

PR #2013, opened 2024-07-23
Change wheels artifact names

dudoslav:
Rename artifacts so that we can see what version of python/OS they are built for.

1721742308_grim

New naming: https://github.com/dudoslav/TileDB-Py/actions/runs/10059940542

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -54,7 +54,7 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


kounelisagis (changes requested):
The ideal would be to have one zip file with all wheels inside, as we do now.

Also, the 0.1 version is surely not correct here:
image
is this solved by https://github.com/TileDB-Inc/TileDB-Py/pull/2014?

dudoslav:
This PR does not concern package naming. This is due to my fork not having any release tag.

kounelisagis (approved):

PR #2014, opened 2024-07-23
Set version in workflow dispatch

dudoslav:
This way we can produce wheels with any version number:

From `workflow-dispatch` with input argument `123.1.7`:
https://github.com/dudoslav/TileDB-Py/actions/runs/10060209953/job/27807250342#step:3:24

From release triggered by creating tag `700.11.3`:
https://github.com/dudoslav/TileDB-Py/actions/runs/10060350710/job/27807753665#step:3:26

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -7,6 +7,11 @@ on:
         description: Version to use are release version
         required: true
         type: string
+      test_pypi:
+        description: Upload packages to test.pypi.org
+        required: false
+        type: boolean
+        default: false
   push:
     tags:
       - "*"
@@ -139,11 +144,12 @@ jobs:
           echo "package_version=$(ls dist/ | head -n 1 | cut -d - -f 2)" >> "$GITHUB_OUTPUT"
 
       - name: Upload to test-pypi
+        if: inputs.test_pypi
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
 
       - name: Upload to pypi
         # If we tagged a version release it on PyPI
-        if: startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/tags/') && inputs.version == ''
         uses: pypa/gh-action-pypi-publish@release/v1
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -142,5 +144,6 @@ jobs:
           repository-url: https://test.pypi.org/legacy/
 
       - name: Upload to pypi
+        # If we tagged a version release it on PyPI
```
ihnorton, line 147: Which one takes precedence here?

dudoslav: On release tags upload to both test and pypi. Perhaps I should change it to only upload to PyPI on release. What do you think?

ihnorton: Oh, sorry - I mean if we set `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_TILEDB` then we should probably skip the upload by default to be safe?

dudoslav: Yes, I can definitely do that. What do you think about also adding another input to `workflow_dispatch` called `upload_to_test_pypi` that would be a boolean value and if we set it to true it would upload otherwise not. This would be the only way to upload to test pypi. Then release to normal pypi would only happen on tag and only if pretend version is not set.

kounelisagis (approved):

.github/workflows/build-wheels.yml
ihnorton: Sounds good!

dudoslav: Then the last commit implemented it that way

.github/workflows/build-wheels.yml
```diff
@@ -137,10 +144,12 @@ jobs:
...
       - name: Upload to test-pypi
+        if: inputs.test_pypi
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
 
       - name: Upload to pypi
-        if: startsWith(github.ref, 'refs/tags/')
+        # If we tagged a version release it on PyPI
+        if: startsWith(github.ref, 'refs/tags/') && inputs.version == ''
```
ihnorton, line 154: Do we need to check `!test_pypi` here?

ihnorton (approved):

PR #2019, opened 2024-07-25
Add setup.py fallback for install and develop, and fix some CMake issues

dudoslav:
- REQUIRE -> REQUIRED typo
- support finding TileDBConfig in subdirectory (like previous setup.py)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,20 +7,27 @@ set(PYBIND11_NEWPYTHON ON)
 find_package(
   Python
   COMPONENTS Interpreter Development.Module NumPy
-  REQUIRE
+  REQUIRED
 )
 
 find_package(pybind11 CONFIG REQUIRED)
 
 if (TILEDB_PATH)
-    # If TILEDB_PATH variable is present prepend it to the CMAKE_PREFIX_PATH
-    # so that find_package will be able to find TileDB core library there
-    set(CMAKE_PREFIX_PATH "${TILEDB_PATH};${CMAKE_PREFIX_PATH}")
+    file(REAL_PATH "${TILEDB_PATH}" USER_TILEDB_PATH EXPAND_TILDE)
+    file(GLOB_RECURSE USER_TILEDB_CONFIG_PATH "${TILEDB_PATH}/**/TileDBConfig.cmake")
+    cmake_path(GET USER_TILEDB_CONFIG_PATH PARENT_PATH USER_TILEDB_DIR)
+    set(TileDB_DIR ${USER_TILEDB_DIR})
 endif()
 
-
 # In the future we should use vcpkg
-find_package(TileDB QUIET)
+if (TILEDB_PATH)
+  find_package(TileDB
+  HINTS ${USER_TILEDB_DIR} ${TILEDB_PATH}
+  QUIET NO_DEFAULT_PATH REQUIRED)
+else()
+  find_package(TileDB QUIET)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -7,20 +7,27 @@ set(PYBIND11_NEWPYTHON ON)
...
+  REQUIRED
 )
 
 find_package(pybind11 CONFIG REQUIRED)
 
 if (TILEDB_PATH)
-    # If TILEDB_PATH variable is present prepend it to the CMAKE_PREFIX_PATH
-    # so that find_package will be able to find TileDB core library there
-    set(CMAKE_PREFIX_PATH "${TILEDB_PATH};${CMAKE_PREFIX_PATH}")
+    file(REAL_PATH "${TILEDB_PATH}" USER_TILEDB_PATH EXPAND_TILDE)
```
dudoslav, line 16: Typo `EXPAND_TILDE`?

CMakeLists.txt
```diff
@@ -7,20 +7,27 @@ set(PYBIND11_NEWPYTHON ON)
...
-find_package(TileDB QUIET)
+if (TILEDB_PATH)
+  find_package(TileDB
+  HINTS ${USER_TILEDB_DIR} ${TILEDB_PATH}
+  QUIET NO_DEFAULT_PATH REQUIRED)
+else()
+  find_package(TileDB QUIET)
+endif()
+
 if (NOT TileDB_FOUND)
```
dudoslav, line 31: I am still not sure if we don't want to have a mechanism to FORCE TileDB download

setup.py
```diff
@@ -0,0 +1,44 @@
...
+    if args.command == "develop":
+        cmd.append("-e")
+
+    cmd.append(os.getcwd())
+
+    if args.tiledb:
+        cmd.append(f"-Cskbuild.cmake.define.TILEDB_PATH={args.tiledb}")
+
+    if args.debug:
+        cmd.append(f"-Cskbuild.cmake.define.CMAKE_BUILD_TYPE=DEBUG")
```
dudoslav: I think that we should not use define here but instead:
```
-Cskbuild.cmake.build-type=Debug
```
https://scikit-build-core.readthedocs.io/en/latest/configuration.html#configuring-cmake-arguments-and-defines

dudoslav: I changed this one

kounelisagis (approved):
Checked with:
```
python setup.py develop
python setup.py install
python setup.py develop -v
python setup.py install -v
python setup.py develop --tiledb=/Users/agis/TileDB_workspace/TileDB/build/dist
python setup.py install --tiledb=/Users/agis/TileDB_workspace/TileDB/build/dist
```
It seems like the functionality is equivalent to before.

dudoslav:
I added some local logs and tested `--debug` flag:

When building without --debug:
  -- Build type Release

when building with --debug:
  -- Build type Debug

eddelbuettel:
As @ihnorton has asked about 'what arguments do you folks use with `setup.py` and as I had followed up, this is sad but not a dealbreaker.

```sh
~/git/tiledb-py(dev)$ python3 setup.py build_ext --inplace
usage: setup.py [-h] [--tiledb TILEDB] [--debug] [-v] {install,develop}
setup.py: error: argument command: invalid choice: 'build_ext' (choose from 'install', 'develop')
~/git/tiledb-py(dev)$ 
```

PR #2021, opened 2024-07-26
Fix deprecation and test

kounelisagis:
This PR fixes:
- API deprecation found in https://github.com/jdblischak/centralized-tiledb-nightlies/issues/14 and not fixed in https://github.com/TileDB-Inc/TileDB-Py/pull/1958
- code mistakenly deleted in https://github.com/TileDB-Inc/TileDB-Py/pull/2016

### DIFF
--- a/tiledb/tests/test_pandas_dataframe.py
+++ b/tiledb/tests/test_pandas_dataframe.py
@@ -28,6 +28,9 @@ from .datatypes import RaggedDtype
 
 if not has_pandas():
     pytest.skip("pandas>=1.0,<3.0 not installed", allow_module_level=True)
+else:
+    pd = pytest.importorskip("pandas")
+    tm = pd._testing
 
 
 def make_dataframe_basic1(col_size=10):
--- a/tiledb/cc/array.cc
+++ b/tiledb/cc/array.cc
@@ -96,10 +96,6 @@ void init_array(py::module &m) {
            })
       .def("vacuum", &Array::vacuum)
       .def("create",
-           py::overload_cast<const std::string &, const ArraySchema &,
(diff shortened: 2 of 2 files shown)


teo-tsirpanis (approved):

tiledb/tests/test_pandas_dataframe.py
```diff
@@ -28,6 +28,9 @@
 
 if not has_pandas():
     pytest.skip("pandas>=1.0,<3.0 not installed", allow_module_level=True)
+else:
+    pd = pytest.importorskip("pandas")
```
ihnorton, line 32: This doesn't make sense to me - what are we trying to do here? The original line worked correctly.

kounelisagis: original code is not checking for the python version. New code is not importing pandas, even it exists.

jdblischak:
```suggestion
    import pandas as pd
```

The above call to `has_pandas()`+`pytest.skip()` already skips the test file if pandas isn't installed (or if its the wrong version). Thus if `has_pandas()` evaluates to `True`, you can simply import pandas directly

jdblischak (commented):
Since it wasn't explicitly stated, the reason that the CI for #2016 passed is because pandas is not in `pyproject.toml`. Thus it is not installed and all the tests in `test_pandas_dataframe.py` are skipped

jdblischak:
@kounelisagis thanks for the quick fixes! In the future, could you please `@` mention me in any PRs that address Issue(s) I am troubleshooting?

PR #2022, opened 2024-07-29
Add pandas dependency to test group

dudoslav:


kounelisagis:
This PR resolves the two issues mentioned in https://github.com/TileDB-Inc/TileDB-Py/pull/2021 by @jdblischak:
1. `pandas` wasn't in `pyproject.toml`. Thus, it is not installed in CI and all the tests in `test_pandas_dataframe.py` were skipped.
2. The call to `has_pandas()`+`pytest.skip()` already skips the test file if pandas isn't installed (or if it's the wrong version). Thus if `has_pandas()` evaluates to True, we can simply `import pandas` directly.

dudoslav:
It seems that `pandas` is being installed in CI/CD:
https://github.com/TileDB-Inc/TileDB-Py/actions/runs/10141853243/job/28039902325?pr=2022#step:11:305

kounelisagis (approved):

jdblischak (approved):

ihnorton (changes requested):
We need to test two scenarios in CI:
- pandas is available
- pandas is not available (there is functionality that doesn't require it)

dudoslav:
Should this CI scenario be in all CI runs or only nightly runs?

ihnorton:
IMO all CI runs, but it doesn't have to cover the entire matrix - just one python version is fine.

kounelisagis:
After my latest commit:
- pandas added to both test and wheel test dependencies
- add a step to re-run tests after uninstalling pandas, in the same CI job (for every job). This prevents from having a bigger matrix

If you check the latest jobs, pandas tests are passed in the initial 'Run tests' step and skipped after uninstalling pandas.

@ihnorton, what do you think?

ihnorton (approved):

PR #2023, opened 2024-07-29
Define `TILEDB_REMOVE_DEPRECATIONS` macro for cc and remove deprecated code

kounelisagis:
This PR defines the `TILEDB_REMOVE_DEPRECATIONS` macro, resulting in an error when deprecated core code is used.
Also removes, deprecated code.

Related PRs:
- https://github.com/TileDB-Inc/TileDB-Py/pull/2021
- https://github.com/TileDB-Inc/TileDB-Py/pull/1958

Solves:
https://github.com/jdblischak/centralized-tiledb-nightlies/issues/14

### DIFF
--- a/tiledb/cc/CMakeLists.txt
+++ b/tiledb/cc/CMakeLists.txt
@@ -33,11 +33,13 @@ target_compile_features(
     cxx_std_20
 )
 
-target_compile_definitions(
-    cc
-    PRIVATE
-    TILEDB_REMOVE_DEPRECATIONS
-)
+if (TILEDB_REMOVE_DEPRECATIONS)
+    target_compile_definitions(
+        cc
+        PRIVATE
+        TILEDB_REMOVE_DEPRECATIONS
+    )
+endif()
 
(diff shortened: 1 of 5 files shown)


tiledb/cc/CMakeLists.txt
```diff
@@ -33,6 +33,12 @@ target_compile_features(
     cxx_std_20
 )
 
+target_compile_definitions(
```
dudoslav, line 36: Do we want to have this conditional based on some CMAKE variable like `-D TILEDB_REMOVE_DEPRECATIONS`?

dudoslav: I mean if this is set here we don't need https://github.com/TileDB-Inc/TileDB-Py/blob/dev/pyproject.toml#L82 anymore

teo-tsirpanis (approved):
LGTM as long as CI passes. Thanks!

ihnorton (approved):
LGTM, thanks

PR #2025, opened 2024-07-31
Expose `setup.py` flag to enable deprecations

kounelisagis:
In https://github.com/TileDB-Inc/TileDB-Py/pull/2023, `TILEDB_REMOVE_DEPRECATIONS` macro was defined.
It would be good to also have this functionality in `setup.py`.

Right now, the default behavior is to remove deprecations when build TileDB-Py. In core, the default is to enable them.
Should we keep the logic, or follow the core's?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,9 @@ def main():
     if args.debug:
         cmd.append(f"-Cskbuild.cmake.build-type=Debug")
 
+    if args.enable_deprecations:
+        cmd.append(f"-Cskbuild.cmake.define.TILEDB_REMOVE_DEPRECATIONS=OFF")
+
     if args.v:
         cmd.append("-v")
 
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ def main():
     parser.add_argument("command", choices=["install", "develop"])
     parser.add_argument("--tiledb", type=str, required=False)
     parser.add_argument("--debug", action="store_true")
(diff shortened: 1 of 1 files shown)


dudoslav (changes requested):
I believe that you also need to call:
```
parser.add_argument("--enable_deprecations", action="store_true")
```

Also try this command out, by for example adding

```
message(FATAL_ERROR "I AM HERE")
```
into CMake's IF condition checking TILEDB_REMOVE_DEPRECATIONS and if it fails it means that you set the variable correctly

kounelisagis:
> I believe that you also need to call:
> 
> ```
> parser.add_argument("--enable_deprecations", action="store_true")
> ```
> 
> Also try this command out, by for example adding
> 
> ```
> message(FATAL_ERROR "I AM HERE")
> ```
> 
> into CMake's IF condition checking TILEDB_REMOVE_DEPRECATIONS and if it fails it means that you set the variable correctly

@dudoslav you are right, thanks!

Tested with:
```
if (TILEDB_REMOVE_DEPRECATIONS)
    message(FATAL_ERROR "DEPRECATIONS REMOVED")
    target_compile_definitions(
        cc
        PRIVATE
        TILEDB_REMOVE_DEPRECATIONS
    )
else()
    message(FATAL_ERROR "DEPRECATIONS NOT REMOVED")
endif()
```

dudoslav (approved):
Looks good, thanks @kounelisagis

PR #2026, opened 2024-08-01
Add `Enumeration` in api ref docs

kounelisagis:


spencerseale:
Thanks @kounelisagis! `extend` is missing a parameter doc str in the source is my only nit

spencerseale (approved):

PR #2028, opened 2024-08-01
Get rid all BNF lines in `QueryCondition` docstring

kounelisagis:
Python's Operator Precedence Table from highest to lowest is
1. **Parentheses**: `()`
2. **Unary operators**: `+`, `-`, `~`
3. **Exponentiation**: `**`
4. **Multiplication, Division, Floor Division, Modulus**: `*`, `/`, `//`, `%`
5. **Addition and Subtraction**: `+`, `-`
6. **Bitwise Shift Operators**: `<<`, `>>`
7. **Bitwise AND**: `&`
8. **Bitwise XOR**: `^`
9. **Bitwise OR**: `|`
10. **Comparison Operators**: `==`, `!=`, `<`, `<=`, `>`, `>=`
11. **Logical NOT**: `not`
12. **Logical AND**: `and`
13. **Logical OR**: `or`
14. **Conditional Expressions**: `if - else`
15. **Lambda Expressions**: `lambda`

However, in our BNF in the docs, we have bitwise and logic operators as equal precedence which is not correct.

---
[sc-51316]

### DIFF
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -49,11 +49,11 @@ class QueryCondition:
 
         ``bool_term ::= bool_expr | bool_term and_op bool_expr``
 
-    Logical ``and`` and bitwise ``&`` Boolean operators are given equal precedence.
+    Logical ``and`` is given lower precedence than bitwise ``&``.
 
         ``and_op ::= and | &``
 
-    Likewise, ``or`` and ``|`` are given equal precedence.
+    Likewise, ``or`` is given lower precedence than bitwise ``|``.
 
         ``or_op ::= or | |``
 
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -45,30 +45,16 @@ class QueryCondition:
     Boolean expressions are chained together with Boolean operators. The ``or_op``
     Boolean operators are given lower presedence than ``and_op``.
 
-        ``query_cond ::= bool_term | query_cond or_op bool_term``
-
-        ``bool_term ::= bool_expr | bool_term and_op bool_expr``
-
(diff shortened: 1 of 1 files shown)


tiledb/query_condition.py
```diff
@@ -49,11 +49,11 @@ class QueryCondition:
...
 
-    Logical ``and`` and bitwise ``&`` Boolean operators are given equal precedence.
+    Logical ``and`` is given lower precedence than bitwise ``&``.
 
         ``and_op ::= and | &``
 
-    Likewise, ``or`` and ``|`` are given equal precedence.
+    Likewise, ``or`` is given lower precedence than bitwise ``|``.
 
         ``or_op ::= or | |``
```
nguyenv, line 58: We actually will need to change the BNF notation to reflect this. The expression `and | &` means those two operators are given equal precedence. The comparison operators are also given lower precedence than bitwise operators so that notation needs to be shifted around as well. I can take over rewriting this documentation.

tiledb/query_condition.py
```diff
@@ -49,15 +49,10 @@ class QueryCondition:
 
         ``bool_term ::= bool_expr | bool_term and_op bool_expr``
```
nguyenv, line 50: Let's just get rid all these metasyntax lines. We would need to completely rewrite the BNF for it to correctly match what Python AST has. It's something that will take a lot more time, but we have higher priorities to take care of than rewriting a BNF at the moment.

nguyenv:
I can't highlight the line, but you'll also need to remove line 42 that says **BNF:** since we've now deleted the BNF metasyntax.

nguyenv (approved):

PR #2030, opened 2024-08-01
Update documentation with corrected precedence for bitwise ops

nguyenv:


johnkerl:
@nguyenv can you please fill out the desccription field? we have some open shortcuts to @kounelisagis you can use

kounelisagis:
why do we need a duplicate of https://github.com/TileDB-Inc/TileDB-Py/pull/2028?

Issue #2031, opened 2024-08-02
pyproject.toml does not declare all run-time dependencies 


selimnairb:
Using TileDB-Py 0.31.1, the [packaging](https://pypi.org/project/packaging/) dependency is not declared in `pyproject.toml`, even though it appears to be needed:

```
<snip>
    return pd.DataFrame(A.df[:])
                        ^^^^
  File "libtiledb.pyx", line 1802, in tiledb.libtiledb.Array.df.__get__
  File "/Users/USER/venv/lib/python3.12/site-packages/tiledb/multirange_indexing.py", line 410, in __init__
    check_dataframe_deps()
  File "/Users/USER/venv/lib/python3.12/site-packages/tiledb/dataframe_.py", line 26, in check_dataframe_deps
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'
```

The [conda-forge feedstock](https://github.com/conda-forge/tiledb-py-feedstock/blob/main/recipe/meta.yaml#L45) does declare the `packaging` dependency.

My expectation is that `pyproject.toml` / installation instructions for TileDB-Py would declare all necessary run-time dependencies and that these would match those of the conda-forge feedstock, as I don't normally use conda (especially in containers), and it should not be required to use TileDB-Py.

kounelisagis:
Closed by https://github.com/TileDB-Inc/TileDB-Py/pull/2032

PR #2032, opened 2024-08-05
Add packaging as dependency

dudoslav:
This PR should fix issue: https://github.com/TileDB-Inc/TileDB-Py/issues/2031

I am not sure why the automated tests did not pick up on this missing dependency.


kounelisagis:
It gets installed from some other dependency when executing `pip install -e '.[test]'`. This is not the case when running `pip install -e .`. It was [there](https://github.com/TileDB-Inc/TileDB-Py/blob/release-build-0.30.1/requirements.txt#L5) before the build system change, so it seems like it's needed.

PR #2034, opened 2024-08-05
Re-enable automatic upload to pypi

dudoslav:
Do not forget to prepare the release environment (pypi)

^^ The release environment should be prepared.

---
[sc-54205]


ihnorton (changes requested):
We need to add a test against S3 and `tiledb://` before we turn this on unconditionally.

kounelisagis:
Cancelling the checks since we are missing 3 environment variables: `TILEDB_TOKEN`, `TILEDB_PATH`, and `S3_PATH`.
We could hardcode the paths but I believe this will give us more flexibility to move the arrays if needed.

tiledb/tests/test_array_from_cloud.py
```diff
@@ -0,0 +1,43 @@
+import tiledb
+import os
+import pytest
+
+tiledb_token = os.getenv('TILEDB_TOKEN')
+tiledb_path = os.getenv('TILEDB_PATH')
+s3_path = os.getenv('S3_PATH')
+
+@pytest.mark.skipif(tiledb_token is None, reason="No TileDB token found. Please set TILEDB_TOKEN env variable.")
```
ihnorton, line 9: Let's also check `env["CI"]` and don't skip, because we don't want to skip this if we're running on CI in case there was a problem in the token setup -- then we would silently fail.

ihnorton (commented):
- Let's remove s3 since we are using core binaries that are already tested
- Let's do a write and a read to `tiledb://` URI

kounelisagis:
We still need `TILEDB_TOKEN`, `TILEDB_NAMESPACE`, and `S3_BUCKET`, so the new test is expected to fail.

ihnorton (approved):

PR #2036, opened 2024-08-07
Bump libtiledb and numpy versions for daily tests / make `CurrentDomain` and `NDRectangle` code conditional

kounelisagis:
Target of this PR: **Fix failing daily tests**

In this PR:
1. The versions in [daily-tests.yml](https://github.com/TileDB-Inc/TileDB-Py/blob/52f109240ef470fbc2157cabae4390d6c4f6ae44/.github/workflows/daily-tests.yml) needed to be bumped to `2.25.0` and `2.24.0`.
2. This was still causing problems since `CurrentDomain` and `NDRectangle` are wrapped in `dev` branch of TileDB-Py but only introduced in TileDB [`2.25.0`](https://github.com/TileDB-Inc/TileDB/releases/tag/2.25.0). Daily tests are also running for `2.24.0`. This means that the code added to wrap these classes should be conditional.
3. `numpy` earliest versions in daily tests were updated to the earliest versions currently supported as defined in https://github.com/TileDB-Inc/TileDB-Py/blob/acb05182baee48a1ca68146dbd249bfa94c92599/pyproject.toml#L44-L47

Manually triggered 'Daily Tests' workflow: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/10326361797
Closes https://github.com/TileDB-Inc/TileDB-Py/issues/2035


dudoslav (approved):

ihnorton (approved):

tiledb/__init__.py
```diff
@@ -2,6 +2,8 @@
 import os
 import sys
 
+from tiledb.libtiledb import version as libtiledb_version
```
ihnorton, line 5: Let's avoid adding this to the public API - if we have to do this to fix imports we should import as `__libtiledb_version`

ihnorton (approved):
LGTM other than minor change.

tiledb/__init__.py
kounelisagis: I hadn't thought of it, but I think I've found a good solution!

PR #2039, opened 2024-08-16
Use the `clang-format` pre-commit hook updated to the latest version

kounelisagis:
Currently, clang-format runs only on linux in a non-standard way: https://github.com/TileDB-Inc/TileDB-Py/blob/d447f12f500884cdd01733d852a1bc7803a8de9e/.github/workflows/format.yml#L15-L19

It should be better to use the pypi edition of it: https://pypi.org/project/clang-format/ which also provides pre-commit hook and it's system agnostic.

This PR:
- Introduces the previously mentioned pre-commit hook. (v19.1.7)
- Copies the `.clang-format` file from [TileDB-SOMA](https://github.com/single-cell-data/TileDB-SOMA).  
- Removes the old method/scripts of executing `clang-format`.  
- Updates all `actions/setup-python` instances to version 5.  
- Updates all `actions/checkout` instances to version 4.  
- Updates `ruff` and `black` to their latest versions.  
- Reformats the code to reflect the changes above.  

---

`[sc-43783]`


johnkerl:
@kounelisagis @ihnorton can we perhaps have a discussion about syncing `clang-format` version changes across packages?

For those of us doing development on a single repo, whatever version the repo is using, the developer can easily switch to.

For those of us doing multi-repo development -- and given that `clang-format` 14 disagrees with `clang-format` 18 -- we could have a situation where a given developer needs to switch `clang-format` versions back & forth when pushing PRs to different packages.

I'm aware of personally:

* Core
* TileDB-Py
* TileDB-SOMA
* TileDB-VCF
* and perhaps others

I'm all in favor of moving "into the future" -- just flagging the opportunity for a synchronize (unless there is such underway and I've missed it).

kounelisagis:
Rebasing onto the main branch

kounelisagis:
Rebased onto the main branch

johnkerl (approved):

PR #2041, opened 2024-08-20
Improve Query Condition error messages to align with TileDB-SOMA-Py

kounelisagis:
Following https://github.com/single-cell-data/TileDB-SOMA/pull/2830, https://github.com/single-cell-data/TileDB-SOMA/pull/2831, and https://github.com/single-cell-data/TileDB-SOMA/pull/2864 this PR improves some `QueryCondition` error messages.
In short, it improves error messages:
- when targeted attribute contains `.`
- on non-constant query-condition RHS
- for invalid operators in query-condition parser

---

[sc-52811]
[sc-52812]
[sc-52813]


johnkerl (approved):
🔥

Issue #2043, opened 2024-08-20
[BUG] exception when call `label_index()` on a Query object


sric0880:
when call

```python
A.query(attrs=['name'], coords=True).label_index(['dt'])
```
has a exception below:

```shell
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[94], [line 24](vscode-notebook-cell:?execution_count=94&line=24)
     [22](vscode-notebook-cell:?execution_count=94&line=22)     print(non_empty_domain)
     [23](vscode-notebook-cell:?execution_count=94&line=23)     print(A.meta['last_day'])
---> [24](vscode-notebook-cell:?execution_count=94&line=24)     print(A.query(attrs=['name'], coords=True).label_index(['dt', 'symbol']))
     [25](vscode-notebook-cell:?execution_count=94&line=25)     # print(A.query(attrs=['name'], coords=True).label_index(['dt', 'symbol'])[:])
     [26](vscode-notebook-cell:?execution_count=94&line=26)     # print(a.keys())
     [28](vscode-notebook-cell:?execution_count=94&line=28) schema = tiledb.ArraySchema.load(array_bars_stock_daily)

File libtiledb.pyx:2158, in tiledb.libtiledb.Query.label_index()

ModuleNotFoundError: No module named 'tiledb.multirange_indexer'
```

In the source file `libtiledb.pyx`

```python
    def label_index(self, labels):
        """Apply Array.label_index with query parameters."""
        from .multirange_indexer import LabelIndexer
        return LabelIndexer(self.array, tuple(labels), query=self)
```

`multirange_indexer.py` has already changed to `multirange_indexing.py`, so module 'tiledb.multirange_indexer' cannot be found.

kounelisagis:
Thanks!

PR #2044, opened 2024-08-20
Fix exception when calling `label_index` on a Query object

sric0880:
Closes #2043

PR #2045, opened 2024-08-20
Fix exception when calling `label_index` on a Query object

kounelisagis:
Closes https://github.com/TileDB-Inc/TileDB-Py/issues/2043


kounelisagis:
Duplicate of https://github.com/TileDB-Inc/TileDB-Py/pull/2044

PR #2046, opened 2024-08-21
Make `from_pandas` respect `column_types` for index dimensions

kounelisagis:
Historically, `from_pandas` did not consider the `column_types` argument for index columns, only for regular columns. This PR resolves that issue.

---

[sc-19316]

### DIFF
--- a/tiledb/dataframe_.py
+++ b/tiledb/dataframe_.py
@@ -420,7 +420,9 @@ def _sparse_from_dtypes(dtypes, sparse=None):
     return sparse if sparse is not None else False
 
 
-def create_dims(df, index_dims, tile=None, full_domain=False, filters=None):
+def create_dims(
+    df, index_dims, column_infos, tile=None, full_domain=False, filters=None
+):
     check_dataframe_deps()
     import pandas as pd
 
@@ -445,7 +447,10 @@ def create_dims(df, index_dims, tile=None, full_domain=False, filters=None):
         else:
             raise ValueError(f"Unknown column or index named {name!r}")
 
-        dtype = ColumnInfo.from_values(values).dtype
+        if name in column_infos:
+            dtype = column_infos[name].dtype
+        else:
+            dtype = ColumnInfo.from_values(values).dtype
         internal_dtype = dtype
 
         if name == "__tiledb_rows" and isinstance(index, pd.RangeIndex):
@@ -659,6 +664,7 @@ def _create_array(uri, df, sparse, full_domain, index_dims, column_infos, tiledb
     dims, dim_metadata = create_dims(
         df,
         index_dims,
+        column_infos,
         full_domain=full_domain,
         tile=tiledb_args.get("tile"),
         filters=tiledb_args.get("dim_filters", True),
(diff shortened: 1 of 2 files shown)


johnkerl:
Just a quick note -- TileDB attrs can handle UTF-8 or ASCII; dims, ASCII only but not UTF-8 at this time. (And per @davisp it would be a significant project to support these.) Just a heads-up in case a user requests UTF-8 on a dim.

kounelisagis:
> Just a quick note -- TileDB attrs can handle UTF-8 or ASCII; dims, ASCII only but not UTF-8 at this time. (And per @davisp it would be a significant project to support these.) Just a heads-up in case a user requests UTF-8 on a dim.

That's a really good point, @johnkerl. Thanks for bringing this to my attention - I wasn't aware of it. When trying to create a standalone UTF-8 `Dim`, it doesn't work:

```
>>> tiledb.Dim(name="d1", domain=(0, 9), dtype="U")
TypeError: invalid Dim dtype dtype('<U')
```

However, there's silent handling of this when using `from_pandas`, for some reason:
https://github.com/TileDB-Inc/TileDB-Py/blob/03336ef0dce484853dcf0e48fd0cd3918743fc97/tiledb/dataframe_.py#L395

After testing this, the TileDB array is created as usual, but with the dimension casted to ASCII ('S').

Would you suggest removing this handling?

kounelisagis:
`dtype` handling differs for strings only because of the inline if-else statement. For other data types, the handling is the same for both attributes and dimensions.

The special thing we are doing when creating a dimension (as opposed to an attribute) is to define variables like `dim_min`, `dim_max`, `tile_max`, etc. However, as far as I can see, this is the only way to create a dimension. So it's hard to just have a unified helper function; we can do it but, again, we need all these _special_ code for dimension.

In this PR, the change is that instead of just deriving the data type from the values, the user might have explicitly provided the dtype. If this is invalid, or if it's UTF8 or something similar, it should not be a problem since the error handling remains the same.

kounelisagis:
> Hi @kounelisagis ! Sorry for the delay in replying.
> 
> > After testing this, the TileDB array is created as usual, but with the dimension casted to ASCII ('S').
> 
> This is good! 🙂
> 
> > Would you suggest removing this handling?
> 
> I'd be wary of changes to bytes/ASCII/UTF-8 handling as this is (in my experience) fiddly across packages and languages.
> 
> Maybe somehow extract some code into a helper function with an `is_dim` or `is_attr` argument? So that we can infer contextually what's the right thing to do for a dim, and the right thing to do for an attr?
> 
> And, of course, plenty of unit-test coverage -- is my advice at present ...

`dtype` handling differs for strings only because of the inline if-else statement. For other data types, the handling is the same for both attributes and dimensions.

The special thing we are doing when creating a dimension (as opposed to an attribute) is to define variables like `dim_min`, `dim_max`, `tile_max`, etc. However, as far as I can see, this is the only way to create a dimension. So it's hard to just have a unified helper function; we can do it but, again, we need all these _special_ code for dimension.

In this PR, the change is that instead of just deriving the data type from the values, the user might have explicitly provided the dtype. If this is invalid, or if it's UTF8 or something similar, it should not be a problem since the error handling remains the same.

johnkerl:
All good, thanks @kounelisagis ! :)

ihnorton (approved):

PR #2049, opened 2024-08-22
Move `tiledb_object_*` API calls from Cython to pybind

kounelisagis:
[sc-53309]


tiledb/core.cc
```diff
@@ -1681,6 +1681,26 @@ py::str as_built_dump() {
...
+auto walk_callback = [](const char *path_ptr, tiledb_object_t obj,
+                        void *pyfunc) {
+  std::string my_path(path_ptr);
+  std::string my_objtype;
+  if (obj == TILEDB_GROUP) {
+    my_objtype = "group";
+  }
+  if (obj == TILEDB_ARRAY) {
+    my_objtype = "array";
+  }
```
nguyenv, line 1693: What happens in the case of a `TILEDB_INVALID`?

tiledb/highlevel.py
```diff
@@ -342,6 +343,72 @@ def as_built(return_json_string=False):
     return json.loads(res)
 
 
+def object_type(uri: str, ctx: tiledb.Ctx = None) -> Optional[str]:
+    """Returns the TileDB object type at the specified URI as a string.
+
+    :param uri: URI of the TileDB resource
+    :param ctx: The TileDB Context
+    :return: object type string or None if something goes wrong
```
nguyenv, line 351:
```suggestion
    :return: object type string ("array" or "group") or None if invalid TileDB object```

File: tiledb/tests/test_util.py [hunk 3]
```diff
@@ -107,3 +108,58 @@ def test_array_exists(self):
...
+        uri = tempfile.mkdtemp()
+
+        tiledb.group_create("{}/my_group".format(uri))
+        tiledb.group_create("{}/my_group/dense_arrays".format(uri))
+        tiledb.group_create("{}/my_group/sparse_arrays".format(uri))
+
+        create_array("{}/my_group/dense_arrays/array_A".format(uri), False)
+        create_array("{}/my_group/dense_arrays/array_B".format(uri), False)
+        create_array("{}/my_group/sparse_arrays/array_C".format(uri), True)
+        create_array("{}/my_group/sparse_arrays/array_D".format(uri), True)
```
nguyenv, line 135: Would suggest using the `Pathlib` module

tiledb/tests/test_util.py
```diff
@@ -107,3 +108,58 @@ def test_array_exists(self):
...
+                tiledb.Dim(name="rows", domain=(1, 4), tile=4, dtype=np.int32),
+                tiledb.Dim(name="cols", domain=(1, 4), tile=4, dtype=np.int32),
+            )
+            schema = tiledb.ArraySchema(
+                domain=dom, sparse=sparse, attrs=[tiledb.Attr(name="a", dtype=np.int32)]
+            )
+            if sparse:
+                tiledb.SparseArray.create(array_name, schema)
+            else:
+                tiledb.DenseArray.create(array_name, schema)
```
nguyenv, line 124:
```suggestion
            tiledb.Array.create(array_name, schema)
```

It will pick up whether to create a sparse or dense array from the passed in schema.

nguyenv (commented):
Should we be binding from the `ObjectIter` CPP API instead of using the `tiledb_object_*` C API for some of these (like `walk`)?

tiledb/core.cc
kounelisagis: Would
```
else {
    return 0;
}
```
be ok?

kounelisagis:
> Should we be binding from the `ObjectIter` CPP API instead of using the `tiledb_object_*` C API for some of these (like `walk`)?

Seems like there is no direct wrapping in the CPP API. It uses the C API to build other (more complex than we need) functionalities. Example: https://github.com/TileDB-Inc/TileDB/blob/release-2.24/tiledb/sm/cpp_api/object_iter.h#L208-L222

tiledb/core.cc
```diff
@@ -1681,6 +1681,27 @@ py::str as_built_dump() {
...
+  if (obj == TILEDB_GROUP) {
+    my_objtype = "group";
+  } else if (obj == TILEDB_ARRAY) {
+    my_objtype = "array";
+  } else {
+    return 0;
+  }
+  try {
+    py::function func =
+        py::reinterpret_borrow<py::function>((PyObject *)pyfunc);
```
ihnorton, line 1697: This is an unchecked cast, we should use `py::cast` here instead right? https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv4I0E18reinterpret_borrow1T6handle

tiledb/core.cc
```diff
@@ -1681,6 +1681,27 @@ py::str as_built_dump() {
...
+    func(my_path, my_objtype);
+  } catch (const std::exception &e) {
```
ihnorton, line 1699: Please add a test with a callback that throws a Python error to make sure this works correctly -- in the current code we are only catching `StopIteration`, and other Python errors should (I think) just be bubbled up by cython.

ihnorton (changes requested):

tiledb/core.cc
kounelisagis: You were right that all exceptions thrown from the callback were not propagated.
After my changes, the behavior is following the cython code which was only catching `StopIteration` (I used the pybind equivalent).

ihnorton (approved):
LGTM, thanks for the changes!

PR #2055, opened 2024-09-02
Don't check `dtype` if `column_types` is not set - `from_pandas`

kounelisagis:
The tests added in #2046 were also checking the initial dtype of the columns as provided by `from_pandas`. However, since this depends on the system, Python, or NumPy (as evidenced by daily test failures), it turns out that this is not a good idea. Let's instead only check the dtypes enforced by the `column_types` argument.

Fixes https://github.com/TileDB-Inc/TileDB-Py/issues/2054


teo-tsirpanis (approved):
I was explained this PR and it looks good to me. Trying to guess numpy's implementation-defined defaults does not make much sense.

PR #2060, opened 2024-09-08
Convert config values to strings in VFS init

gspowley:
Enable passing config values with non-string types, for example:

```python
config = {"vfs.s3.no_sign_request": True}

vfs = tiledb.VFS(config=config)
```

which currently generates this error:
```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. tiledb.cc.Config(arg0: tiledb.cc.Config)
    2. tiledb.cc.Config()
    3. tiledb.cc.Config(arg0: dict[str, str])
    4. tiledb.cc.Config(arg0: str)

Invoked with: {'vfs.s3.no_sign_request': True}
```

### DIFF
--- a/tiledb/vfs.py
+++ b/tiledb/vfs.py
@@ -5,6 +5,7 @@ from typing import Callable, List, Optional, Type, Union
 
 import numpy as np
 
+import tiledb
 import tiledb.cc as lt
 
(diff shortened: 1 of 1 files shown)


kounelisagis (approved):
Nice finding! Thanks!

PR #2062, opened 2024-09-12
Make dump calls conditional to fix errors when building against <2.26

kounelisagis:
After https://github.com/TileDB-Inc/TileDB-Py/commit/ce8310e60a493be0fc412c92a8fb937575385649 `operator<<` is used instead of `dump` functions. Since this is introduced in `2.26.0`, building against older libtiledb versions results in failing tests: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/10824513851

Daily tests workflow: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/10832807692

Fixes: https://github.com/TileDB-Inc/TileDB-Py/issues/2061

### DIFF
--- a/tiledb/cc/attribute.cc
+++ b/tiledb/cc/attribute.cc
@@ -108,9 +108,13 @@ void init_attribute(py::module &m) {
       .def("_set_enumeration_name", set_enumeration_name)
 
       .def("_dump", [](Attribute &attr) {
+#if TILEDB_VERSION_MAJOR >= 2 && TILEDB_VERSION_MINOR >= 26
         std::stringstream ss;
         ss << attr;
         return ss.str();
+#else
+        attr.dump();
+#endif
       });
 }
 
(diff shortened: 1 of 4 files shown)


ihnorton (approved):
Seems correct, but I'm still unclear about what happened here because on latest TileDB-Py+TileDB `dev` branches, I still get `Array type: sparse` in the dump output.

```
In [10]: schema.dump()
- Array type: sparse
- Cell order: row-major
- Tile order: row-major
- Capacity: 10000
- Allows duplicates: false
- Coordinates filters: 1
  > ZSTD: COMPRESSION_LEVEL=-1
- Offsets filters: 1
  > ZSTD: COMPRESSION_LEVEL=-1
- Validity filters: 1
  > RLE: COMPRESSION_LEVEL=-1

### Dimension ###
- Name: d
- Type: STRING_ASCII
- Cell val num: var
- Domain: null
- Tile extent: null
- Filters: 0

### Attribute ###
- Name: a
- Type: STRING_UTF8
- Nullable: true
- Cell val num: var
- Filters: 0
- Fill value:
- Fill value validity: 0
### Current domain ###
- Version: 0
- Empty: 1



In [11]: tiledb.version(), tiledb.libtiledb.version()
Out[11]: ((0, 32, 0), (2, 27, 0))
```

PR #2065, opened 2024-09-16
Move `SparseArrayImpl` to pure Python

kounelisagis:
[sc-55282]


tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+class SparseArrayImpl(Array):
+    """Class representing a sparse TileDB array (internal).
+
+    Inherits properties and methods of :py:class:`tiledb.Array`.
+    """
+
+    def __init__(self, *args, **kw):
+        super().__init__(*args, **kw)
+        if not self.schema.sparse:
+            raise ValueError("Array at '{}' is not a sparse array".format(self.uri))
```
nguyenv, line 27:
```suggestion
            raise ValueError(f"Array at '{self.uri}' is not a sparse array")
```

tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+
+        dom = self.schema.domain
+        idx = index_as_tuple(selection)
+        idx = replace_ellipsis(dom.ndim, idx)
+        if check_for_floats(selection):
+            warnings.warn(
+                "The use of floats in selection is deprecated. "
+                "It is slated for removal in 0.31.0.",
+                DeprecationWarning,
+            )
```
nguyenv, line 352: Looks like we can remove this now

tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+        subarray.add_ranges([list([x]) for x in dim_ranges])
+        return self._read_sparse_subarray(subarray, attr_names, cond, layout)
+
+    def __repr__(self):
+        if self.isopen:
+            return "SparseArray(uri={0!r}, mode={1}, ndim={2})".format(
+                self.uri, self.mode, self.schema.ndim
+            )
+        else:
+            return "SparseArray(uri={0!r}, mode=closed)".format(self.uri)
```
nguyenv, line 365:
```suggestion
        if self.isopen:
            return f"SparseArray(uri={self.uri}, mode={self.mode}, ndim={self.schema.ndim})"
        else:
            returnf "SparseArray(uri={self.uri}, mode=closed)"
```

tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+
+        if dim is not None and not self.domain.has_dim(dim):
+            raise ValueError("Array does not contain Dimension '{}'.".format(dim))
+
+        query = self.query(attrs=[])[:]
+
+        if dim:
+            dim_values = tuple(np.unique(query[dim]))
+        else:
+            from collections import OrderedDict
```
nguyenv, line 449: Do we want to move all these `from collections import OrderedDict`  imports to the top level?

tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+            if self.schema.has_attr(final_name) and self.attr(final_name).isnullable:
+                out[final_name] = np.ma.array(
+                    out[final_name], mask=~results[name][2].astype(bool)
+                )
+
+        return out
+
+    def unique_dim_values(self, dim=None):
+        if dim is not None and not isinstance(dim, str):
+            raise ValueError("Given Dimension {} is not a string.".format(dim))
```
nguyenv, line 439:
```suggestion
            raise ValueError(f"Given Dimension {dim} is not a string.")
```

tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+
+        if dim is not None and not self.domain.has_dim(dim):
+            raise ValueError("Array does not contain Dimension '{}'.".format(dim))
```
nguyenv, line 442:
```suggestion
            raise ValueError(f"Array does not contain Dimension '{dim}'.")
```

tiledb/sparse_array.py
```diff
@@ -0,0 +1,455 @@
...
+        raise TypeError("SparseArray length is ambiguous; use shape[0]")
+
+    def __setitem__(self, selection, val):
+        """Set / update sparse data cells
+
+        :param tuple selection: N coordinate value arrays (dim0, dim1, ...) where N in the ndim of the SparseArray,
+            The format follows numpy sparse (point) indexing semantics.
+        :param value: a dictionary of nonempty array attribute values, values must able to be converted to 1-d numpy arrays.\
+            if the number of attributes is one, then a 1-d numpy array is accepted.
+        :type value: dict or :py:class:`numpy.ndarray`
```
nguyenv, line 39: This mismatch between the arg name and doc existed prior to this PR, but let's correct it here

```suggestion
    def __setitem__(self, selection, val):
        """Set / update sparse data cells

        :param tuple selection: N coordinate value arrays (dim0, dim1, ...) where N in the ndim of the SparseArray,
            The format follows numpy sparse (point) indexing semantics.
        :param val: a dictionary of nonempty array attribute values, values must able to be converted to 1-d numpy arrays.\
            if the number of attributes is one, then a 1-d numpy array is accepted.
        :type val: dict or :py:class:`numpy.ndarray`
```

nguyenv (approved):
I have some minor nits but overall looks good to me. We might want to consider adding in type annotations in this PR or a follow-up one.

kounelisagis:
> I have some minor nits but overall looks good to me. We might want to consider adding in type annotations in this PR or a follow-up one.

Thanks for your comments! We have a dedicated story specifically for implementing type hints (and use [mypy](https://mypy.readthedocs.io/)) across the entire project. It's always on my mind, but for this PR, I wanted to minimize changes compared to the Cython version.

PR #2071, opened 2024-09-23
Move `DenseArrayImpl` to pure Python

kounelisagis:
Leaving the `write_direct` function (renamed to `write_direct_dense`) in Cython since it contains some C API calls that have no direct counterparts in the C++ API. It will be moved in a future PR.

---

[sc-55313]


tiledb/dense_array.py
```diff
@@ -0,0 +1,785 @@
...
+
+        return result
+
+    def __repr__(self):
+        if self.isopen:
+            return "DenseArray(uri={0!r}, mode={1}, ndim={2})".format(
+                self.uri, self.mode, self.schema.ndim
+            )
+        else:
+            return "DenseArray(uri={0!r}, mode=closed)".format(self.uri)
```
nguyenv, line 113:
```suggestion
        if self.isopen:
            return f"DenseArray(uri={self.uri!r}, mode={self.mode}, ndim={self.schema.ndim})"
        else:
            return f"DenseArray(uri={self.uri!r}, mode=closed)"
```

tiledb/dense_array.py
```diff
@@ -0,0 +1,785 @@
...
+            attr_names.extend(self.schema.attr(a).name for a in attrs)
+
+        selection = index_as_tuple(selection)
+        idx = replace_ellipsis(self.schema.domain.ndim, selection)
+        if check_for_floats(selection):
+            warnings.warn(
+                "The use of floats in selection is deprecated. "
+                "It is slated for removal in 0.31.0.",
+                DeprecationWarning,
+            )
```
nguyenv, line 264: Deprecate here or in follow-up PR?

tiledb/dense_array.py
```diff
@@ -0,0 +1,785 @@
...
+
+        return out
+
+    def __setitem__(self, selection, val):
+        """Set / update dense data cells
+
+        :param tuple selection: An int index, slice or tuple of integer/slice objects,
+            specifiying the selected subarray region for each dimension of the DenseArray.
+        :param value: a dictionary of array attribute values, values must able to be converted to n-d numpy arrays.\
+            if the number of attributes is one, then a n-d numpy array is accepted.
```
nguyenv, line 358: nit: Match doc to arg name

```suggestion
    def __setitem__(self, selection, val):
        """Set / update dense data cells

        :param tuple selection: An int index, slice or tuple of integer/slice objects,
            specifiying the selected subarray region for each dimension of the DenseArray.
        :param val: a dictionary of array attribute values, values must able to be converted to n-d numpy arrays.\
            if the number of attributes is one, then a n-d numpy array is accepted.
```

nguyenv (approved):
Looks good! In parallel with the prior `SparseArrayImpl`, we will add in type annotations in a future PR

PR #2072, opened 2024-09-23
Move `PackedBuffer` and `pack_metadata_val` to pure Python, and wrap `tiledb_datatype_size` with pybind

kounelisagis:
[sc-55643]


nguyenv (approved):
lgtm!

PR #2076, opened 2024-09-30
Move `Array` to pure Python

kounelisagis:
This PR moves `Array` class from Cython to pure Python.

Attention is needed regarding the fact that the `Array` class in the pybind layer has breaking changes, as all functions and properties now begin with an underscore, similar to our other pybind classes.

---

[sc-54826]


tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+    def __init__(
+        self, uri, mode="r", key=None, timestamp=None, attr=None, ctx=None, **kwargs
+    ):
+        if not ctx:
+            ctx = default_ctx()
+
+        if "preloaded_array" not in kwargs:
+            self.array = preload_array(uri, mode, key, timestamp, ctx)
+        else:
+            self.array = kwargs["preloaded_array"]
```
nguyenv, line 324:
```suggestion
        self.array = kwargs.get("preloaded_array", preload_array(uri, mode, key, timestamp, ctx))
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+
+        # view on a single attribute
+        schema = self.array._schema()
+        if attr and not any(
+            attr == schema._attr(i)._name for i in range(schema._nattr)
+        ):
+            self.array._close()
+            raise KeyError("No attribute matching '{}'".format(attr))
+        else:
+            self.view_attr = attr if attr is not None else None
```
nguyenv, line 334: `None` check is superflorous.

```suggestion
            self.view_attr = attr
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        else:
+            self.array = kwargs["preloaded_array"]
+
+        # view on a single attribute
+        schema = self.array._schema()
+        if attr and not any(
+            attr == schema._attr(i)._name for i in range(schema._nattr)
+        ):
+            self.array._close()
+            raise KeyError("No attribute matching '{}'".format(attr))
```
nguyenv, line 332:
```suggestion
            raise KeyError(f"No attribute matching '{attr}'")
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        if "preloaded_array" not in kwargs:
+            self.array = preload_array(uri, mode, key, timestamp, ctx)
+        else:
+            self.array = kwargs["preloaded_array"]
+
+        # view on a single attribute
+        schema = self.array._schema()
+        if attr and not any(
+            attr == schema._attr(i)._name for i in range(schema._nattr)
+        ):
```
nguyenv, line 330:
```suggestion
        if attr is not None and not schema.has_attr(attr):
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+    def __repr__(self):
+        if self.isopen:
+            return "Array(type={0}, uri={1!r}, mode={2}, ndim={3})".format(
+                "Sparse" if self.schema.sparse else "Dense",
+                self.uri,
+                self.mode,
+                self.schema.ndim,
+            )
+        else:
+            return "Array(uri={0!r}, mode=closed)"
```
nguyenv, line 358:
```suggestion
            return f"Array(uri={self.uri!r}, mode=closed)"
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        return self.array.__capsule__()
+
+    def __repr__(self):
+        if self.isopen:
+            return "Array(type={0}, uri={1!r}, mode={2}, ndim={3})".format(
+                "Sparse" if self.schema.sparse else "Dense",
+                self.uri,
+                self.mode,
+                self.schema.ndim,
+            )
```
nguyenv, line 356:
```suggestion
            array_type = "Sparse" if self.schema.sparse else "Dense"
            return f"Array(type={array_type }, uri={self.uri!r}, mode={self.mode}, ndim={self.schema.ndim})"
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+                        by default uses the ArraySchema's associated context
+                        (*not* necessarily ``tiledb.default_ctx``).
+
+        """
+        if ctx and not isinstance(ctx, Ctx):
+            raise TypeError(
+                "tiledb.Array.create() expected tiledb.Ctx " "object to argument ctx"
+            )
+        if not ctx:
+            ctx = schema.ctx
```
nguyenv, line 378:
```suggestion
        if ctx is not None and not isinstance(ctx, Ctx):
            raise TypeError(
                "tiledb.Array.create() expected tiledb.Ctx object to argument ctx"
            )
        if ctx is None:
            ctx = schema.ctx
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        if issubclass(cls, DenseArrayImpl) and schema.sparse:
+            raise ValueError(
+                "Array.create `schema` argument must be a dense schema for DenseArray and subclasses"
+            )
+        if issubclass(cls, SparseArrayImpl) and not schema.sparse:
+            raise ValueError(
+                "Array.create `schema` argument must be a sparse schema for SparseArray and subclasses"
+            )
+
+        if key:
```
nguyenv, line 392:
```suggestion
        if key is not None:
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+            config = tiledb.Config(ctx.config())
+            config["sm.encryption_type"] = "AES_256_GCM"
+            config["sm.encryption_key"] = key
+            ctx = tiledb.Ctx(config)
+
+        if overwrite:
```
nguyenv, line 398:
```suggestion
        if overwrite is not None:
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+                    raise TypeError("Cannot overwrite non-local array.")
+            else:
+                warnings.warn("Overwrite set, but array does not exist")
+
+        lt.Array._create(ctx, uri, schema)
+
+    @classmethod
+    def load_typed(cls, uri, mode="r", key=None, timestamp=None, attr=None, ctx=None):
+        """Return a {Dense,Sparse}Array instance from a pre-opened Array (internal)"""
+        if not ctx:
```
nguyenv, line 420:
```suggestion
        if ctx is None:
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        if tmp_array._schema()._array_type == lt.ArrayType.SPARSE:
+            new_array_typed = SparseArrayImpl(
+                uri, mode, key, timestamp, attr, ctx, preloaded_array=tmp_array
+            )
+            return new_array_typed
+        else:
+            new_array_typed = DenseArrayImpl(
+                uri, mode, key, timestamp, attr, ctx, preloaded_array=tmp_array
+            )
+            return new_array_typed
```
nguyenv, line 437:
```suggestion
        if tmp_array._schema()._array_type == lt.ArrayType.SPARSE:
            return SparseArrayImpl(
                uri, mode, key, timestamp, attr, ctx, preloaded_array=tmp_array
            )
        else:
            return DenseArrayImpl(
                uri, mode, key, timestamp, attr, ctx, preloaded_array=tmp_array
            )
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+
+    _mode_to_query_type = {
+        "r": lt.QueryType.READ,
+        "w": lt.QueryType.WRITE,
+        "m": lt.QueryType.MODIFY_EXCLUSIVE,
+        "d": lt.QueryType.DELETE,
+    }
+    if mode not in _mode_to_query_type:
+        raise ValueError("TileDB array mode must be 'r', 'w', 'm', or 'd'")
+    query_type = _mode_to_query_type[mode]
```
nguyenv, line 275:
```suggestion
    _mode_to_query_type = {
        "r": lt.QueryType.READ,
        "w": lt.QueryType.WRITE,
        "m": lt.QueryType.MODIFY_EXCLUSIVE,
        "d": lt.QueryType.DELETE,
    }
    try:
        query_type = _mode_to_query_type[mode]
    except KeyError:
        raise ValueError(f"TileDB array mode must be one of {_mode_to_query_type.keys()}")
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        if self.view_attr:
+            return 1
+        else:
+            return self.schema.nattr
+
+    def view_attr(self):
+        """The view attribute of this array."""
+        if self.view_attr:
+            return self.view_attr
+        return None
```
nguyenv, line 534:
```suggestion
        return self.view_attr
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        """The shape of this array."""
+        return self.schema.shape
+
+    @property
+    def nattr(self):
+        """The number of attributes of this array."""
+        if self.view_attr:
+            return 1
+        else:
+            return self.schema.nattr
```
nguyenv, line 528:
```suggestion
        return 1 if self.view_attr is not None else self.schema.nattr
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        >>> tiledb.Array.delete_array(path)
+
+        >>> tiledb.array_exists(path)
+        False
+
+        """
+        if not ctx:
+            ctx = default_ctx()
+
+        lt.Array._delete_array(ctx, uri)
```
nguyenv, line 648:
```suggestion 
        lt.Array._delete_array(ctx or default_ctx(), uri)
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        >>> with tiledb.open(path, 'r') as A:
+        ...     A[:]
+        array([0., 0., 0., 0.])
+
+        """
+        if isinstance(uri, str):
+            if not ctx:
+                ctx = default_ctx()
+
+            lt.Array._delete_fragments(ctx, uri, timestamp_start, timestamp_end)
```
nguyenv, line 617:
```suggestion
            lt.Array._delete_fragments(ctx or default_ctx(), uri, timestamp_start, timestamp_end)
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        results = list()
+        dom = self.schema.domain
+
+        for dim_idx in range(dom.ndim):
+            dim_dtype = dom.dim(dim_idx).dtype
+
+            if self.array._non_empty_domain_is_empty(dim_idx, dim_dtype, self.ctx):
+                results.append(
+                    (None, None),
+                )
```
nguyenv, line 667:
```suggestion
                results.append((None, None))
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+            self.array._consolidate(self.ctx, fragment_uris, config)
+            return
+        elif timestamp is not None:
+            warnings.warn(
+                "The `timestamp` argument is deprecated; pass a list of "
+                "fragment URIs to consolidate with `fragment_uris`",
+                DeprecationWarning,
+            )
+
+            if not isinstance(timestamp, tuple) and len(timestamp) != 2:
```
nguyenv, line 740:
```suggestion
            if not isinstance(timestamp, tuple) and len(timestamp) != 2 and not (isinstance(timestamp[0], int) and isinstance(timestamp[1], int))::
```

tiledb/array.py
```diff
@@ -243,3 +251,761 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        from .main import PyQuery
+
+        q = PyQuery(self.ctx, self, ("",), (), 0, False)
+        q.set_serialized_query(serialized_query)
+        q.submit()
+
+        from collections import OrderedDict
+
+        results = OrderedDict()
+        results = q.results()
```
nguyenv, line 978:
```suggestion
        results = q.results()
```

tiledb/cc/array.cc
```diff
@@ -176,17 +279,20 @@ void init_array(py::module &m) {
...
+      .def("_has_metadata",
            [](Array &self, std::string &key) -> py::tuple {
              tiledb_datatype_t has_type;
              bool has_it = self.has_metadata(key, &has_type);
              return py::make_tuple(has_it, has_type);
            })
       .def("metadata_num", &Array::metadata_num)
-      .def("delete_array",
+      .def("delete_metadata",
+           [](Array &self, std::string &key) { self.delete_metadata(key); })
```
nguyenv, line 291: Methods to delete the metadata in the array are in here twice.

tiledb/cc/array.cc
```diff
@@ -84,20 +110,113 @@ void init_array(py::module &m) {
...
+               } else if (n_type.is(py::dtype::of<int16_t>())) {
+                 domain = new int16_t[2];
+               } else if (n_type.is(py::dtype::of<uint8_t>())) {
+                 domain = new uint8_t[2];
+               } else if (n_type.is(py::dtype::of<int8_t>())) {
+                 domain = new int8_t[2];
+               } else if (n_type.is(py::dtype::of<double>())) {
+                 domain = new double[2];
+               } else if (n_type.is(py::dtype::of<float>())) {
+                 domain = new float[2];
```
nguyenv, line 161: You're going to need to call a `delete[]` so that `domain` doesn't leak here.

nguyenv: I would suggest implementing a templated function along the lines of

```
template <typename T>
bool _non_empty_domain_is_empty_aux<T>(...) {
    int32_t is_empty = 0;
    T domain [2];
    ctx.handle_error(tiledb_array_get_non_empty_domain_from_index(
                   ctx.ptr().get(), self.ptr().get(), dim_idx, &domain,
                   &is_empty));
    return is_empty == 1;
}
```

tiledb/array.py
kounelisagis: `overwrite` is a boolean. I suggest keeping this condition unchanged.

tiledb/cc/array.cc
kounelisagis: Special thanks for this. It was my mistake for not noticing the memory leak here.

kounelisagis:
> This looks great! Moving `Array` from cython to pybind11 is diffcult work, but well worth the effort because it's going to improve a lot of things. The vast majority of these suggestions are actually just minor nits that I know existed in this code prior to this PR but just taking a chance to clean those up.
> 
> The only actual issue I see is that when `new` is called in the pybind11 code, there is not a paired `delete` call to deallocate the memory for `domain`, so we have a memory leak there.

Thanks for your comments @nguyenv! Really helpful and right on point!

nguyenv (approved):
Thanks! The changes look good to me! Maybe also wait for @ihnorton to approve (it sounds like we also want to check any performance regressions?)

tiledb/array.py
```diff
@@ -243,3 +251,748 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+        if issubclass(cls, SparseArrayImpl) and not schema.sparse:
+            raise ValueError(
+                "Array.create `schema` argument must be a sparse schema for SparseArray and subclasses"
+            )
+
+        if key is not None:
+            config = tiledb.Config(ctx.config())
+            config["sm.encryption_type"] = "AES_256_GCM"
+            config["sm.encryption_key"] = key
+            ctx = tiledb.Ctx(config)
```
teo-tsirpanis, line 391:
```suggestion
```

tiledb/array.py
```diff
@@ -243,3 +251,748 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+                        by default uses the ArraySchema's associated context
+                        (*not* necessarily ``tiledb.default_ctx``).
+
+        """
+        if ctx is not None and not isinstance(ctx, Ctx):
+            raise TypeError(
+                "tiledb.Array.create() expected tiledb.Ctx " "object to argument ctx"
+            )
+        if ctx is None:
+            ctx = schema.ctx
```
teo-tsirpanis, line 373:
```suggestion
        ctx = ctx or default_ctx()
```

We should not be using the internal context of "data objects" like array schemas.

tiledb/array.py
```diff
@@ -243,3 +251,748 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+            ctx = tiledb.Ctx(config)
+
+        if overwrite:
+            from .highlevel import object_type
+            from .vfs import VFS
+
+            if object_type(uri) == "array":
+                if uri.startswith("file://") or "://" not in uri:
+                    try:
+                        VFS().remove_dir(uri)
```
teo-tsirpanis, line 400:
```suggestion
                        VFS(ctx).remove_dir(uri)
```

tiledb/array.py
```diff
@@ -243,3 +251,748 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+            raise tiledb.TileDBError(
+                "cannot consolidate array opened in readonly mode (mode='r')"
+            )
+
+        if config is None:
+            config = tiledb.Config(self.ctx.config())
+
+        if key is not None:
+            config["sm.encryption_type"] = "AES_256_GCM"
+            config["sm.encryption_key"] = key
```
teo-tsirpanis, line 704:
```suggestion
        if key is not None:
            config = tiledb.Config(config or self.ctx.config())
            config["sm.encryption_type"] = "AES_256_GCM"
            config["sm.encryption_key"] = key
```

This will avoid needlessly allocating a config, and modifying the user-provided config if encryption is used.

tiledb/array.py
```diff
@@ -243,3 +251,748 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+                )
+
+            if timestamp[0] is not None:
+                config["sm.consolidation.timestamp_start"] = timestamp[0]
+            if timestamp[1] is not None:
+                config["sm.consolidation.timestamp_end"] = timestamp[1]
+
+        self.array._consolidate(self.ctx, config)
+
+    def upgrade_version(self, config=None):
```
teo-tsirpanis, line 740: In the future this should be updated to accept a `self_or_uri` argument, like we did for `delete_fragments`. Same with `consolidate` above, but that might need to be deprecated because we already have `tiledb.consolidate`.

teo-tsirpanis (changes requested):

tiledb/array.py
kounelisagis: Story opened!

tiledb/array.py
kounelisagis: This is necessary to prevent the following error:

`E       tiledb.cc.TileDBError: [TileDB::Array] Error: ArrayDirectory: GenericTileIO: Error reading generic tile; tile is encrypted with NO_ENCRYPTION but given key is for AES_256_GCM`

kounelisagis: we need to create a new `Ctx` object with the modified `config`.

tiledb/cc/array.cc
```diff
@@ -30,45 +59,46 @@ void init_array(py::module &m) {
...
+      .def_property_readonly("_open_timestamp_end", &Array::open_timestamp_end)
+      .def("_set_config", &Array::set_config)
+      .def("_config", &Array::config)
+      .def("_close", &Array::close)
+      .def("_get_enumeration",
+           [](Array &self, const Context &ctx, const std::string &attr_name) {
+             return ArrayExperimental::get_enumeration(ctx, self, attr_name);
+           })
+      .def("_consolidate",
            [](Array &self, const Context &ctx, Config *config) {
```
teo-tsirpanis, line 83: The `consolidate` functions should accept a URI instead of an array object.

ihnorton: IIUC you mean this signature instead?
https://github.com/TileDB-Inc/TileDB/blob/034ed809077f6b804583438387fe248aabca230a/tiledb/sm/cpp_api/array.h#L588-L594

teo-tsirpanis: Yes.

kounelisagis: Good catch, @teo-tsirpanis! There was no point in accepting an array object as an argument since the only thing we're using it for is the uri.

teo-tsirpanis (approved):
Thanks!

ihnorton (commented):
A couple comments before merging:
1. I'd like to keep `tests/tiledb/test_libtiledb.py` named as-is for now to make reviewing/history clearer (it looks like the diff is very small other than the rename)
2. Let's discuss testing strategy for dependencies before releasing

kounelisagis:
> A couple comments before merging:
> 
> 1. I'd like to keep `tests/tiledb/test_libtiledb.py` named as-is for now to make reviewing/history clearer (it looks like the diff is very small other than the rename)
> 2. Let's discuss testing strategy for dependencies before releasing


Thanks for the comment!
1. Makes total sense. Let's keep it as it is, at least for this PR.
2. No releases will be made unless we test the dependencies.

PR #2079, opened 2024-09-30
Drop Python 3.8 support

kounelisagis:
This PR removes support for TileDB-Py on Python 3.8. Some reasons are:
- Python 3.8 reaches end-of-life in October 2024: https://devguide.python.org/versions/#supported-versions
- Continuing support for Python 3.8 has become untenable because NumPy [no longer supports it](https://pypi.org/project/numpy/1.26.4/#files)
- Reducing the number of wheels will decrease overall maintenance costs

---

[sc-48532]

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -96,7 +96,7 @@ jobs:
           - macos-14
           - windows-2022
           - ubuntu-22.04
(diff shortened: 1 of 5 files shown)


nguyenv (approved):
Nice that we could get rid of all those 3.8 conditionals now.

ihnorton (approved):

PR #2083, opened 2024-10-09
Add test for dropping a fixed attribute and adding it back as var-sized

kounelisagis:
Following https://github.com/TileDB-Inc/TileDB/pull/5321 and https://github.com/TileDB-Inc/TileDB/pull/5362, let's add a TileDB-Py test.

---

[sc-57039]

### DIFF
--- a/tiledb/tests/test_schema_evolution.py
+++ b/tiledb/tests/test_schema_evolution.py
@@ -227,3 +227,39 @@ def test_schema_evolution_extend_check_bad_type():
     with pytest.raises(tiledb.TileDBError):
         enmr.extend([1, 2, 3])
     enmr.extend([True, False])
+
+
+@pytest.mark.skipif(
+    tiledb.libtiledb.version() < (2, 27),
+    reason="Dropping a fixed-sized attribute and adding it back"
+    "as a var-sized attribute is not supported in TileDB < 2.27",
+)
+def test_schema_evolution_drop_fixed_attribute_and_add_back_as_var_sized(tmp_path):
+    ctx = tiledb.default_ctx()
+    uri = str(tmp_path)
+    attrs = [
+        tiledb.Attr(name="a", dtype=np.int32),
+        tiledb.Attr(name="b", dtype=np.int32),
+    ]
+    dims = [tiledb.Dim(domain=(1, 10), dtype=np.int32)]
+    domain = tiledb.Domain(*dims)
+    schema = tiledb.ArraySchema(domain=domain, attrs=attrs, sparse=False)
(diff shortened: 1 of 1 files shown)


kounelisagis:
Note for myself: A larger random data generation is needed.

tiledb/tests/test_schema_evolution.py
```diff
@@ -227,3 +227,36 @@ def test_schema_evolution_extend_check_bad_type():
...
+    se.array_evolve(uri)
+
+    se = tiledb.ArraySchemaEvolution(ctx)
+    newattr = tiledb.Attr("a", dtype="S0", var=True)
+    se.add_attribute(newattr)
+    se.array_evolve(uri)
+
+    with tiledb.open(uri) as A:
+        res = A[:]
+        assert_array_equal(res["a"], data)
```
ihnorton, line 261: Shouldn't the result be empty here?

bekadavis9: +1

kounelisagis: Right! Test was modified.

tiledb/tests/test_schema_evolution.py
```diff
@@ -227,3 +227,58 @@ def test_schema_evolution_extend_check_bad_type():
...
+        assert A.schema.has_attr("a")
+        assert A.schema.attr("a").dtype == "S"
+        assert A.schema.attr("b").dtype == np.int32
+
+    # add new data to the array
+    new_data = np.array(
+        ["tiledb-string-n.{}".format(i) for i in range(1, 11)], dtype="S"
+    )
+    with tiledb.open(uri, "w") as A:
+        A[:] = {"a": new_data, "b": original_data}
```
bekadavis9, line 278: You don't need to write to `b` again; the original data will still be there. I would also discourage writing to `a` here. This test should validate that `a != original_data`. The expected behavior is that the original (`int`) data is lost and the new `char` data is uninitialized. Checking that each value `== b'\x80'` is sufficient, as long as a comment accompanies the check to explain the behavior.

bekadavis9: Alternatively, you could change the second-to-last line of this test to some `assert_array_not_equal(res["a"], original_data)`. 
I don't know what specific check could be used - that's your wheelhouse :)

ihnorton: we have to write `b` again because single-attribute writes are not supported.

ihnorton (approved):
LGTM after addressing the fill value check request from @bekadavis9

PR #2084, opened 2024-10-10
Fix skipif condition for `test_save_and_open_array_from_cloud`

kounelisagis:
This PR makes the skip-if condition for `test_save_and_open_array_from_cloud` stricter by requiring both a CI environment and each environment variable that it utilizes.

cc @jdblischak

### DIFF
--- a/tiledb/tests/test_cloud.py
+++ b/tiledb/tests/test_cloud.py
@@ -15,7 +15,10 @@ s3_bucket = os.getenv("S3_BUCKET")
 
 
 @pytest.mark.skipif(
-    not os.getenv("CI") and tiledb_token is None,
+    os.getenv("CI") == None
+    or tiledb_token == None
+    or tiledb_namespace == None
+    or s3_bucket == None,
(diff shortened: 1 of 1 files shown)


tiledb/tests/test_cloud.py
```diff
@@ -15,7 +15,10 @@
 
 
 @pytest.mark.skipif(
-    not os.getenv("CI") and tiledb_token is None,
+    os.getenv("CI") == None
```
ihnorton, line 18: I'm pretty sure this will be set in the nightly setup?

kounelisagis: Yes, but the others will not be set.

jdblischak: The env var `CI` is set to `true` by default on GitHub Actions

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

jdblischak (approved):

ihnorton (approved):

PR #2085, opened 2024-10-11
Extend the `GroupMetadata` functionality to support NumPy arrays

kounelisagis:
Even though the `GroupMetadata` and `Metadata` (used by the `Array` class) classes  utilize the same underlying code (similar C++ APIs), it was found that `GroupMetadata` does not support NumPy arrays and also lacks tests for handling such data.

This PR extends the functionality of `GroupMetadata` to handle the same data as `Metadata` and includes tests for the changes.

The modifications were made so that `GroupMetadata` uses only pybind11 and Python, without any Cython. In a future follow-up PR, the Cython `Metadata` class will be removed, and `GroupMetadata` will be renamed to `Metadata`. Additionally, the pybind11 code will be moved to a separate `metadata.cc` file. The new class will be shared across Array and Group classes.

---

[sc-57551]

### DIFF
--- a/tiledb/cc/group.cc
+++ b/tiledb/cc/group.cc
@@ -27,9 +27,9 @@ void put_metadata_numpy(Group &group, const std::string &key, py::array value) {
 
   py::size_t ncells = get_ncells(value.dtype());
   if (ncells != 1)
-    throw py::type_error(py::str(py::str("Unsupported dtype '") +
-                                 py::str(value.dtype()) +
-                                 py::str("' for metadata")));
+    throw py::type_error("Unsupported dtype '" +
+                         std::string(py::str(value.dtype())) +
+                         "' for metadata");
 
   auto value_num = is_tdb_str(value_type) ? value.nbytes() : value.size();
   group.put_metadata(key, value_type, value_num,
@@ -66,20 +66,12 @@ py::object unpack_metadata_val(tiledb_datatype_t value_type, uint32_t value_num,
     throw TileDBError("internal error: unexpected value_num==0");
 
   if (value_type == TILEDB_STRING_UTF8) {
-    if (value_ptr == nullptr)
-      return py::str();
-    else {
-      return py::str(value_ptr, value_num);
-    }
+    return value_ptr == nullptr ? py::str() : py::str(value_ptr, value_num);
   }
 
   if (value_type == TILEDB_BLOB || value_type == TILEDB_CHAR ||
       value_type == TILEDB_STRING_ASCII) {
-    if (value_ptr == nullptr)
-      return py::bytes();
-    else {
-      return py::bytes(value_ptr, value_num);
-    }
+    return value_ptr == nullptr ? py::bytes() : py::bytes(value_ptr, value_num);
   }
 
   if (value_ptr == nullptr)
--- a/tiledb/cc/group.cc
+++ b/tiledb/cc/group.cc
@@ -22,16 +22,14 @@ void put_metadata_numpy(Group &group, const std::string &key, py::array value) {
     throw py::type_error(e.what());
   }
 
-  if (is_tdb_str(value_type) && value.size() > 1)
-    throw py::type_error("array/list of strings not supported");
-
-  py::buffer_info value_buffer = value.request();
-  if (value_buffer.ndim != 1)
+  if (value.ndim() != 1)
     throw py::type_error("Only 1D Numpy arrays can be stored as metadata");
 
   py::size_t ncells = get_ncells(value.dtype());
   if (ncells != 1)
-    throw py::type_error("Unsupported dtype for metadata");
+    throw py::type_error(py::str(py::str("Unsupported dtype '") +
+                                 py::str(value.dtype()) +
+                                 py::str("' for metadata")));
 
   auto value_num = is_tdb_str(value_type) ? value.nbytes() : value.size();
   group.put_metadata(key, value_type, value_num,
@@ -40,8 +38,10 @@ void put_metadata_numpy(Group &group, const std::string &key, py::array value) {
 
 void put_metadata(Group &group, const std::string &key,
                   tiledb_datatype_t value_type, uint32_t value_num,
-                  const char *value) {
-  group.put_metadata(key, value_type, value_num, value);
+                  py::buffer &value) {
+
+  py::buffer_info info = value.request();
+  group.put_metadata(key, value_type, value_num, info.ptr);
 }
 
 bool has_metadata(Group &group, const std::string &key) {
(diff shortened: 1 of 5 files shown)


tiledb/cc/group.cc
```diff
@@ -22,16 +22,14 @@ void put_metadata_numpy(Group &group, const std::string &key, py::array value) {
...
-  if (value_buffer.ndim != 1)
+  if (value.ndim() != 1)
     throw py::type_error("Only 1D Numpy arrays can be stored as metadata");
 
   py::size_t ncells = get_ncells(value.dtype());
   if (ncells != 1)
-    throw py::type_error("Unsupported dtype for metadata");
+    throw py::type_error(py::str(py::str("Unsupported dtype '") +
+                                 py::str(value.dtype()) +
+                                 py::str("' for metadata")));
```
nguyenv, line 32:
```suggestion
    throw py::type_error(
        "Unsupported dtype '" + std::string(py::str(value.dtype())) + "' for metadata"
    );
```

tiledb/cc/group.cc
```diff
@@ -60,28 +60,110 @@ std::string get_key_from_index(Group &group, uint64_t index) {
...
+  if (value_num == 0)
+    throw TileDBError("internal error: unexpected value_num==0");
+
+  if (value_type == TILEDB_STRING_UTF8) {
+    if (value_ptr == nullptr)
+      return py::str();
+    else {
+      return py::str(value_ptr, value_num);
+    }
+  }
```
nguyenv, line 74:
```suggestion
  if (value_type == TILEDB_STRING_UTF8) {
     return value_ptr == nullptr ? py::str() : py::str(value_ptr, value_num);
  }
```

tiledb/cc/group.cc
```diff
@@ -60,28 +60,110 @@ std::string get_key_from_index(Group &group, uint64_t index) {
...
+
+  if (value_type == TILEDB_BLOB || value_type == TILEDB_CHAR ||
+      value_type == TILEDB_STRING_ASCII) {
+    if (value_ptr == nullptr)
+      return py::bytes();
+    else {
+      return py::bytes(value_ptr, value_num);
+    }
+  }
```
nguyenv, line 83:
```suggestion
  if (value_type == TILEDB_BLOB || value_type == TILEDB_CHAR ||
      value_type == TILEDB_STRING_ASCII) {
      return value_ptr == nullptr ? py::bytes() : py::bytes(value_ptr, value_num);
  }
```

tiledb/group.py
```diff
@@ -112,19 +113,19 @@ def __setitem__(self, key: str, value: GroupMetadataValueType):
...
-                put_metadata(key, lt.DataType.BLOB, len(value), value)
-            elif isinstance(value, str):
-                value = value.encode("UTF-8")
-                put_metadata(key, lt.DataType.STRING_UTF8, len(value), value)
-            elif isinstance(value, (list, tuple)):
-                put_metadata(key, np.array(value))
+                flat_value = value.ravel()
+                put_metadata(f"{Group._NP_DATA_PREFIX}{key}", flat_value)
+                if value.shape != flat_value.shape:
+                    self.__setitem__(f"{Group._NP_SHAPE_PREFIX}{key}", value.shape)
```
nguyenv, line 119: Can you document what this code does here? If it is not a 1D array, then store the actual shape associated with the values? Also, does the code need to be called recursively or could it use `put_metadata`?

kounelisagis: Added more comments. We need the recursion because `value.shape` has to go through `pack_metadata_val` (avoiding code duplication). Since `value.shape` will always be [a tuple of ints](https://numpy.org/doc/2.0/reference/generated/numpy.shape.html), the else statement will be executed when `__setitem__` is given a numpy shape. I think this solution makes sense because the shape is like 'new' metadata.

tiledb/cc/group.cc
```diff
@@ -22,16 +22,14 @@ void put_metadata_numpy(Group &group, const std::string &key, py::array value) {
     throw py::type_error(e.what());
   }
 
-  if (is_tdb_str(value_type) && value.size() > 1)
```
ihnorton, line 25: We can't serialize this.

kounelisagis: After adding three tests, arrays of strings weren't handled the same way as for Array metadata.

These cases were supposed to be caught here: https://github.com/TileDB-Inc/TileDB-Py/blob/0be88d46010dd8f4f2087081b4d83a6acb48bb16/tiledb/cc/group.cc#L30-L32

`get_ncells` didn’t exhibit the same behavior as its Python counterpart: https://github.com/TileDB-Inc/TileDB-Py/blob/0be88d46010dd8f4f2087081b4d83a6acb48bb16/tiledb/datatypes.py#L47-L55

After updating `get_ncells`, only single-character numpy arrays are now allowed, similar to the handling of Array metadata.

nguyenv (approved):
Thanks for making the corrections to match this to the Array metadata. I think you added good test cases to cover what was missing before.

tiledb/tests/test_group.py
```diff
@@ -636,3 +689,76 @@ def test_consolidation_and_vac_no_config(self):
...
+        grp.meta["abc"] = "xyz"
+        grp.close()
+
+        grp = tiledb.Group(uri, "r")
+        assert grp.meta["abc"] == "xyz"
+        grp.meta.dump()
+        assert_captured(capfd, "Type: DataType.STRING_UTF8")
+        grp.close()
+
+    def test_array_or_list_of_strings_metadata_error(self):
```
nguyenv, line 710: Actually, was wondering how this handles empty strings like

`np.array(["1", "2", "3", "", "f", "o", "o"], dtype="U1")` (in the middle)
`np.array(["1", "2", "3", "f", "o", "o", ""], dtype="U1")` (trailing)
`np.array([""], dtype="U1")` (by itself)

kounelisagis: Thanks for your comment! It's a good one!

The way this works is by assigning a byte to each string, even if it's empty. ASCII ("S") strings are a bit easier to visualize. So, when you assign the following array as metadata: `np.array(["1", "2", "3", "", "f", "o", "o"], dtype="S1")`, what you're essentially sending to the `Group::put_metadata` C++ API is `b'123\x00foo'`. As you can see, there is a null byte (`\x00`) which represents the empty string.

Similarly, `np.array([""], dtype="S1")` sends `b'\x00'`.

For `np.array(["1", "2", "3", "", "f", "o", "o"], dtype="U1")`, we send `b'1\x00\x00\x002\x00\x00\x003**\x00\x00\x00\x00**\x00\x00\x00f\x00\x00\x00o\x00\x00\x00o\x00\x00\x00'`. 

Here, null bytes are also used as padding because the characters can be represented in ASCII. The empty string is represented by null bytes, which you can see between the `**` symbols.

I hope this answers your question.

nguyenv: Awesome - thanks for the explaination!

PR #2086, opened 2024-10-11
Make default value for `vfs.s3.region` in test conditional

kounelisagis:
Following https://github.com/TileDB-Inc/TileDB/pull/5317, the default value of `vfs.s3.region` has been changed to empty. This change allows AWS to select the default region from sources such as environment variables, profile configuration, and instance metadata.

This PR modifies some assertions that check the default value of `vfs.s3.region` when related environment variables are present.

cc @jdblischak

### DIFF
--- a/tiledb/tests/test_fixes.py
+++ b/tiledb/tests/test_fixes.py
@@ -196,7 +196,7 @@ class FixesTest(DiskTestCase):
     def test_sc23827_aws_region(self):
         # Test for SC-23287
         # The expected behavior here for `vfs.s3.region` is:
-        # - default to 'us-east-1' if no environment variables are set
+        # - default to '' if no environment variables are set
         # - empty if AWS_REGION or AWS_DEFAULT_REGION is set (to any value)
 
         def get_config_with_env(env, key):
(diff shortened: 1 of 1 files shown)


tiledb/tests/test_fixes.py
```diff
@@ -209,9 +209,17 @@ def get_config_with_env(env, key):
...
+        assert (
+            get_config_with_env({"AWS_DEFAULT_REGION": "us-east-1"}, "vfs.s3.region")
+            == "us-east-1"
+        )
+
         assert get_config_with_env({"AWS_REGION": ""}, "vfs.s3.region") == ""
+        assert (
+            get_config_with_env({"AWS_REGION": "us-east-1"}, "vfs.s3.region")
+            == "us-east-1"
+        )
```
teo-tsirpanis, line 222: As an alternative to removing the test, can we condition the new assertions on `libtiledb.version()` being >= 2.27?

teo-tsirpanis (commented):
This test is too reliant on the Core's implementation details to IMHO include in the Python API's test suite. It should be moved to the Core and removed from here.

tiledb/tests/test_fixes.py
kounelisagis: You are right Teo. Thanks for the suggestion!

PR #2087, opened 2024-10-11
Make error message for non-existing `Enumeration` in test conditional

kounelisagis:
cc @jdblischak

### DIFF
--- a/tiledb/tests/test_enumeration.py
+++ b/tiledb/tests/test_enumeration.py
@@ -93,7 +93,7 @@ class EnumerationTest(DiskTestCase):
             with self.assertRaises(tiledb.TileDBError) as excinfo:
                 assert A.enum("enmr3") == []
             assert (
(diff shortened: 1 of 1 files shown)


ihnorton (approved):

tiledb/tests/test_enumeration.py
```diff
@@ -93,7 +93,7 @@ def test_array_schema_enumeration(self):
             with self.assertRaises(tiledb.TileDBError) as excinfo:
                 assert A.enum("enmr3") == []
             assert (
-                "ArraySchema: Unable to check if unknown enumeration is loaded. No enumeration named 'enmr3'."
+                "Array: Unable to get enumeration; Enumeration 'enmr3' does not exist."
```
teo-tsirpanis, line 96: Until 2.27 releases we could use a different expected error message depending on `libtiledb.version()`, and switch to only the new one afterwards.

kounelisagis: You are right Teo. Thanks for the suggestion!

PR #2088, opened 2024-10-11
Handle removal of 'StorageManager' from stats dumps in 2.27

kounelisagis:
Following https://github.com/TileDB-Inc/TileDB/pull/5297, this PR removes the deprecated `StorageManager` word when accessing stats dump.

cc @jdblischak

### DIFF
--- a/tiledb/tests/test_fixes.py
+++ b/tiledb/tests/test_fixes.py
@@ -161,9 +161,15 @@ class FixesTest(DiskTestCase):
         with tiledb.open(uri) as A:
             tiledb.stats_enable()
             A[:]
-            assert """"Context.Query.Reader.loop_num": 1""" in tiledb.stats_dump(
-                print_out=False
-            )
+
+            stats_dump_str = tiledb.stats_dump(print_out=False)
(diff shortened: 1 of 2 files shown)


ihnorton (approved):

tiledb/tests/test_fixes.py
```diff
@@ -161,9 +161,8 @@ def test_sc16301_arrow_extra_estimate_dense(self):
         with tiledb.open(uri) as A:
             tiledb.stats_enable()
             A[:]
-            assert (
-                """"Context.StorageManager.Query.Reader.loop_num": 1"""
-                in tiledb.stats_dump(print_out=False)
+            assert """"Context.Query.Reader.loop_num": 1""" in tiledb.stats_dump(
```
teo-tsirpanis, line 164: This test is too reliant on the Core's implementation details to IMHO include in the Python API's test suite. It should be moved to the Core and removed from here. Also noting that the test is currently skipped.

ihnorton: I agree the message is core-internal, but the behavior test is also closely tied to TileDB-Py's query management logic, so IMO we should leave it here for now.

PR #2091, opened 2024-10-15
Add CI to test against the TileDB core library built from source with a custom version

dudoslav:
There are several PRs that we need to test against a branch of libtiledb, most commonly the `dev` branch, as they implement features/fixes we want to test before the release of the libtiledb binaries.

This PR implements a workflow to test against the TileDB core library built from source using a custom version.

---

[sc-57260]


kounelisagis:
I removed the `pull_request` trigger. As I wrote in the commit description, it doesn't make much sense because we will have failing jobs while libtiledb development is ongoing. We only want it for specific PRs.

PR #2098, opened 2024-10-23
Handle removal of 'StorageManager' from stats dumps in 2.27 - Part 2

kounelisagis:
Forgot about this in https://github.com/TileDB-Inc/TileDB-Py/pull/2088.

cc @jdblischak


teo-tsirpanis (approved):
Thanks!

PR #2099, opened 2024-10-24
Merge `Metadata` and `GroupMetadata` into one Python class

kounelisagis:
Following https://github.com/TileDB-Inc/TileDB-Py/pull/2085, this PR removes the Cython `Metadata` class and renames Python's `GroupMetadata` class to `Metadata`. The new `Metadata` class is now shared between the `Array` and `Group` classes, as both classes now have a `self._meta` property that is an instance of this class.
Additionally, the common pybind11 code has been moved to a separate `metadata.cc` file, within an adapter templated class that works for both the `Array` and `Group` classes.

CI: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/11503873688


kounelisagis:
After further consideration, I believe this change will break backward compatibility. More testing is needed, specifically writing metadata with older tiledb-py versions and reading it with this version, and vice versa.

kounelisagis:
In https://github.com/TileDB-Inc/TileDB-Py/pull/2099, backward compatibility has been fixed.

After rebasing the [pybind11-full](https://github.com/TileDB-Inc/TileDB-Py/tree/pybind11-full) branch onto the `dev` branch, and then rebasing the current branch onto `pybind11-full`, this PR is now ready for review.

nguyenv (approved):
This looks great! Using a template for the `MetadataAdapter` class so that it can take both arrays and groups is really clever.

tiledb/cc/metadata.h
```diff
@@ -0,0 +1,184 @@
...
+#include <pybind11/pytypes.h>
+#include <pybind11/stl.h>
+
+#include "common.h"
+
+using namespace tiledb;
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+template <typename T> class MetadataAdapter {
```
ihnorton, line 18: Nice!

PR #2100, opened 2024-10-24
Revert "Do not run CI in PR twice"

kounelisagis:
TileDB-Inc/TileDB-Py#2082 might have fixed the initial issue, but it causes PRs that we want to merge into non-dev branches to not run CI at all. For example: https://github.com/TileDB-Inc/TileDB-Py/pull/2099.

If there is no other solution, we should revert.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,6 @@
 name: TileDB Python CI
 
-on: 
-  push:
-    branches: [dev]
-  pull_request:
-    branches: [dev]
-  workflow_dispatch:
+on: [push, pull_request, workflow_dispatch]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,11 +1,6 @@
 name: TileDB Python CI
 
-on: 
-  push:
-    branches: [dev]
-  pull_request:
-    branches: [dev]
```
dudoslav, line 7: I think that we should just remove this condition for `pull_request` but leave other untoched

dudoslav: After some consideration and a small chat with Agis I have decided that sometimes it is also necessary to run CI on every branch, even without PR.

dudoslav (approved):
:unicorn:

PR #2103, opened 2024-10-30
Change `np.bool` to `np.bool_` - Daily Tests fix

kounelisagis:
Should fix https://github.com/TileDB-Inc/TileDB-Py/actions/runs/11587357551

Triggered daily tests workflow: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/11591425398


nguyenv (approved):
ty!

PR #2106, opened 2024-11-05
Move `Query` to pure Python

kounelisagis:
[sc-14915]


tiledb/query.py
```diff
@@ -14,20 +19,272 @@ def __init__(
...
+            self._dims = False
+        elif dims != None and dims != True:
+            domain = array.schema.domain
+            for dname in dims:
+                if not domain.has_dim(dname):
+                    raise TileDBError(f"Selected dimension does not exist: '{dname}'")
+            self._dims = dims
+        elif coords == True or dims == True:
+            domain = array.schema.domain
+            self._dims = [domain.dim(i).name for i in range(domain.ndim)]
```
nguyenv, line 65: I know this code is pre-existed and you just copied it over, but I really want to see this simplied. I think 

> `dims != None and dims != True` == 
> `dims != None and dims == False` == 
> `dims == False` 

which is already accounted for in the first conditional block. 

```suggestion
        if dims == False:
            self._dims = False
        elif coords or dims: 
            domain = array.schema.domain
            self._dims = [domain.dim(i).name for i in range(domain.ndim)]
```

tiledb/query.py
```diff
@@ -14,20 +19,272 @@ def __init__(
...
         ctx: Ctx = None,
+        attrs=None,
+        cond=None,
+        dims=None,
+        coords=False,
+        index_col=True,
+        order=None,
+        use_arrow=None,
+        return_arrow=False,
+        return_incomplete=False,
```
nguyenv, line 30: I think in general if we can set a specific default value for the argument here and avoid `None` that it would make the code easier to follow and less verbose.

tiledb/query.py
nguyenv: Oh wait I'm confused. `dims` is `Optional[bool | Sequence[str]`?

kounelisagis: yes

kounelisagis: I don't think you code simplification is possible.

kounelisagis: I added my own one. Checking that dimension is present seems like something we need.

tiledb/query.py
kounelisagis: Yes, did that. Good idea.

nguyenv (approved):
Thanks for making the changes. I think your conditional refactor reads much easier than what was there previously. The type hints are extremely useful here since `dims` and `attrs` being both `bool` and `Sequence[str]` was not very obvious.

PR #2107, opened 2024-11-06
Enable Python 3.13

kounelisagis:
Since October 2024, Python 3.13 has been in "bugfix" mode (also known as "maintenance" or "stable" mode): https://devguide.python.org/versions/#supported-versions.

To add support for TileDB-Py, an upgrade to the latest version of `cibuildwheel` was required, as it supports Python 3.13 and CPython 3.13. (More info: https://github.com/pypa/cibuildwheel/releases)

Testing:
- Wheels workflow: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/11796573304
- Tested locally (manylinux_x86_64) on a fresh conda env with Python 3.13 and works fine!

---

[sc-57688]

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -46,7 +46,7 @@ jobs:
           brew install automake pkg-config ninja llvm
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.18.1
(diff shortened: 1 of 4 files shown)


dudoslav (approved):
:cactus:

PR #2109, opened 2024-11-15
Factor `write_direct_dense` out of Cython

kounelisagis:
This PR integrates the `write_direct_dense` functionality directly into Python's `DenseArray::write_direct`, using only the pybind11/C++ bindings.

<del>Regarding test changes:</del>
- <del>For `tiledb/tests/test_hypothesis.py::test_bytes_npdf[np]`, the change from `uri = "mem://" + checked_path.path()` to `uri = checked_path.path()` was necessary to make it work. This is because, for example, a URI could be: `mem:///var/folders/gw/5rx7wg1d6397n6nwy86qjf7c0000gn/T/tiledb-disktestcasepvi959vz/tmphgkelovb`. The issue seems to be the use of `///`. I guess that speed was the primary concern for using `mem://`.</del>
- <del>For `tiledb/tests/test_libtiledb.py::test_fixed_string`, a change in the dtype of the numpy array was required. This is because the `np_to_tdb_dtype` function (see [here](https://github.com/TileDB-Inc/TileDB-Py/blob/1517565b693bc84dd07b4851c145c6bb9dc714e7/tiledb/cc/common.cc#L150)) cannot handle numpy arrays with multiple dtypes. This issue is expected to arise elsewhere in the codebase wherever this function is called with numpy arrays containing multiple dtypes. Addressing this limitation could be implemented in a future PR, as I believe it falls outside the scope of this one.</del>


tiledb/tests/test_hypothesis.py
```diff
@@ -20,7 +20,7 @@
 def test_bytes_npdf(checked_path, mode, data):
     start = time.time()
 
-    uri = "mem://" + checked_path.path()
+    uri = checked_path.path()
```
ihnorton, line 23: I think we might have an issue here with using the same context consistently for array operations, because `mem://` state is attached to the context.

kounelisagis: > I think we need to dig in to this one more, because it doesn't exactly have multiple dtypes -- the dtype is a record type.

The `DataType` class supports that feature, and I used it.

> I think we might have an issue here with using the same context consistently for array operations, because `mem://` state is attached to the context.

No changes related to this test anymore.

ihnorton (approved):
LGTM, would like to get @nguyenv review as well.

tiledb/dense_array.py
```diff
@@ -601,9 +601,104 @@ def __array__(self, dtype=None, **kw):
...
+            raise ValueError("cannot write_direct to a multi-attribute DenseArray")
+        if not array.flags.c_contiguous and not array.flags.f_contiguous:
+            raise ValueError("array is not contiguous")
+
+        try:
+            use_global_order = (
+                self.ctx.config().get("py.use_global_order_1d_write") == "true"
+            )
+        except KeyError:
+            use_global_order = False
```
nguyenv, line 630:
```suggestion
        use_global_order = (
            self.ctx.config().get("py.use_global_order_1d_write", "false") == "true"
        )
```

nguyenv (approved):
LGTM

PR #2110, opened 2024-11-18
Implement serialization functionality for `Config` and `VFS` objects

kounelisagis:
Also removes unnecessary `PatchedConfig` and `PatchedCtx`.

---

[sc-55042]


teo-tsirpanis (changes requested):
We should support pickling in `Config` as well.

teo-tsirpanis (approved):
Thanks!

PR #2112, opened 2024-11-19
Change default value of `TILEDB_REMOVE_DEPRECATIONS` to `OFF`

kounelisagis:
Following https://github.com/TileDB-Inc/conda-forge-nightly-controller/pull/152, it was decided that nightlies should be the place where we check for deprecations. The goal is to eliminate the use of deprecations over time. To enhance the user experience, TileDB-Py will make use of deprecated functions when necessary until their usage is replaced.


johnkerl:
Thanks @kounelisagis !

jdblischak (approved):

ihnorton (approved):

Issue #2005, opened 2024-07-11
Provide Linux wheels for aarch64


selimnairb:
Hello,

Would it be possible to start providing Linux wheels for aarch64? These aarch64 machines are often much cheaper to lease on cloud services, and it appears to be non-trivial to build the wheels via pip install. Thanks.

kounelisagis:
Hello Brian,

Thank you for reaching out. We will evaluate whether it makes sense for us to provide wheels for Linux aarch64 at that point of time. In the meantime, could you let us know about any errors/problems you encounter when installing from the source distribution?

Thanks!

---
[sc-50707]

ihnorton:
Thanks for opening the issue - also wanted to note that we do currently provide aarch64 packages via conda-forge: https://anaconda.org/conda-forge/tiledb/files

selimnairb:
> Thanks for opening the issue - also wanted to note that we do currently provide aarch64 packages via conda-forge: https://anaconda.org/conda-forge/tiledb/files

Thank you, this is good to know for others. However, then deploying applications to containers, I usually avoid conda if I can manage as it can greatly increase image sizes.

selimnairb:
> Hello Brian,
> 
> Thank you for reaching out. We will evaluate whether it makes sense for us to provide wheels for Linux aarch64 at that point of time. In the meantime, could you let us know about any errors/problems you encounter when installing from the source distribution?
> 
> Thanks!
> 
> [[sc-50707](https://app.shortcut.com/tiledb-inc/story/50707)]

Hello Agisilaos,

Assuming one has a base container image that builds libtiledb, building the Python wheel pretty much just works. The only problem I ran into was that my libtiledb gets installed to /lib, even though I pass `--prefix=/usr` to bootstrap. This means that when building the tiledb wheel, the compiler is looking for include files in `/lib/include/tiledb`, which is...odd. Unfortunately, when I build libtiledb the includes get installed to `/usr/include/tiledb` (where I would expect them to be installed by using /usr as the prefix). I can get around this in my image build by symlinking, but it's messy.

The bigger point however is that on x86_64, I can just `pip install tiledb` and it works without having to build my own libtiledb, so my image size can be much smaller as I don't need to have a compiler, etc. installed.

Let me know if there is any way I can help getting Linux aarch64 wheels going.

Thanks,

Brian

ihnorton:
[...]

PR #2114, opened 2024-11-22
Enable aarch64 wheel builds

dudoslav:
After the next TileDB release we should be able to produce aarch64 wheels

In order for this to work, we need to create/share linux-arm64-ubuntu24 runner

---

fixes https://github.com/TileDB-Inc/TileDB-Py/issues/2005


kounelisagis (approved):
LGTM, thanks! Don't forget to uncomment everything and we are good to merge after https://github.com/TileDB-Inc/TileDB-Py/pull/2116.

---

I checked a clone branch after cherry-picking the sole commit from https://github.com/TileDB-Inc/TileDB-Py/pull/2116, and everything looks great!

.github/workflows/build-wheels.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       matrix:
         buildplat:
           - [ubuntu-22.04, manylinux_x86_64]
+          - [linux-arm64-ubuntu24, manylinux_aarch64]
```
ihnorton, line 33: Do we need both of these?

PR #2116, opened 2024-11-29
Fix/Generalize test added in #2083

kounelisagis:
In https://github.com/TileDB-Inc/TileDB-Py/pull/2083, the `"S"` dtype is used in the new test to verify that dropping a fixed-size attribute and adding it back as variable-size now works as expected. However, this caused the issue described here: [https://github.com/TileDB-Inc/conda-forge-nightly-controller/issues/151#issuecomment-2507587087](https://github.com/TileDB-Inc/conda-forge-nightly-controller/issues/151#issuecomment-2507587087). 

Instead of testing with a specific value, we should test with both `"S"` and `"U"` dtypes but assert the fill value of the attribute before writing any data to it. This approach makes the test more generic for all types.

The same problem also appears in https://github.com/TileDB-Inc/TileDB-Py/pull/2114, so it would be better to apply this fix before merging that PR.

cc @jdblischak, @dudoslav 

---

[sc-59851]

### DIFF
--- a/tiledb/tests/test_schema_evolution.py
+++ b/tiledb/tests/test_schema_evolution.py
@@ -234,7 +234,10 @@ def test_schema_evolution_extend_check_bad_type():
     reason="Dropping a fixed-sized attribute and adding it back"
     "as a var-sized attribute is not supported in TileDB < 2.27",
 )
-def test_schema_evolution_drop_fixed_attribute_and_add_back_as_var_sized(tmp_path):
+@pytest.mark.parametrize("dtype_str", ["S", "U"])
+def test_schema_evolution_drop_fixed_attribute_and_add_back_as_var_sized(
+    tmp_path, dtype_str
+):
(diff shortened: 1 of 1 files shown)


jdblischak:
The only CI failure was for [macOS with Python 3.9](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/12295398546/job/34312204593?pr=2116#step:12:1022): `ERROR at teardown of TestDaskSupport.test_dask_multiattr_2d`

I suspect this was a spurious failure, and regardless appears unrelated to the changes this PR makes to `tiledb/tests/test_schema_evolution.py`. I'd restart the failed job but I don't have sufficient priveleges.

jdblischak (approved):
This test failure is blocking the feedstock release of 0.33.0 in https://github.com/conda-forge/tiledb-py-feedstock/pull/249

ihnorton (approved):

PR #2111, opened 2024-11-18
Remove workaround from `CurrentDomain`/`NDRectangle`

kounelisagis:
In https://github.com/TileDB-Inc/TileDB-Py/pull/2015, `CurrentDomain` and `NDRectangle` were wrapped using pybind11. In that PR, a workaround was introduced that involved storing a reference to the underlying `Domain` object to determine the dtype of the dimension of interest (commit of interest: https://github.com/TileDB-Inc/TileDB-Py/pull/2015/commits/63fb4baeb8d36603ce24602f7ba82d416d0e239d).

This PR takes advantage of https://github.com/TileDB-Inc/TileDB/pull/5229 to remove the workaround, thus making the code cleaner at both the Python and pybind11 levels, and reducing the complexity of the Python API.

PR #2117, opened 2024-12-02
Change minimum libtiledb version for `CurrentDomain` to fix daily tests

kounelisagis:
The changes in https://github.com/TileDB-Inc/TileDB-Py/pull/2111 caused the daily tests to fail. The error indicates that the latest code is not compatible with the libtiledb 2.25.X binaries, since I make use of `range_dtype` function which was introduced in 2.26.0.

> `error: no member named 'range_dtype' in 'tiledb::NDRectangle'`  
> [Workflow Run](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/12113046765/job/33767415747)

To address this, I suggest updating the required version for the current domain. Manually triggering the workflow for this branch appears to resolve the issue: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/12122419572. Our wheels built against 2.25.X should remain unaffected.

Other possible solutions include:
- Reverting the mentioned PR, which would keep the workaround in our code.  
- Branching out both codebases, though this could lead to a more chaotic structure.

To make something out of this, I would also suggest adding a backward compatibility job to our regular CI, derived from the daily tests, to help prevent similar issues in the future.

### DIFF
--- a/tiledb/__init__.py
+++ b/tiledb/__init__.py
@@ -21,7 +21,7 @@ else:
 
 from tiledb.libtiledb import version as libtiledb_version
 
-if libtiledb_version()[0] == 2 and libtiledb_version()[1] >= 25:
+if libtiledb_version()[0] == 2 and libtiledb_version()[1] >= 26:
     from .current_domain import CurrentDomain
     from .ndrectangle import NDRectangle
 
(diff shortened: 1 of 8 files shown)


ihnorton (approved):
LGTM.

Note that _IF_ we ever create a new 0.31.* release against 2.25, this PR cannot be included. However, making such a release would require a release branch from the original release PR, manual backports, etc. So please just make a note in the merge commit that `this PR cannot be included in a `0.31.*` release.`

PR #2126, opened 2025-01-02
Update references after renaming `dev` branch to `main`

kounelisagis:
This PR updates all references to the `dev` branch, changing them to `main`. After merging, the `dev` branch should be renamed to `main`.
This is a follow-up to https://github.com/TileDB-Inc/TileDB/pull/5407, which makes the same change for that repository.

cc. @dudoslav


ihnorton (approved):
LGTM

Issue #2127, opened 2025-01-03
some kind of weird error regarding float32 I never got before upgrading


BBC-Esq:
Below I paste a weird error I get when upgrading from 0.33 to 0.33.1.  Not sure if it matters, but I also have installed the following for my vector database program:

1 - tiledb-cloud==0.13.0
2 - tiledb-vector-search==0.11.0
3 - tblib==1.7.0
4 - numpy==1.26.4

<details><summary>LONG ERROR HERE EXPAND DOWNWARDS</summary>

```
Traceback (most recent call last):
  File "C:\Users\Airflow\AppData\Local\Programs\Python\Python311\Lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "C:\Users\Airflow\AppData\Local\Programs\Python\Python311\Lib\multiprocessing\process.py", line 108, in run
    self._target(self._args, **self._kwargs)
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\database_interactions.py", line 29, in create_vector_db_in_process
    create_vector_db.run()
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\Lib\site-packages\torch\utils\_contextlib.py", line 116, in decorate_context
    return func(args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\database_interactions.py", line 398, in run
    self.create_database(texts, embeddings)
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\Lib\site-packages\torch\utils\_contextlib.py", line 116, in decorate_context
    return func(*args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\database_interactions.py", line 246, in create_database
    db = TileDB.from_texts(
         ^^^^^^^^^^^^^^^^^^
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\Lib\site-packages\langchain_community\vectorstores\tiledb.py", line 726, in from_texts
    return cls.from(
           ^^^^^^^^^^^
  File "D:\Scripts\VectorDB-Plugin-for-LM-Studio\working\Lib\site-packages\langchain_community\vectorstores\tiledb.py", line 578, in from
    tiledb_vs.ingestion.ingest(
[...]

PR #2128, opened 2025-01-03
Explicitly set the numpy type passed in `QueryExperimental::set_data_buffer`

kounelisagis:
`QueryExperimental::set_data_buffer` requires an unsigned integer as its last argument. It seems that with numpy versions earlier than 2, this argument becomes a numpy float64 (with `buffer_sizes[i]` being a numpy uint64 and `ncells` being a Python int).

Let's fix it by explicitly cast to numpy uint64.

Closes https://github.com/TileDB-Inc/TileDB-Py/issues/2127, closes https://github.com/TileDB-Inc/TileDB-Py/issues/2124.

### DIFF
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -899,7 +899,7 @@ class Array:
             q.set_data_buffer(
                 buffer_name,
                 output_values[i],
-                buffer_sizes[i] * ncells,
+                np.uint64(buffer_sizes[i] * ncells),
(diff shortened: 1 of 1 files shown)


ihnorton (approved):
LGTM, thanks (confirmed locally). We need to add a test variation against numpy 1 as discussed, but that can be done separately.

PR #2139, opened 2025-01-10
Fix `view_attr` not being respected by `__getitem__` and `subarray`

kounelisagis:
This PR fixes an issue where the `attr` argument in the `Array.open` method (internally stored as `view_attr`) was not respected during query operations. Consequently, the `Array.query` method ignored the specified attribute(s) and returned all attributes instead. The issue has been resolved by updating the `__getitem__` and `subarray` methods of both Sparse and Dense Arrays.

---

[sc-61659]

### DIFF
--- a/tiledb/dense_array.py
+++ b/tiledb/dense_array.py
@@ -77,8 +77,7 @@ class DenseArrayImpl(Array):
 
         """
         if self.view_attr:
-            result = self.subarray(selection, attrs=(self.view_attr,))
-            return result[self.view_attr]
+            return self.subarray(selection)
 
         result = self.subarray(selection)
         for i in range(self.schema.nattr):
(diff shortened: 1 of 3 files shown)


ihnorton (approved):

tiledb/sparse_array.py
```diff
@@ -292,6 +293,9 @@ def __getitem__(self, selection):
         >>> # A[5.0:579.9]
 
         """
+        if self.view_attr:
```
nguyenv, line 296:
```suggestion
        if self.view_attr is not None:
```

tiledb/dense_array.py
```diff
@@ -269,6 +268,8 @@ def subarray(self, selection, attrs=None, cond=None, coords=False, order=None):
             attr = self.schema.attr(0)
             if attr.isanon:
                 return out[attr._internal_name]
+        if self.view_attr:
```
nguyenv, line 271:
```suggestion
        if self.view_attr is not None:
```

nguyenv (approved):
lgtm!

PR #2142, opened 2025-01-14
Add `commits` to vacuum modes

kounelisagis:
[sc-55294]


ihnorton (approved):
LGTM after the nit fix.

PR #2144, opened 2025-01-15
Bugfix: Arrays losing CloudArray interface

kounelisagis:
This PR fixes the case when `tiledb-cloud` package is present but instances of Dense and Sparse Array are returned without the correct inheritance from: https://github.com/TileDB-Inc/TileDB-Py/blob/e1f57bd940731460d8cfbcdb74a1f763aef521c9/tiledb/array.py#L425-L431.

---

[sc-61938]

### DIFF
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -416,17 +416,14 @@ class Array:
         if ctx is None:
             ctx = default_ctx()
 
-        from .dense_array import DenseArrayImpl
-        from .sparse_array import SparseArrayImpl
-
         tmp_array = preload_array(uri, mode, key, timestamp, ctx)
 
         if tmp_array._schema()._array_type == lt.ArrayType.SPARSE:
-            return SparseArrayImpl(
+            return tiledb.SparseArray(
                 uri, mode, key, timestamp, attr, ctx, preloaded_array=tmp_array
             )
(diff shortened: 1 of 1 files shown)


tiledb/array.py
```diff
@@ -300,7 +300,14 @@ def preload_array(uri, mode, key, timestamp, ctx=None):
...
 
-class Array:
+# If tiledb.cloud is installed, add CloudArray methods to TileDB Arrays
+try:
+    from tiledb.cloud.cloudarray import CloudArray
+except ImportError:
+    CloudArray = object
+
+
+class Array(CloudArray):
```
sgillies, line 310: @kounelisagis Great!

sgillies (changes requested):
@kounelisagis Thank you! I'd like to see the test at https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/tests/test_threaded_import.py moved to the TileDB-Py project to guard against possible future regressions.

ihnorton (approved):
g2g as long as @sgillies agrees. thanks!

sgillies (approved):
Thanks @kounelisagis ! I ran tiledb-cloud's tests with this branch and the problem is resolved.

PR #2150, opened 2025-02-04
Bug fix: Data buffer is not resized for byte objects after completing the buffer and offsets conversion

kounelisagis:
This PR fixes an issue with the incorrect calculation of the `data_buffer` size when empty bytes are present in the Python list being written to a TileDB Array. The size of the `data_buffer` was incorrectly set to the size of the Python list, which is incorrect because empty bytes are not included in the `data_buffer` being sent to the `QueryExperimental::set_data_buffer` C++ API.

---

[sc-62594]

### DIFF
--- a/tiledb/npbuffer.cc
+++ b/tiledb/npbuffer.cc
@@ -211,6 +211,8 @@ if (PyUnicode_Check(u.ptr())) {
       output_p += sz;
       idx++;
     }
+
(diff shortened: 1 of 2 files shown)


kounelisagis:
The macos CI error seems unrelated. I didn't experience it locally.

ihnorton (approved):

PR #2151, opened 2025-02-04
Add pytest flaky-retry mechanism for handling intermittent test failures

kounelisagis:
This PR introduces the `@pytest.mark.flaky` decorator to retry `test_dask_multiattr_2d` up to 3 times with a 2-second delay between attempts. This helps mitigate intermittent test failures caused by factors such as network instability, timeouts, or race conditions.

Fixes https://github.com/TileDB-Inc/TileDB-Py/issues/2149

### DIFF
--- a/tiledb/tests/test_dask.py
+++ b/tiledb/tests/test_dask.py
@@ -42,6 +42,7 @@ class TestDaskSupport(DiskTestCase):
 
         tiledb.DenseArray.create(uri, schema)
 
+    @pytest.mark.flaky(reruns=3, reruns_delay=2)
     @pytest.mark.filterwarnings("ignore:There is no current event loop")
     def test_dask_multiattr_2d(self):
         uri = self.path("multiattr")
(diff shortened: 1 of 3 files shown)


tiledb/tests/test_dask.py
```diff
@@ -42,6 +42,7 @@ def _make_multiattr_2d(self, uri, shape=(0, 100), tile=10):
 
         tiledb.DenseArray.create(uri, schema)
 
+    @pytest.mark.flaky(reruns=3, reruns_delay=2)
```
ihnorton, line 45: Should we use `exclude=[TileDBError]` here?

ihnorton (approved):
One comment, otherwise LGTM. Thanks

tiledb/tests/test_dask.py
kounelisagis: I added [only-rerun](https://pytest-rerunfailures.readthedocs.io/latest/mark.html#only-rerun) argument.

PR #2015, opened 2024-07-23
Wrap current domain API

kounelisagis:
Wrapping https://github.com/TileDB-Inc/TileDB/pull/5056

PR #2156, opened 2025-02-10
Consider the Current Domain on indexing

kounelisagis:
After introducing Current Domain in https://github.com/TileDB-Inc/TileDB-Py/pull/2015, there is still work to be done to fully utilize it. This PR addresses cases where the domain was used without considering the Current Domain.

The original issue arose from the following TileDB-SOMA example, where using the full domain of the array eventually led to an OOM crash:

```
import tiledb, tiledbsoma as soma
import numpy as np
import pyarrow as pa

# write simple dense array using SOMA
A = soma.DenseNDArray.create("foo",type=pa.int8(),shape=(10,5))
A.write((slice(None),slice(None)), pa.Tensor.from_numpy(np.arange(0, 50).reshape(10,5)))
A.close()

# read back using TileDB-Py
A = tiledb.open("foo", mode="r")
A.query().df[:]
```

---

[sc-61914]

### DIFF
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -127,6 +127,22 @@ def index_domain_subarray(array, dom, idx: tuple):
         )
 
     subarray = list()
+
+    # In case that current domain is not empty, we need to consider it
+    if (
+        hasattr(array.schema, "current_domain")
+        and not array.schema.current_domain.is_empty
+    ):
+        for i in range(array.schema.domain.ndim):
+            subarray.append(
+                (
+                    array.schema.current_domain.ndrectangle.range(i)[0],
+                    array.schema.current_domain.ndrectangle.range(i)[1],
+                )
+            )
+
+        return subarray
+
     for r in range(ndim):
         # extract lower and upper bounds for domain dimension extent
         dim = dom.dim(r)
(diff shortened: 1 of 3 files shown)


tiledb/array.py
```diff
@@ -127,6 +127,22 @@ def index_domain_subarray(array, dom, idx: tuple):
         )
 
     subarray = list()
+
+    # In case that current domain is not empty, we need to consider it
```
johnkerl, line 131:
```suggestion
    # In the case that current domain is non-empty, we need to consider it
```

tiledb/multirange_indexing.py
```diff
@@ -195,6 +195,22 @@ def label_ranges_from_selection(selection):
 
 def getitem_ranges(array: Array, idx: Any) -> Sequence[Sequence[Range]]:
     ranges: List[Sequence[Range]] = [()] * array.schema.domain.ndim
+
+    # In case that current domain is not empty, we need to consider it
```
johnkerl, line 199:
```suggestion
    # In the case that current domain is non-empty, we need to consider it
```

johnkerl (commented):
🚀 

LGTM, thank you!! :)

johnkerl (approved):

PR #2164, opened 2025-02-20
Fix the error caused by non-decodable bytes when retrieving the non-empty domain for `TILEDB_ASCII` dimensions

kounelisagis:
As discovered in [TileDB-ML](https://github.com/TileDB-Inc/TileDB-ML/actions/runs/13406770379/job/37448019451#step:8:35492), calling the `Array::_non_empty_domain` pybind11 function on a `TILEDB_ASCII` dimension containing non-decodable bytes (in terms of UTF-8) resulted in errors such as:  
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0x81 in position 0: invalid start byte`.  
The `pybind11::make_tuple` function assumed that the input would always be a printable string, requiring special handling for this case.  
This PR addresses the issue and also ensures that an exception is thrown for dimensions of unicode string type, as this is not supported until TileDB 2.27.0.

### DIFF
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -673,10 +673,12 @@ class Array:
 
             res_x, res_y = self.array._non_empty_domain(dim_idx, dim_dtype)
 
+            # convert to bytes if needed
             if dim_dtype == np.bytes_:
-                # convert to bytes if needed
-                res_x = bytes(res_x, encoding="utf8")
-                res_y = bytes(res_y, encoding="utf8")
+                if not isinstance(res_x, bytes):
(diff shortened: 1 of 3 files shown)


ihnorton (approved):
LGTM after double-checking the return type matches 0.32

ktsitsi (approved):

kounelisagis:
> LGTM after double-checking the return type matches 0.32

Both the structure of the returned value from `Array::_non_empty_domain` (a tuple of tuples - one for each dimension) and the type of each internal tuple (bytes, in this case) remain the same.

PR #2167, opened 2025-02-27
Add tests with heterogeneous arrays that contain variable-length nullable numeric attributes

kounelisagis:
This PR adds tests with heterogeneous arrays that contain variable-length nullable numeric attributes. Additionally, it provides an appropriate error message when `Arrow` is used in such cases. The new message replaces the misleading `Expected 2 buffers for imported type uint32, ArrowArray struct has 3` error that was previously displayed.

In a future PR, support for Arrow could be added once this is supported by Arrow C Data Interface.

---

[sc-41582]

### DIFF
--- a/tiledb/multirange_indexing.py
+++ b/tiledb/multirange_indexing.py
@@ -239,6 +239,15 @@ def getitem_ranges_with_labels(
     return dim_ranges, label_ranges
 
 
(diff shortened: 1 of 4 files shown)


kounelisagis:
What do you think about the semantics?

PR #2168, opened 2025-03-04
Improve the logic for using `Arrow` when it is not explicitly requested

kounelisagis:
This PR expands the cases where a Pandas DataFrame can be retrieved from a TileDB `Array` or `Query`. This improvement applies _only_ when the `use_arrow` argument is not set.  

### Logic before this PR:  
If the `pyarrow` package was available, `use_arrow` would automatically be set to `True`. This caused issues in cases involving multi-value attributes. The problem was that the user never explicitly requested Arrow, yet encountered an Arrow-related issue. Additionally, the expected `TileDBError` for multi-value attributes was never triggered because the loop `for attr in map(array.attr, query.attrs or ())` did not work as intended. The reason is that `array.attr` is a method that takes an attribute index or name as a parameter, rather than returning attributes directly.  

### Logic after this PR:  
A new method, `attr_names`, has been introduced in the `ArraySchema` and `Array` classes. Utilizing this method, we can now determine whether Arrow can be used for specific attributes. If Arrow is supported, we use it. Otherwise, we don't.  

---

[sc-41582]

### DIFF
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -569,6 +569,11 @@ class Array:
         :raises TypeError: invalid key type"""
         return self.schema.attr(key)
 
+    @property
+    def attr_names(self):
+        """Returns a list of attribute names"""
+        return self.schema.attr_names
+
     def dim(self, dim_id):
         """Returns a :py:class:`Dim` instance given a dim index or name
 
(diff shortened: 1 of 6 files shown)


gspowley (approved):
Thanks @kounelisagis, nice improvement!

PR #2170, opened 2025-03-05
Expand `.df` Array/Query accessor to allow indexing with NumPy and PyArrow arrays

kounelisagis:
This PR expands the functionality of the `.df` TileDB Array/Query accessor to also allow indexing with both PyArrow and NumPy arrays, instead of only scalar, slice, or a list of scalars or slice objects, as it currently does.

cc. @johnkerl for visibility.

---

[sc-63089]

### DIFF
--- a/tiledb/multirange_indexing.py
+++ b/tiledb/multirange_indexing.py
@@ -155,10 +155,8 @@ def iter_ranges(
         assert len(sel) == 2
         yield to_scalar(sel[0]), to_scalar(sel[1])
 
-    elif (
-        isinstance(sel, list)
-        or isinstance(sel, np.ndarray)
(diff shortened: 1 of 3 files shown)


tiledb/multirange_indexing.py
```diff
@@ -145,7 +155,11 @@ def iter_ranges(
         assert len(sel) == 2
         yield to_scalar(sel[0]), to_scalar(sel[1])
 
-    elif isinstance(sel, list):
+    elif (
+        isinstance(sel, list)
+        or isinstance(sel, np.ndarray)
```
nguyenv, line 160:
```suggestion
        isinstance(sel, (list, np.ndarray))
```

nguyenv (approved):

ihnorton (approved):

PR #2171, opened 2025-03-05
Fix for query conditions that contain attributes with dot(s) when using the `in` operator

kounelisagis:
Similarly to https://github.com/single-cell-data/TileDB-SOMA/pull/3746, let's fix the issue where a query condition contains attributes with dot(s) when using the `in` operator.

---

[sc-63843]


johnkerl (approved):
🚢

ihnorton (approved):

PR #2174, opened 2025-03-07
Skip `test_no_warning_multiprocessing_without_ctx` on macOS

kounelisagis:
The test `test_no_warning_multiprocessing_without_ctx` has been failing locally on macOS, in the normal macOS CI, and consistently in the macOS "Build and Test Wheels" workflow. This failure is preventing us from building the macOS wheels. This PR skips that test on macOS, extending the existing skip for Windows.

### DIFF
--- a/tiledb/tests/test_fork_ctx.py
+++ b/tiledb/tests/test_fork_ctx.py
@@ -78,7 +78,8 @@ def test_no_warning_multiprocessing_without_ctx():
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32", reason="fork() is not available on Windows"
+    sys.platform != "linux",
+    reason=f"fork() is not available on Windows, and this test randomly fails on macOS.",
 )
 @pytest.mark.filterwarnings("ignore::DeprecationWarning")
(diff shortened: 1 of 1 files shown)


tiledb/tests/test_fork_ctx.py
```diff
@@ -64,7 +64,8 @@ def test_warning_fork_with_default_ctx():
 
 
 @pytest.mark.skipif(
-    sys.platform == "win32", reason="fork() is not available on Windows"
+    sys.platform != "linux",
+    reason=f"fork() is not available on Windows, and this test randomly fails on macOS.",
```
ypatia, line 68: is this something we want to open a ticket to investigate later?

kounelisagis: [sc-64149]

dudoslav:
1. Do we know the cause of this error?
2. Can you post link to the error message/ or just the error message?

I would prefer fixing the issue instead of skipping the test (of course it possible/if it makes sense).

kounelisagis:
Closing because of #2181.

PR #2176, opened 2025-03-13
Fix URI passed to consolidation

spencerseale:
`self.array.uri` would need to be changed to private method `self.array._uri()`. But the attribute is stored to the Py class so use it as is done in logic down on ln 766.

This impacts APIs including tiledb-py, cloud-py and any other API consolidating with specific fragments URIs


kounelisagis (approved):
Nice find! Thanks!

PR #2178, opened 2025-03-19
Fix: Consider index passed, even if the current domain is present

kounelisagis:
This PR fixes an issue originating from https://github.com/TileDB-Inc/TileDB-Py/pull/2156, which resulted in the current domain being used whenever it was present. This caused problems when indexing with an explicit `idx` while a current domain was present, leading to `idx` and nonempty domain not being used.

cc. @ypatia

---

[sc-64437]

### DIFF
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -150,18 +150,13 @@ def index_domain_subarray(array, dom, idx: tuple):
 
         # In the case that current domain is non-empty, we need to consider it
         if (
-            start is None
-            and stop is None
-            and hasattr(array.schema, "current_domain")
+            hasattr(array.schema, "current_domain")
             and not array.schema.current_domain.is_empty
         ):
-            subarray.append(
-                (
-                    array.schema.current_domain.ndrectangle.range(r)[0],
-                    array.schema.current_domain.ndrectangle.range(r)[1],
-                )
-            )
-            continue
+            if start is None:
+                dim_lb = array.schema.current_domain.ndrectangle.range(r)[0]
+            if stop is None:
+                dim_ub = array.schema.current_domain.ndrectangle.range(r)[1]
 
         if np.issubdtype(dim_dtype, np.str_) or np.issubdtype(dim_dtype, np.bytes_):
             if start is None or stop is None:
--- a/tiledb/array.py
+++ b/tiledb/array.py
@@ -128,21 +128,6 @@ def index_domain_subarray(array, dom, idx: tuple):
 
     subarray = list()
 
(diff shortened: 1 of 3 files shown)


johnkerl (changes requested):
I would like to see a user-level repro script demonstrating the problem at the UX level, but, I suppose your unit-test case suffices for me to be able to reconstruct one.

I would also like to see some provenance of how/where this error was reported. Maybe just a Shortcut hyperlink in the Description section.

kounelisagis:
> I would like to see a user-level repro script demonstrating the problem at the UX level, but, I suppose your unit-test case suffices for me to be able to reconstruct one.
> 
> I would also like to see some provenance of how/where this error was reported. Maybe just a Shortcut hyperlink in the Description section.

A user-level repro script is available on the Shortcut mentioned.

tiledb/array.py
```diff
@@ -163,6 +148,21 @@ def index_domain_subarray(array, dom, idx: tuple):
 
         start, stop, step = dim_slice.start, dim_slice.stop, dim_slice.step
 
+        # In the case that current domain is non-empty, we need to consider it
+        if (
+            start is None
+            and stop is None
+            and hasattr(array.schema, "current_domain")
```
teo-tsirpanis, line 155: When is this line false?

bkmartinjr (approved):

tiledb/array.py
```diff
@@ -163,6 +148,21 @@ def index_domain_subarray(array, dom, idx: tuple):
...
+            and not array.schema.current_domain.is_empty
+        ):
+            subarray.append(
+                (
+                    array.schema.current_domain.ndrectangle.range(r)[0],
+                    array.schema.current_domain.ndrectangle.range(r)[1],
+                )
+            )
+            continue
+
```
ihnorton, line 165: How does this interact with the `if` statement below? Is there any possibility of adding a subarray twice?

tiledb/array.py
kounelisagis: This is true when no explicit index is requested (i.e., `A[:]`) and the array has a non-empty current domain. In all other cases, this is false.

tiledb/array.py
kounelisagis: I don't see a way for a subarray to be added twice for the same dimension. The `continue` statement will prevent this from happening if both `start` and `stop` are `None` and there is a non-empty current domain.

ihnorton: Nevermind, I overlooked the continue - thanks!

ihnorton (approved):

PR #2181, opened 2025-04-02
Run tests of `test_fork_ctx` as subprocesses

kounelisagis:
This PR modifies the tests in `test_fork_ctx.py` to run as subprocesses, fixing #2175. This has been verified by running the "Build and test wheels" workflow against the current branch: [run](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/14227535058/job/39870535651). The failure rate was previously 100% for macos.

---

[sc-64149][sc-57975]

### DIFF
--- a/tiledb/tests/test_fork_ctx.py
+++ b/tiledb/tests/test_fork_ctx.py
@@ -9,14 +9,35 @@ Python 3.14 will make it so you never accidentally fork():
 multiprocessing will default to "spawn" on Linux.
 """
 
-import multiprocessing
-import os
+import subprocess
 import sys
-import warnings
 
 import pytest
 
+
+def run_in_subprocess(code):
+    """Runs code in a separate subprocess."""
+    script = f"""
+import os
+import warnings
 import tiledb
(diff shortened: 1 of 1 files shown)


sgillies (approved):
@kounelisagis I don't understand why crashes were happening in the builds and not locally or in our CI tests, but this seems like a pragmatic solution.

tiledb/tests/test_fork_ctx.py
```diff
@@ -9,96 +9,129 @@
...
 )
-@pytest.mark.filterwarnings("ignore::DeprecationWarning")
 def test_warning_multiprocessing_with_ctx():
     """Get a warning if we fork after creating a tiledb context."""
+    run_in_subprocess(
+        """
     _ = tiledb.Ctx()
     mp = multiprocessing.get_context("fork")
     p = mp.Process()
-    with pytest.warns(UserWarning, match="TileDB is a multithreading library"):
```
ihnorton, line 88: We want to _assert that this warning is raised..._

PR #2184, opened 2025-04-15
Remove failing test now that HDFS support has been removed

jdblischak:
Support for HDFS was removed upstream in TileDB in https://github.com/TileDB-Inc/TileDB/pull/5496. Now the nightly TileDB-Py builds against nightly TileDB fail this test with a `KeyError`.

xref: https://github.com/TileDB-Inc/conda-forge-nightly-controller/issues/182#issuecomment-2806755307, https://github.com/TileDB-Inc/centralized-tiledb-nightlies/issues/46#issuecomment-2806899886

PR #2185, opened 2025-04-15
Remove support for HDFS

jdblischak:
Follow-up to https://github.com/TileDB-Inc/TileDB-Py/pull/2184. HDFS support was removed upstream in TileDB via https://github.com/TileDB-Inc/TileDB/pull/5496. However, this change hasn't been released yet, so I leave it to @kounelisagis to decide when it makes sense to merge this PR.


kounelisagis:
All of the changes should be made conditionally based on the libtiledb version. The same should also have been done for https://github.com/TileDB-Inc/TileDB-Py/pull/2184.

kounelisagis (changes requested):
All of the changes should be made conditionally based on the libtiledb version. The same should also have been done for https://github.com/TileDB-Inc/TileDB-Py/pull/2184.

jdblischak:
> All of the changes should be made conditionally based on the libtiledb version.

Doesn't TileDB-Py only support a single minor release of TileDB at a time? I don't see the value in adding conditional logic to support TileDB<2.28 if that is not supported.

> The same should also have been done for #2184.

I'm not overly worried about #2184. That test was super basic. Do you want me to send a follow-up PR so restore that test for TileDB<=2.27?

tiledb/libtiledb/enum.cc
```diff
@@ -163,9 +163,8 @@ void init_enums(py::module& m) {
         .value("S3", TILEDB_S3)
         .value("AZURE", TILEDB_AZURE)
         .value("GCS", TILEDB_GCS)
-        .value("HDFS", TILEDB_HDFS);
```
teo-tsirpanis, line 166: Removing this is a breaking change. Are we sure we want that? I didn't remove `TILEDB_HDFS` in my Core PR for this reason.

jdblischak: I defer to you for any major decisions. I made this simple PR by naively removing all mentions of HDFS. I just added this line back. Please let me know if there are any other changes you would like me to make.

ihnorton: @kounelisagis is going to add a warning here.

kounelisagis:
The test is failing because the token is missing after forking.

kounelisagis (approved):

teo-tsirpanis (approved):

tiledb/libtiledb/enum.cc
jdblischak: Did we decide against adding a warning? I don't feel strongly one way or the other; but the file `enum.cc` is no longer edited by this PR.

tiledb/tests/test_vfs.py
```diff
@@ -23,8 +23,12 @@ def test_supports(self):
         self.assertIsInstance(vfs.supports("gcs"), bool)
         self.assertIsInstance(vfs.supports("azure"), bool)
 
-        with self.assertRaises(ValueError):
-            vfs.supports("invalid")
+        # Supports is not raising an error for invalid VFS as it used to
+        if tiledb.version() <= (0, 33, 6):
+            with self.assertRaises(ValueError):
+                vfs.supports("invalid")
```
jdblischak, line 29: It looks like this test is [failing](https://github.com/TileDB-Inc/TileDB-Py/actions/runs/14618645566/job/41012737649?pr=2185#step:12:1553). Is this a concern?

```
FAILED ../home/runner/work/TileDB-Py/TileDB-Py/tiledb/tests/test_vfs.py::TestVFS::test_supports - Failed: DID NOT RAISE <class 'ValueError'>
```

Because this workflow performs a shallow clone, the automated version string is `0.1.dev1+gdd66f83`, which is `<=0.33.6`

teo-tsirpanis: We should unconditionally `assertFalse`. I don't see the point of this condition; the tests always use the version of tiledb-py they are next to.

tiledb/libtiledb/enum.cc
kounelisagis: I couldn't find a way to do so.

teo-tsirpanis: We had discussed this and my suggestion was to not do anything since this is an internal API.

jdblischak (commented):
I approve!

PR #2205, opened 2025-06-06
Fix .df[:] indexing for enumerations on dense arrays

nguyenv:
TileDB sets unmatched elements in query conditions with an attribute's fill value (by default, for integers, this is the min values; for example -9223372036854775808 for int32). For enumerated types, when converting from dictionary indexes to dictionary values, Pyarrow cannot map the index -9223372036854775808 onto a corresponding value because it does not exist. To workaround this, replace fill values with -1, which is understood to mean null in Pandas, before converting to Pyarrow.


johnkerl (commented):
With this repro script

```
#!/usr/bin/env python

import tiledb
import numpy as np
import pandas as pd
import pyarrow as pa

import sys, os, shutil
uri = 'data-enum-bug-py'
if len(sys.argv) == 2:
    uri = sys.argv[1]
if os.path.exists(uri):
    shutil.rmtree(uri)

pandas_df = pd.DataFrame(
    {
        "id": pd.Series([10, 20, 30, 40, 50, 60, 70, 80], dtype=np.int64),
        "color": pd.Categorical(
            ["red", "yellow", "green", "green", "yellow", "red", "yellow", "green"],
            ordered=True,
            categories=["red", "yellow", "green"],
        ),
    },
)

tiledb.from_pandas(uri, pandas_df, sparse=False)

with tiledb.open(uri) as A:
    print("READ FULL DATAFRAME")
    print(A.df[:])

    print()
    print("QUERY FOR ID== 40, DICT OUTPUT")
    slice_query = A.query(cond='id== 40', use_arrow=False)
    print(slice_query[:])

    print()
    print("QUERY FOR ID== 40, PANDAS OUTPUT")
    slice_query = A.query(cond='id== 40', use_arrow=False)
    print(slice_query.df[:])

    print()
    print("QUERY FOR COLOR == YELLOW, DICT OUTPUT")
    slice_query = A.query(cond='color == "yellow"', use_arrow=False)
    print(slice_query[:])

    print()
    print("QUERY FOR COLOR == YELLOW, PANDAS OUTPUT")
    slice_query = A.query(cond='color == "yellow"', use_arrow=False)
    print(slice_query.df[:])
```

and with this PR's branch checked out, I am getting the output

```
READ FULL DATAFRAME
   id   color
0  10     red
1  20  yellow
2  30   green
3  40   green
4  50  yellow
5  60     red
6  70  yellow
7  80   green

QUERY FOR ID== 40, DICT OUTPUT
  File "/Users/kerl/Desktop/endoc/bug.py", line 35, in <module>
    print(slice_query[:])
          ~~~~~~~~~~~^^^
  File "/Users/kerl/git/TileDB-Inc/TileDB-Py/tiledb/query.py", line 132, in __getitem__
    return self._array.subarray(
           ~~~~~~~~~~~~~~~~~~~~^
        selection,
        ^^^^^^^^^^
    ...<3 lines>...
        order=self._order,
        ^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/kerl/git/TileDB-Inc/TileDB-Py/tiledb/dense_array.py", line 282, in subarray
    out = self._read_dense_subarray(subarray, attr_names, cond, layout, coords)
  File "/Users/kerl/git/TileDB-Inc/TileDB-Py/tiledb/dense_array.py", line 316, in _read_dense_subarray
    q.submit()
    ~~~~~~~~^^
tiledb.libtiledb.TileDBError: ValueError: negative dimensions are not allowed (/Users/kerl/git/TileDB-Inc/TileDB-Py/tiledb/core.cc:105)
```

PR #2223, opened 2025-07-24
Restructure `Profile` REST test and example

kounelisagis:
This PR addresses an issue that emerged during CI runs of other repositories that run TileDB-Py tests. Specifically, in https://github.com/TileDB-Inc/centralized-tiledb-nightlies/issues/53#issuecomment-3108263515, it's clear that the `profile.py` example is not being properly skipped. When executed in an environment where `TILEDB_TOKEN` is not set, it runs as part of `test_examples.py` and fails.

This issue makes sense because, in https://github.com/TileDB-Inc/TileDB-Py/pull/2189, the example was implemented differently from all other existing examples. In particular, it was made to perform REST operations - something that no other example does.

To standardize the behavior of tests and examples (since they are executed via `test_examples.py` and must pass), this PR moves the REST-related test code to `test_cloud.py`, which already contains a test with similar logic and is the only one equipped to handle such behaviors.

The example is updated to demonstrate the intended usage flow for users who want to use the `Profile` feature. However, it no longer executed via `test_examples.py`.

### DIFF
--- a/examples/profile.py
+++ b/examples/profile.py
@@ -26,6 +26,9 @@
 #
 
 # This example demonstrates how to create, save, use, and remove a Profile in TileDB.
+# NOTE: This example is not intended to be run as a test (test_examples.py excludes it)
+# since it requires a TileDB REST token.
+# Profiles are getting checked in test_cloud.py.
 
 import tiledb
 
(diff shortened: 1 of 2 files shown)


examples/profile.py
```diff
@@ -25,48 +25,36 @@
...
-        datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
-        + "-"
-        + "".join(random.choice(string.ascii_letters) for _ in range(5))
-    )
-    uri = f"tiledb://{tiledb_namespace}/s3://{s3_bucket}/{array_name}"
+    # By using the context to create a new array, the REST credentials from the profile will be used.
+
+    uri = f"local_array"
+    # Change that to something like the following if you want to use the REST credentials:
+    # uri = f"tiledb://<workspace>/<teamspace>/<array-name>"
```
ypatia, line 51: To avoid this kind of potentially confusing clarifications, is there a way to keep the example as close to reality as possible and just skip that example from running - similar to what we did in Core?

ypatia (approved):
Thanks!

PR #2235, opened 2025-08-19
Fix numpy comparisons; use `==` for dtypes and `isinstance` for arrays

### LINKED ISSUE #2234: Python class equivalence usage questions
I've noticed in several places in `dense_array.py` and `sparse_array.py` that `tiledb-py` is using a non-accepted equivalence when looking at an object's underlying numpy type. 

Example: https://github.com/TileDB-Inc/TileDB-Py/blob/main/tiledb/dense_array.py#L444-L447

Using `is` for class equivalence looks at the memory address of the instantiation of the class. In my case, in which I'm using `dask` and other distributed computing tools, this can create false errors. 

For example, if I create a local cluster of `N` dask workers that are `processes`, I'm creating `N` instantiations of `numpy.ndarray`. If I then pass a `Pandas Dataframe` (with an underlying `numpy` array as its datatype), then I am comparing the memory address of `numpy` from the scheduler with the memory address of `numpy` in the worker. This will catch non-existent problems and make my processes fail. 

In the [Numpy docs](https://numpy.org/doc/stable/reference/arrays.dtypes.html#checking-the-data-type), they mention that equivalence should be implemented using `==` instead of `is`. 

From that link:
> When checking for a specific data type, use == comparison.

> Do not use is because data type objects may or may not be singletons.

If this is intentional, I would ask that there be comments made in the code explaining what outcome you're checking for that requires a non-standard approach.

Thank you!

kounelisagis:
This PR fixes NumPy object comparison issues in `dense_array.py` and `sparse_array.py` by implementing two changes:

1. **dtype comparisons**: Updated to use `==` operator instead of `is`, following the [NumPy recommendation](https://numpy.org/doc/stable/reference/arrays.dtypes.html#checking-the-data-type):
   > When checking for a specific data type, use `==` comparison.
   > Do not use `is`, because data type objects may or may not be singletons.
2. **Array type checks**: Updated to use `isinstance(x, np.ndarray)` instead of `type(x) is np.ndarray`.

These changes resolve errors in Dask and other distributed computing environments where worker processes compare objects across process boundaries, while also improving compatibility with the broader NumPy ecosystem.

Closes #2234

cc. @kylemann16

### DIFF
--- a/tiledb/dense_array.py
+++ b/tiledb/dense_array.py
@@ -441,7 +441,7 @@ class DenseArrayImpl(Array):
             labels = {
                 name: (
                     data
-                    if not type(data) is np.ndarray or data.dtype == np.dtype("O")
+                    if not isinstance(data, np.ndarray) or data.dtype == np.dtype("O")
                     else np.ascontiguousarray(
                         data, dtype=self.schema.dim_label(name).dtype
                     )
@@ -458,7 +458,7 @@ class DenseArrayImpl(Array):
 
                 attributes.append(attr._internal_name)
                 # object arrays are var-len and handled later
-                if type(attr_val) is np.ndarray and attr_val.dtype != np.dtype("O"):
+                if isinstance(attr_val, np.ndarray) and attr_val.dtype != np.dtype("O"):
                     if attr.isnullable and name not in nullmaps:
                         try:
                             nullmaps[name] = ~np.ma.masked_invalid(attr_val).mask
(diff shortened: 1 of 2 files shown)


tiledb/dense_array.py
```diff
@@ -441,7 +441,7 @@ def _setitem_impl(self, selection, val, nullmaps: dict):
             labels = {
                 name: (
                     data
-                    if not type(data) is np.ndarray or data.dtype is np.dtype("O")
+                    if not type(data) is np.ndarray or data.dtype == np.dtype("O")
```
kylemann16, line 444: I forgot to bring this up in the issue, but is there a reason that `tiledb-py` is doing `if type(attr_val) is np.ndarray` instead of `isinstance(attr_val, np.ndarray)`? I think this could present a similar pitfall.

kylemann16:
Thank you for taking the time to look at this!

ihnorton (approved):
LGTM (regression test would be good though, if possible)

kylemann16 (approved):
Looks good to me, thank you!

kounelisagis:
> LGTM (regression test would be good though, if possible)

I thought about it longer than I should have, but I couldn't create one.

PR #2248, opened 2025-09-24
Initialize `uint32_t` to prevent OS-dependent undefined behavior

kounelisagis:
This PR fixes an issue introduced with the latest pybind11 release (version 3.0). Specifically, `test_cc.py::test_array` was failing in some macOS-13 jobs with the following error:

```
        arr._open(lt.QueryType.WRITE)
        arr._set_open_timestamp_start(2)
        arr._set_open_timestamp_end(2)
        arr._delete_metadata("key")
        arr._close()
    
        arr._set_open_timestamp_start(3)
        arr._set_open_timestamp_end(3)
        arr._open(lt.QueryType.READ)
>       with pytest.raises(KeyError):
             ^^^^^^^^^^^^^^^^^^^^^^^
E       Failed: DID NOT RAISE <class 'KeyError'>
```

It turns out that initializing the `uint32_t value_num`, which is passed to the `Array::get_metadata` C++ API, fixes the problem. The CI results for this branch should not be considered, since pybind11 is still pinned to <3. I would like to handle this in a separate PR.

CI with pybind11 3.0 after applying the same fix: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/17979314898

---

cc. @jdblischak - I believe you are currently skipping this test.

Ref CORE-304

### DIFF
--- a/tiledb/libtiledb/metadata.h
+++ b/tiledb/libtiledb/metadata.h
@@ -164,13 +164,13 @@ class MetadataAdapter {
     py::tuple get_metadata(
         T& array_or_group, const std::string& key, bool is_ndarray) {
         tiledb_datatype_t tdb_type;
-        uint32_t value_num;
+        uint32_t value_num = 0;
         const char* value_ptr;
 
         array_or_group.get_metadata(
             key, &tdb_type, &value_num, (const void**)&value_ptr);
 
-        if (value_ptr == nullptr && value_num != 1)
+        if (value_ptr == nullptr && value_num == 0)
             throw py::key_error("Metadata key '" + key + "' not found");
 
         if (is_ndarray) {


ypatia (approved):

kounelisagis:
> I confirmed that we can drop the patch for the next release. I did a nightly build of https://github.com/TileDB-Inc/TileDB-Py/commit/36e60873bd2b6312f6914cd1fc8a141b9a778c15 in my feedstock fork after deleting the patch, and the linux-ppc64le build with Python 3.10 passed (https://github.com/jdblischak/tiledb-py-feedstock/commit/77c9010ccc8f9b2f300725db7dde18e97bfb3e9b)


But, did you relax the pybind11 constraint so that it installs pybind11 3.0?

jdblischak:
> But, did you relax the pybind11 constraint so that it installs pybind11 3.0?

No. The pybind11 constraint wasn't removed until this morning (https://github.com/TileDB-Inc/TileDB-Py/pull/2249). I am testing https://github.com/TileDB-Inc/TileDB-Py/commit/84af4d5992829c3e1a9454ac0aa59ae61c969ff4 now in https://github.com/jdblischak/tiledb-py-feedstock/commit/3336c4fd6e69bd599b77395b47a5313a1ff22db9

PR #2250, opened 2025-09-26
fix: `.from_pandas()` now creates sparse arrays with no `sparse` param

nickvigilante:
Fixes DOC-843


kounelisagis (approved):
Thanks Nick! Much appreciated.

PR #2253, opened 2025-10-01
Update macOS runners to macOS 15 and macOS 26

nickvigilante:
Also format workflow YAML


ktsitsi:
We can't yank the `macos-13` entirely cause its the only `macos` version that still targets `x86_64` architecture and not `arm64`, so even though we need to address macos-13 sunsetting across the board, entirely removing it will break the support.

.github/workflows/build-wheels.yml
```diff
@@ -32,7 +32,8 @@ jobs:
           - [ubuntu-22.04, manylinux_x86_64]
           - [ubuntu-24.04-arm, manylinux_aarch64]
           - [macos-13, macosx_x86_64]
```
ihnorton, line 34: Please change to `macos-15-intel`

.github/workflows/build-wheels.yml
```diff
@@ -103,53 +103,54 @@ jobs:
...
-            PROJECT_CWD=$PWD
-            rm tiledb/__init__.py
-            cd ..
-            pytest -vv --showlocals $PROJECT_CWD
-
-        - name: "Re-run tests without pandas"
-          run: |
-            pip uninstall -y pandas
-            pytest -vv --showlocals $PROJECT_CWD
+    name: Test source distribution package
```
ihnorton, line 106: Was there a reason to change indentation here? More difficult to review

.github/workflows/ci.yml
```diff
@@ -1,6 +1,13 @@
 name: TileDB Python CI
 
-on: [push, pull_request, workflow_dispatch]
+# on: [push, pull_request, workflow_dispatch]
```
ihnorton, line 3: Change here LGTM but please remove the commented line

ihnorton (changes requested):

.github/workflows/ci.yml
```diff
@@ -1,6 +1,13 @@
...
+
+on:
+  push:
+    branches: ["main"]
+    tags: ["*"]
+  pull_request:
```
ihnorton, line 9: Please add https://github.com/TileDB-Inc/TileDB/blob/d3d2cc816e3d066a1eb40adbbf7478601772bb4e/.github/workflows/full-ci.yml#L16-L18

.github/workflows/build-wheels.yml
nickvigilante: Sorry, my editor auto-formats YAML files on save with Prettier. Do you want me to revert?

kounelisagis:
LGTM. But I'd like to test the build workflow and the wheels first. How should we trigger this?

nickvigilante:
As someone who's not a [full] software engineer, it would make sense to maintain either the macos-15 ARM runner or the macos-26 runner. I'd opt for keeping the macos-26 runner and removing the macos-15 runner, so we just have macos-15-intel and macos-26.

kounelisagis (approved):
Thanks @nickvigilante!

PR #2255, opened 2025-10-06
Hotfix for env var pass

ktsitsi:
In this PR:

- We fix the way the CI_BUILD_WHEEL enviromental variable is being passed in the macos environment and a typo.

Outcome:
- The build now follows the correct path in both Linux and Macos environments and build the wheel.

Requirements:

- macos runners should be updated  #2253 before merge


kounelisagis (approved):
Thanks!

PR #2256, opened 2025-10-07
Update expected error message to adapt test for py314

jdblischak:
I was able to successfully build and test TileDB-Py with Python 3.14 in the feedstock by updating the expected error message for a single test (https://github.com/conda-forge/tiledb-py-feedstock/pull/272).

I wasn't able to add `"3.14"` to `ci.yaml` because `actions/setup-python@v5` doesn't support it yet. It only supports release candidates, but when I installed `"3.14.0-rc.3"`, then the build failed due to issues with finding the correct Python version in `PATH`.


kounelisagis (approved):
Thanks!

PR #2260, opened 2025-10-10
Add support for write mode in `Query`

kounelisagis:
This PR adds support for array writes through the `Query` class on the Python layer. This is particularly useful for global-order writes, since incremental writes in that mode were not possible until now.

---

Closes CORE-211


kounelisagis:
Marking as draft after internal discussions.

PR #2268, opened 2025-11-06
Remove llvm dependency

kounelisagis:
The llvm dependency has been causing issues when building on macOS. The problem was that llvm was fetching specific Python versions, which were then used for the rest of the build process.

Build workflow on this branch: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/19149122864


ihnorton (approved):
Not sure why this would even be here.. thanks!

PR #2274, opened 2025-11-13
Expose the fill value setter at the Python layer

kounelisagis:
Closes CORE-401

### DIFF
--- a/tiledb/tests/test_attribute.py
+++ b/tiledb/tests/test_attribute.py
@@ -83,6 +83,26 @@ class AttributeTest(DiskTestCase):
         attr.fill = fill
         assert attr.fill == fill
 
+    def test_fill_value_setter_wrong_types(self):
+        # Test setting a number on a string attribute
+        # Should fail because integers don't have .encode() method
+        attr_str = tiledb.Attr("test_attr", dtype=str)
+        with pytest.raises(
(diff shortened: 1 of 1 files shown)


tiledb/tests/test_attribute.py
```diff
@@ -63,6 +63,26 @@ def test_attribute_fill(self, dtype, fill):
...
+            (np.dtype("complex64"), (1.0 + 2.0j)),
+            (np.dtype("complex128"), (3.0 + 4.0j)),
+            (np.dtype("complex64"), np.complex64([1.0 + 2.0j])),
+            (np.dtype("complex128"), np.complex128([3.0 + 4.0j])),
+        ],
+    )
+    def test_fill_value_setter(self, dtype, fill):
+        attr = tiledb.Attr("test_attr", dtype=dtype)
+        attr.fill = fill
+        assert attr.fill == fill
```
rroelke: I can only assume that the `@fill.setter` annotation somehow connects this to a C API but as far as I can tell this test will pass even if it does not do so.

If you added a negative test (what happens if you try to set a number on a string attribute? I assume the `encode` raises an exception?) it will demonstrate this crucial aspect well enough, I think

rroelke (approved):

PR #2295, opened 2026-02-05
Add support for Pandas 3

kounelisagis:
This PR adds Pandas 3 support while maintaining Pandas 2 compatibility. The main change is handling the new `StringDtype` that Pandas 3 uses by default for strings.

Also, added a CI step to continue on running tests using Pandas 2.

Closes CORE-487


ihnorton (commented):
LGTM overall but I'm nervous about the changes around string types (see inline) based on past experience.

PR #2302, opened 2026-02-25
Fix Windows wheel builds failing due to shell redirect in version specifier

kounelisagis:
This PR switches cibuildwheel from `test-requires` to `test-extras` so that test dependencies are read from wheel metadata instead of being passed as shell arguments. This fixes `pandas<3` breaking all Windows builds because `<` is a redirect operator in `cmd.exe`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,12 +112,5 @@ select = ["NPY201"]
 "tiledb/__init__.py" = ["F401"]
 
 [tool.cibuildwheel]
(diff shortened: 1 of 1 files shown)


kounelisagis:
Build wheels workflow: https://github.com/TileDB-Inc/TileDB-Py/actions/runs/22396864837

teo-tsirpanis (approved):

PR #2303, opened 2026-03-06
Allow query conditions on dense array dimensions

kounelisagis:
This PR removes Python-side guard that blocked query conditions on dimensions for dense arrays. This was originally added in #722 when Core didn't support QC on dimensions at all. Core has supported this since PR TileDB-Inc/TileDB#3887 but the Python guard was never updated.

### DIFF
--- a/tiledb/query_condition.py
+++ b/tiledb/query_condition.py
@@ -22,7 +22,7 @@ QueryConditionNodeElem = Union[ast.Name, ast.Constant, ast.Call]
 class QueryCondition:
     """
     Class representing a TileDB query condition object for attribute and dimension
-    (sparse arrays only) filtering pushdown.
+    filtering pushdown.
 
     A query condition is set with a string representing an expression
     as defined by the grammar below. A more straight forward example of usage is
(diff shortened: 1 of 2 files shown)


ktsitsi (commented):
What happens if the attribute and the dimension have the same id? How is this resolved?

kounelisagis:
> What happens if the attribute and the dimension have the same id? How is this resolved?

Core's `ArraySchema::check_attribute_dimension_label_names()` throws on `ArraySchema` creation. So, that's not possible:

```
>>> import tiledb, numpy as np
>>> dom = tiledb.Domain(tiledb.Dim(name='x', domain=(0, 9), tile=10, dtype=np.int32))                                                               
>>> attr = tiledb.Attr(name='x', dtype=np.int32)                               
>>> schema = tiledb.ArraySchema(domain=dom, attrs=[attr])                      
TileDBError: ArraySchema: Array schema check failed; Attributes, dimensions and
   dimension labels must have unique names
```

ktsitsi (approved):
LGTM!

